PR #21, opened 2017-06-15
MiniMe dividends implementation 

### LINKED ISSUE #15: Implement dividend sharing using MiniMe cloning
Dividends workflow: 
- Create dividend sharing contract and transfer to it all the tokens that is going to share.
- Contract will do a clone of the token at that block height.
- When a holder calls the contract `collect()` function, his tokens get burned and gets a proportional share of all assets being shared (be careful with the rounding! https://medium.com/@weka/dividend-bearing-tokens-on-ethereum-42d01c710657)

izqui:
Resolves #15 

Initial naive implementation of dividend sharing by cloning and burning tokens. 

Allows every holder of the original token to get a proportional share of the assets the contract owns.


JGcarv:
Hello @izqui !

I could test this, if still needed. Or is it being taken care on the Accounting App already?

Thanks!

onbjerg:
Relevant? @izqui

DeveloppSoft:
Is this still relevant? I have some code ready to be pushed who relies on the minime features instead of burning/cloning tokens

izqui:
Not in this repo anymore, no. This would belong to aragon-apps.

PR #29, opened 2017-06-20
Replace Abstract* for I* for interface like contracts

foopang:
Fix https://github.com/aragon/aragon-core/issues/27


izqui:
Thanks so much!

PR #55, opened 2017-07-05
WIP AccountingApp

### LINKED ISSUE #13: New Accounting app on top of TokenVault
Was blocked by #5

### LINKED ISSUE #14: Implement multi-token accounting settings
Blocked by #13 

Think about whether we want per token budgeting or convert it all to one currency and denominate in that one (better UX)

If we go the one currency route: 
- Initial implementation will need an oracle to feed token prices in
- Ideally will rely on 0x for price discovery when it is ready

readevalprint:
The initial idea is that there are `Transactions` that describe the high level details: from, to, how much, exchange rate and also `TransactionUpdates` which shows the state of the `Transaction` over time.

> Jorge edit: close #13, close #14


contracts/kernel/organs/AccountingApp.sol
```diff
@@ -0,0 +1,98 @@
...
+        int value;
+        address baseToken;
+        address baseValue;
+        
+        address from;
+        address to;
+        string reference;
+
+        uint period;
+        // Back reference the states?
```
izqui: We have to see what is cheaper gas wise. Having a `mapping (uint => TransactionAction)` could turn out to be cheaper

contracts/kernel/organs/AccountingApp.sol
```diff
@@ -0,0 +1,98 @@
+pragma solidity ^0.4.8;                                                         
+
+contract AccountingApp {
```
izqui: Need to add the Application subclass and conform to needed methods

contracts/kernel/organs/AccountingApp.sol
```diff
@@ -0,0 +1,98 @@
+pragma solidity ^0.4.8;
```
izqui: We are using 0.4.11

contracts/kernel/organs/AccountingApp.sol
```diff
@@ -0,0 +1,98 @@
...
+contract AccountingApp {
+    
+    // The concept of sending tokens from one address to another
+    struct Transaction {
+        address token;
+        int value;
+        address baseToken;
+        address baseValue;
+        
+        address from;
```
izqui: Either from or to are always the DAO address. We could save some gas just saving a `senderOrRecipient` address field (or a better name) and then a boolean that indicates what is the transaction direction.

contracts/kernel/organs/AccountingApp.sol
```diff
@@ -0,0 +1,98 @@
...
+        Approved,
+        Pending,
+        Canceled,
+        Failed,
+        Succeeded
+    }
+        
+    
+    struct TransactionAction {
+        uint transactionId;
```
izqui: If we end up doing the mapping thing explained above, we can remove the id here and save 1 sstore per state change :)

contracts/kernel/organs/AccountingApp.sol
```diff
@@ -0,0 +1,98 @@
...
+        TransactionState state;
+        string reason;
+        address actor;
+        uint timestamp;
```
izqui: uint64 are good enough for timestamps and save gas :)

readevalprint:
This is the current state machine, any feedback?
screen shot 2017-07-07 at 11 28 42 am

izqui:
Makes sense

I like the idea of spliting sending a transaction in two actions, send for a approval and 'execute'. This will allow us to have different bylaws for both actions.

readevalprint:
Updated: added flow if the sender is an admin
 image

readevalprint:
That could be overkill, it might be reasonable to make all outgoing transactions require two steps.

contracts/apps/accounting/AccountingApp.sol
```diff
@@ -0,0 +1,133 @@
...
+        bytes4 sig = getSig(payload);
+        return (
+            sig == SET_TRANSACTION_SUCCEEDED_SIG || 
+            sig == SET_TRANSACTION_PENDING_APPROVAL_SIG || 
+            sig == SET_TRANSACTION_FAILED_SIG || 
+            sig == GET_TRANSACTION_STATE_SIG
+        );
+    }
+
+    function setDAOMsg(address sender, address token, uint value) onlyDAO {
```
izqui: setDAOMsg is already done for you by the `Application` base class :)

contracts/apps/accounting/AccountingApp.sol
```diff
@@ -0,0 +1,133 @@
...
+            stringState = "Succeeded";
+        } else if (state == TransactionState.New) {
+            stringState = "New";
+        } else if (state == TransactionState.PendingApproval) {
+            stringState = "PendingApproval";
+        } else if (state == TransactionState.Failed) 
+            stringState = "Failed";
+    }
+
+    function AccountingApp() {
```
izqui: If you don't initialize Application with the DAO address, this contract won't produce bytecode as it is abstract. 

Example https://github.com/aragon/aragon-core/blob/master/contracts/apps/bylaws/BylawsApp.sol#L58

contracts/apps/accounting/AccountingApp.sol
```diff
@@ -0,0 +1,133 @@
...
+    Transaction[] public transactions;
+    TransactionUpdate[] public transactionUpdates;
+
+    // Reverse relation of a Transaction ID  -> TransactionsUpdatesIds[]
+    // transactionUpdatesRelation[tid] = [tuid_0..tuid_N]
+    mapping (uint => uint[]) public transactionUpdatesRelation;
+
+    // Create a new transaction and return the id of the new transaction.
+    // externalAddress is where the transication is coming or going to.
+    bytes4 constant NEW_TRANSACTION_SID = bytes4(sha3('newTransaction(address token, int value, address externalAddress, string reference, TransactionState initialState)'));
```
izqui: I like this right above the function.

Minor detail function signatures are calculated without the parameter names `sha3('newTransaction(address,int..`

contracts/apps/Application.sol
```diff
@@ -1,6 +1,7 @@
 pragma solidity ^0.4.11;
 
 import "../dao/DAOStorage.sol";
+import "../misc/Crontab.sol";
```
izqui: does this have any reason to be here?

test/dao_apps.js
```diff
@@ -57,6 +63,35 @@ contract('Applications', accounts => {
       await mockApp.unprotectedDoStuff()
       assert.isTrue(await mockApp.didStuff(), 'should have done stuff')
     })
+
+    it('can create new transaction', async () => {
```
izqui: We should move all this to its own test file (`integration_accounting` or something like that) as `dao_apps.js` tests that you can install apps in the DAO and not one app in particular.

contracts/apps/accounting/AccountingApp.sol
```diff
@@ -0,0 +1,186 @@
...
+    //
+
+    struct AccountingPeriod {
+        address baseToken;
+        
+        bytes2 ct_hour;
+        bytes2 ct_day;
+        bytes2 ct_month;
+        bytes2 ct_weekday;
+        uint start;
```
izqui: uint64 for timestamps is more than enough

contracts/apps/accounting/AccountingApp.sol
```diff
@@ -0,0 +1,186 @@
...
+    // Create a new transaction and return the id of the new transaction.
+    // externalAddress is where the transication is coming or going to.
+    bytes4 constant NEW_TRANSACTION_SID = bytes4(sha3('newTransaction(address,int,address,string,TransactionState)'));
+    function newTransaction(address token, int value, address externalAddress, string reference, TransactionState initialState) returns (uint) {
+
+        uint tid = transactions.push(Transaction({
+            externalAddress: externalAddress, 
+            token: token, 
+            value: value, 
+            // TODO: get base token and exchange rate from oracle
```
izqui: base token should be gotten from the current accounting period base token, right?

contracts/apps/accounting/AccountingApp.sol
```diff
@@ -0,0 +1,186 @@
...
+        Transaction t = transactions[transactionId];
+        uint lastTransactionUpdate = transactionUpdatesRelation[transactionId][transactionUpdatesRelation[transactionId].length - 1];
+        TransactionUpdate tu = transactionUpdates[lastTransactionUpdate];
+        token = t.token;
+        value = t.value;
+        reference = t.reference;
+        timestamp = t.timestamp;
+        state = tu.state;
+        accountingPeriodId = t.accountingPeriodId;
+        if (state == TransactionState.Succeeded) {
```
izqui: I wouldn't return strings from the contract. Should really be a frontend concern.

contracts/apps/accounting/AccountingApp.sol
readevalprint: ahh yes, english is ambiguous at times, it meant: get base token and then get exchange from oracle

contracts/apps/Application.sol
readevalprint: yes I am about to add the crontab `next()` function now

contracts/apps/accounting/AccountingApp.sol
```diff
@@ -0,0 +1,209 @@
...
+            reason: reason,
+            actor: dao_msg.sender
+        })) - 1;
+        transactionUpdatesRelation[transactionId].push(tuid);
+    }
+
+     function approveTransaction(uint transactionId, string reason) onlyDAO {
+        Transaction memory t = transactions[transactionId];
+        if (t._type == TransactionType.Withdrawal){
+            VaultOrgan(dao).transfer(t.token, t.externalAddress, t.amount);
```
izqui: For avoiding reentrancy the token transfer should be the last thing executed. 

Also, how is this making sure a withdrawal cannot happen more than once?

readevalprint: Excellent point, I will make `updateTransaction` internal and all the `setTransactionX` functions will validate the existing state first. and `approveTransaction` will require the current state to be `pending`. Actually I was thinking of renaming it `approveWithdrawal` because that's a bit more descriptive.

PR #57, opened 2017-07-06
Remove TokensOrgan

izqui:
Most of the logic regarding TokensOrgan had to be duplicated in OwnershipApp. Makes more sense to be app specific (even if it is a very common app), as a DAO doesn't necessarily need to have tokens. 

Also changed logic in OwnershipApp to making the tokenAddress the main identifier for all operations. Given that tokenIds are not unique (they change when a token is deleted), there was this edge case in which a miner could front run a transaction intended for another token by putting a remove token transaction of a less important token (or having bad client side caching and mistaking ids). I know it is too paranoiac, but I think it is cleaner to use the address given it is always unique :)


izqui:
I don't understand why this test is failing as there are no occurrences of TokensOrgan in the directory and I'm fully merged with master...
2017-07-06 at 9 25 57 pm

contracts/apps/capital/CapitalApp.sol
```diff
@@ -38,7 +37,7 @@ contract CapitalApp is Application {
     reverseSales[_saleAddress] = saleIndex;
     saleIndex += 1;
 
-    address tknAddr = TokensOrgan(dao).getToken(sale.tokenId());
+    address tknAddr = address(0);
```
readevalprint: Why is this now 0?

contracts/apps/ownership/OwnershipApp.sol
```diff
@@ -1,59 +1,87 @@
...
 
-// OwnershipApp requires TokensOrgan and ActionsOrgan to be installed in DAO
+// OwnershipApp requires ActionsOrgan to be installed in DAO
 
 // At the moment OwnershipApp intercepts MiniMe hook events, if governance app
 // needs them, it has to have higher priority than ownership app
 
-contract OwnershipApp is Application, Controller, Requestor {
+contract OwnershipConstants {
+  bytes4 constant addTokenSig = bytes4(sha3('addToken(address,uint256,uint128,uint128)'));
```
readevalprint: Can we make constants all caps?
`bytes4 constant ADD_TOKEN_SIG = bytes4(sha3('addToken(address,uint256,uint128,uint128)'));`
http://solidity.readthedocs.io/en/develop/style-guide.html#constants

contracts/apps/ownership/OwnershipApp.sol
```diff
@@ -1,59 +1,87 @@
...
   }
 
-  function removeOrgToken(uint256 tokenId) onlyDAO {
-    TokensOrgan(dao).removeToken(tokenId);
-    if (tokens.length > 1) tokens[tokenId] = tokens[tokens.length - 1];
+  function removeToken(address tokenAddress) onlyDAO {
+    uint tokenId = tokenIdForAddress[tokenAddress];
+    require(tokenId > 0);
+    if (tokens.length > 1) {
+      tokens[tokenId] = tokens[tokens.length - 1];
```
readevalprint: Why would do we need to reshuffle the last one here?

contracts/apps/ownership/OwnershipApp.sol
```diff
@@ -1,59 +1,87 @@
...
+
+    AddedToken(tokenAddress, tokenId);
+
+    if (issueAmount > 0) issueTokens(tokenAddress, issueAmount);
   }
 
-  function removeOrgToken(uint256 tokenId) onlyDAO {
-    TokensOrgan(dao).removeToken(tokenId);
-    if (tokens.length > 1) tokens[tokenId] = tokens[tokens.length - 1];
+  function removeToken(address tokenAddress) onlyDAO {
```
readevalprint: Why would we remove tokens?

contracts/apps/ownership/OwnershipApp.sol
```diff
@@ -96,8 +124,8 @@ contract OwnershipApp is Application, Controller, Requestor {
   }
 
   function isHolder(address _holder) constant returns (bool) {
-    uint tokenCount = tokens.length;
-    for (uint i = 0; i < tokenCount; i++) {
+    uint tokenCount = getTokenCount();
```
readevalprint: This is a small nitpick, but won't this be functionally equivalent to before, but will now incur the gas cost of the function call?

contracts/apps/capital/CapitalApp.sol
izqui: Capital app is going away with my next PR that adds token sales into ownership.

I just changed it to 0 so it would compile after removing tokensorgan

contracts/apps/ownership/OwnershipApp.sol
izqui: Fair enough. Doing it

contracts/apps/ownership/OwnershipApp.sol
izqui: An organization might decide that they no longer one a particular token as a governance token in the org. I agree is very sensitive and the default bylaw should make it very hard.

readevalprint: What happens if we they remove a token that they still hold?

contracts/apps/ownership/OwnershipApp.sol
izqui: GIven that getTokenCount() has specific logic `tokens.length - 1`, i'd rather just have it once. 

Also i'm not sure how expensive internal functions are gas wise, researching it

contracts/apps/ownership/OwnershipApp.sol
izqui: You would still hold your token balance, but the token would no longer be relevant in the context of the organization.

contracts/apps/ownership/OwnershipApp.sol
izqui: The way that we do the remove very cheaply is to remove the last item of the array (with `tokens.length -= 1`) so the token at that index is replaced at the index of the token that is being deleted

PR #60, opened 2017-07-19
Transition from 'canHandlePayload' to a registry based system for dispatching

### LINKED ISSUE #31: Change 'canHandlePayload' constant calls for a storage registry
Similar to how organs register the payload size some functions return, we could replace the current functionality that requires the DispatcherOrgan to make a static call to every Organ (`600 gas * organ length` for worst case scenario, aka, ApplicationOrgan).

The idea is that when installed organs would register in global DAO storage their intent to handle a certain payload. Problem is, we can only save and check against function signatures so this could work. Another solution could be to register to be called for a certain function signature instead of dispatching it directly.

izqui:
Closes #31 

Registry implementation: https://github.com/aragon/aragon-core/pull/60/files#diff-b71e3028d366500d9f183691ddbc64dcR6

#### Advantages
Dramatically reduces gas overhead in every transaction.

Installing apps and organs now use the same storage and dispatching occurs at the Kernel level, saving one `delegatecall` to the `DispatcherOrgan`.

Savings also include not having to do a `call` to every organ or application to ask if they knew how to handle the payload.

We also save `sstores` in every app call not having to do `setDAOmsg` twice.

####  Setbacks
App and organ installations now require sending a `bytes4` array with all the function signature the app will handle. This is something that can easily be done from the client side, but it will make an app behave differently depending on the install parameters (by registering more signatures than it needs or less), could be solved with sha3 of the sigs that we are already calculating.

This also makes it a bit harder to create on-chain factories for DAOs (also the gas costs for deploy is dramatically increased by the number of `sstore` required to install everything).

We should revisit this once dynamic array returns are permitted inside the EVM (Metropolis), and we could implement a standard interface in which organs and applications return all their signatures.

#### Conclusion
Even it is still a work in progress, once it is more polished it is a very nice optimization and simplifies dispatching logic dramatically. 

**WIP**: weird failing test case on `integration_bylaws.js` testing `voting normal flow`, looks like a memory offsetting error.


contracts/kernel/Kernel.sol
```diff
@@ -114,32 +133,16 @@ contract Kernel is IKernel, DAOStorage {
   }
 
   function vaultDeposit(address token, uint256 amount) internal {
-    address vaultOrgan = getOrgan(3);
+    var (vaultOrgan,) = get(0x47e7ef24);
```
izqui: it is the deposit function to the vault. Will make it more explicit

PR #74, opened 2017-07-28
Upgrade coverage for CI

cgewecke:
Hi @izqui - thanks so much for using `solidity-coverage`. You have a fantastic test suite here so it's very exciting for us. 

Looks like it was fun to install! (Sorry about that). In any case, this PR:

+ Updates to the newest `solidity-coverage` version which uses testrpc 4.0.1 and provides branch coverage for `require` and `assert` statements.
+ Uses a newer version of frangio's  Zeppelin shell script to launch. 
+ Adds a small fix to the gas value in `truffle.js`
+ Adds a `coverage` command to the docker script so you can run coverage locally.  (I put this in so I could debug - happy to remove if it's not useful).

**Also Important**: This PR pins Truffle to `3.4.5`. When I ran `docker-compose run core test` it downloaded `3.4.6` which uses `solc 0.4.13` which now throws:
```
,/usr/src/app/coverageEnv/contracts/dao/DAO.sol:55:19: DeclarationError: Identifier not found.
            jumpi(invalidJumpLabel, iszero(result))
                  ^--------------^
```

Maybe that's a bug in solc, or maybe they changed the way you throw in assembly? I found [this](https://github.com/ethereum/solidity/blob/develop/test/libsolidity/SolidityEndToEndTest.cpp#L9512-L9522) in their unit tests. 

Two additional things:
+ The build is SLOW with coverage added: 17 - 19 minutes. (We're working on this.)
+ Its sometimes crashes with an error coming from LevelDB: `key not found`. I think that's a non-deterministic testrpc bug introduced with testrpc v 3.9.2 + and can be worked around just by restarting the build in the Travis container. 

Pressing the green button . . . fingers crossed.


cgewecke:
LOL. I'm not sure this is an improvement! If you hit the rebuild button I think the tests might pass, but no data made it to coveralls so . . . I don't know. I see it's been working before. Will check back tomorrow and maybe close this if it adds nothing.

izqui:
Nice contribution! Just restarted the build. 

Regarding coveralls, it is not working perfectly for us. Some builds make it there, but the badge for the readme still says coverage unknown.

izqui:
Tests are passing now, but still not getting the results in coveralls apparently.

Let me know if you want to make more changes or if it is ready to be merged

cgewecke:
@izqui Do you have any interested in trying [codecov.io](https://codecov.io/) instead of coveralls? They're also free for open source and their webhooks are more reliable. The only downside is that their reports don't show branch coverage, so you can't be sure the tests are hitting every path in the code by inspecting the files on their site. However, you can always do that locally by looking at the HTML report Istanbul generates and places in the `coverage` folder.

If you'd like to try them, just toggle the project on there and I'll rebase & change the script / deps / badge to use them today. Otherwise I have nothing to add. It could be coveralls is just having some problems for the last few days.

cgewecke:
This PR exposed a regression bug in `solidity-coverage` - the Istanbul HTML reports that Coveralls consumes were getting corrupted. Apologies - it's fixed now with `0.2.1` and I've updated the package here. 

Holding off on rebasing until you suggest I should because Master build is currently red.

izqui:
Was that the issue preventing coveralls from reporting?

I think merging into master will fix it given that this pins truffle to the previous version. The upgrade to the new one is in another open PR (but in the middle of more changes).

cgewecke:
Yes - coveralls was failing because of a bug in our code. Thanks!

cgewecke:
Just out of curiosity - is there a box at the bottom of your Coveralls report which shows the badge and has a refresh button? 

I've refreshed my own browser cache and I'm still seeing an 'Unknown' badge which doesn't link directly to the report. Conversely, the 'Repo on GitHub' link at coveralls sends me to `aragon-core/tree` which is `404`. Your badge code looks correct though.

izqui:
I think the issue is related to coveralls not detecting the branch correctly for some reason...
2017-07-31 at 4 06 17 pm
2017-07-31 at 4 09 57 pm

and sync branches doesn't do anything

cgewecke:
One possibility is that it's related to the token in `coveralls.yml`. In theory if you build an open-source repo on Travis you don't need one. Maybe in a subsequent build you could try removing that file? And seeing if it self-corrects? Removing it might force Coveralls to re-process the repo or something . . .

PR #76, opened 2017-07-30
Add update function to MetaOrgan + tests

### LINKED ISSUE #69: Implement atomic organ or app updates
Right now upgrading an app or organ requires two different actions to deregister and the register the upgraded version.

MetaOrgan should provide a way for this actions to occur in the same call so it isn't open attacks in the period between the deregister and register

Quazia:
If I understand the issue correctly I think this handles it. I added an update function for MetaOrgan and then the added corresponding tests into dao_registry. I also slightly changed around the ordering of functions to group the app functions and group the organ functions in MetaOrgan.
Let me know if I've somehow misunderstood the issue.

Closes #69


izqui:
Yes, this is exactly what we needed. 

Will merge soon, thanks for the contribution!

PR #80, opened 2017-08-08
Fix solidity-inspector being a local dependency by publishing it to npm

### LINKED ISSUE #79: Generate factory dependency
Just discovered that [scripts/generate_factory.js](https://github.com/aragon/aragon-core/blob/master/scripts/generate_factory.js) is using a dependency (`solidity-inspector`) with a relative path (see https://github.com/aragon/aragon-core/blob/master/scripts/generate_factory.js#L5) which makes it unusable.

After a quick search I wasn't able to find it online.

izqui:
Fixes #79

Had forgotten to publish our own fork of [`solidity-structure`](https://www.npmjs.com/package/solidity-structure) that `scripts/generate-factory.js` uses to get the needed annotations from source code files.

Running `truffle exec scripts/generate_factory.js` will analyze all Solidity files and generate `BasicFactory.sol` following the a Handlebars template saved at `BasicFactory.sol.tmpl`

### DIFF
--- a/contracts/factories/BasicFactory.sol
+++ b/contracts/factories/BasicFactory.sol
@@ -52,15 +52,17 @@ contract BasicFactory {
     }
 
     function installOrgans(MetaOrgan dao) internal {
-        bytes4[] memory metaorganSigs = new bytes4[](8);
+        bytes4[] memory metaorganSigs = new bytes4[](10);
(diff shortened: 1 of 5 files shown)


DeveloppSoft:
Submitted a pull request on that one from #81, sorry if it is done the wrong way.

PR #85, opened 2017-08-10
Unstuck assets

### LINKED ISSUE #66: Send stuck assets in Applications to DAO
Applications have a different address than the DAO. By mistake someone could send tokens to apps. 

We need to add a method that anyone can call on apps that will send assets stuck to the DAO. This should be implemented in the base Application contract so all apps inherit it.

Similar idea (but `onlyOwner` isn't needed as the funds would go to the DAO)
https://github.com/status-im/status-network-token/blob/master/contracts/SGTExchanger.sol#L107

izqui:
Closes #66

### DIFF
--- a/contracts/apps/Application.sol
+++ b/contracts/apps/Application.sol
@@ -2,6 +2,8 @@ pragma solidity ^0.4.11;
 
 import "../organs/IOrgan.sol";
 import "./IApplication.sol";
+import "zeppelin/token/ERC20.sol";
+import "../kernel/Kernel.sol";
 
 contract Application is IApplication {
     IOrgan.DAOMessage dao_msg;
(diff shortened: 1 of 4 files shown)


onbjerg:
@izqui Still relevant?

izqui:
The problem persists indeed.

We should add this in the Fundraising and Finance apps so if they have any stuck assets they should send them to the Vault they depend on.

For all other apps and kernel, they currently have no way of knowing where they should send those assets inside the DAO. In these cases we could add an ACL protected function that allows to send the stuck assets to an arbitrary address, though I'm not sure it justifies the added complexity to the user.

izqui:
not relevant anymore

PR #86, opened 2017-08-14
Added token whitelist [Draft]

### LINKED ISSUE #72: Implement token whitelist for token transactions
Tokens are arbitrary contracts that can run malicious code. Some sort of whitelist has to be implemented for the tokens that are allowed to execute transactions in the DAO. 

There is an attack possible where you can create a fake token contract and make transactions to the DAO as another person. Only tokens with trusted code (that doesn't allow impersonation) should be added to this whitelist.

I think a good solution for this would be to implement a token whitelist in BylawsApp, and return false in `canPerformAction` if the token that performed the action is not in the whitelist (and therefore the sender cannot be trusted)

JGcarv:
Hello @izqui 

Here's the current state of the contracts. I've done this in different forms but never managed to make it work.

All tests(introduced by me) are failing for different reasons:

`'correctly sets an whitlisted token'` is failing with: `could not unlock signer account`. Which I presume is because I'm trying to send from the dao address, but I could not go around this since the function itself is protected with `onlyDao` modifier.

`allows non listed token deposit` throws somewhere. I honestly don't know why.

The last two `blocks non listed token exectuion` and ` allows listed token execution` are incomplete, because I'm getting `Invalid number of arguments to Solidity function`, even though there are two signatures of the `transfer` function in `Standard23Token` and one of them takes 3 arguments.

Thanks for the help and sorry for the trouble in such a simple issue!

Closes #72

### DIFF
--- a/contracts/apps/bylaws/BylawsApp.sol
+++ b/contracts/apps/bylaws/BylawsApp.sol
@@ -131,6 +131,9 @@ contract BylawsApp is IBylawsApp, Application {
         bytes data
     ) constant returns (bool)
     {
+        if(!isTokenWhitelisted[token] || token != 0)
+           return false
+
         return canPerformAction(
             getSig(data),
             sender,
--- a/contracts/apps/bylaws/BylawsApp.sol
+++ b/contracts/apps/bylaws/BylawsApp.sol
@@ -55,6 +55,8 @@ contract BylawsApp is IBylawsApp, Application {
     Bylaw[] bylaws;
     mapping (bytes4 => uint) public bylawEntrypoint;
 
+    mapping (address => bool) public isTokenWhitelisted;
+
     uint constant PCT_BASE = 10 ** 18;
 
     function BylawsApp(address dao)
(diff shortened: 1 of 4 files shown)


izqui:
- Regarding the first test, `from` cannot be `dao.address` but any external account. What the `onlyDAO` modifier makes sure is that the call was made through the DAO, so all checks are performed. 
- Regarding the second one, no idea what would fail there.
- The third problem is related to web3/truffle-contract poor support for functions with the same name and different parameters (it is trying to execute the normal 2 parameter transfer), you can access the desired function like `token.contract.transfer['address,uint256,bytes]()`, the problem is that this functions aren't promisified and you gotta use callbacks :( See [here](https://github.com/aragon/aragon-core/blob/master/test/dao_dispatch.js#L118)

JGcarv:
Hey @izqui!

I fixed the tests cases and improved the linting a bit. 

What was happening with test 2 is that in the `Kernel`, it would check if `payload.length == 0`, meaning that there were no data and allowing the deposit, but in the token implementation the data, when there was no action, was a byte array filled with zeros. 

I couldn't cast the incoming `bytes` to any other type so I overcame this by checking the first four bytes of the data. It isn't probably the most efficient way, gas-wise, but I not at all familiar with assembly.

If there's anything else that comes to mind, please let me know!

izqui:
Oh, you are right it sends a 32 byte array filled with 0s. Changing [this](https://github.com/aragon/aragon-core/blob/master/test/helpers/Standard23Token.sol#L45) from 32 to 0 should do the trick and we should be able to get rid of the `isPayloadEmpty` check

JGcarv:
Done!

izqui:
It is fine, thanks for your contribution!

PR #87, opened 2017-08-17
Change bylaw ids from incremental integers to hashes and other security measures

izqui:
This PR focuses on improving security of multiple things:
- Bylaw ids being incremental integers were vulnerable to chain reorg attacks (submit a bylaw and expect it to have id x and it end up being y). Now the bylaw id is a hash of the parameters of the bylaw so it is deterministic. 
- Makes sure the DAO reference of an app cannot be changed once it has been set.

### DIFF
--- a/contracts/apps/bylaws/BylawsApp.sol
+++ b/contracts/apps/bylaws/BylawsApp.sol
@@ -132,10 +132,14 @@ contract BylawsApp is IBylawsApp, Application {
         uint id = uint(sha3(msg.sig, _statusNeeded, _isTokenHolderStatus, _not));
         Bylaw storage bylaw = bylaws[id];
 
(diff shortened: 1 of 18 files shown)


contracts/apps/bylaws/BylawsApp.sol
```diff
@@ -207,14 +208,16 @@ contract BylawsApp is IBylawsApp, Application {
     existing_bylaw(_leftBylawId)
     external returns (uint)
     {
-        var (id, bylaw) = newBylaw();
+        uint id = uint(sha3(msg.sig, _combinatorType, _leftBylawId, _rightBylawId, _not));
+        Bylaw storage bylaw = bylaws[id];
```
readevalprint: Event or store to list?

PR #95, opened 2017-09-06
Implement Entity based VotingApp

izqui:
Implements the new voting app for v0.5 in which the app is its own identity (no need to have individual vote contracts any more).

In this initial implementation no voting metadata is saved on chain. Should we include any? Should that be a string saved on-chain or a reference to a content addressable document in IPFS or Swarm? cc @luisivan 

What a voting will execute is determined by its `executionScript`.  The script allows multiple actions to be voted in just one voting allowing more complex workflows.

A script is just a bytes array that contains an arbitrary number of (destination address, calldata) pairs. When executed, the voting app will do as many `CALL` as specified in the script. In case one of the calls fails or it goes out of gas, the entire execution is reverted.

The script idea was inspired by: https://github.com/reverendus/tx-manager/blob/master/src/tx.sol

### DIFF
--- a/contracts/apps/App.sol
+++ b/contracts/apps/App.sol
@@ -4,7 +4,7 @@ import "./AppStorage.sol";
 
 contract App is AppStorage {
     modifier auth {
-        require(kernel.canPerform(msg.sender, address(this), msg.sig));
+        require(address(kernel) == 0 || kernel.canPerform(msg.sender, address(this), msg.sig));
         _;
     }
 }
--- a/contracts/apps/voting/EVMScriptRunner.sol
+++ b/contracts/apps/voting/EVMScriptRunner.sol
@@ -19,7 +19,7 @@ contract EVMScriptRunner {
         }
     }
 
-    function makeSingleScript(address to, bytes calldata) internal returns (bytes script) {
+    function makeSingleScript(address to, bytes calldata) constant returns (bytes script) {
         uint l = 20 + 32 + calldata.length;
 
         uint srcPointer; uint dstPointer;
--- a/contracts/apps/voting/EVMScriptRunner.sol
+++ b/contracts/apps/voting/EVMScriptRunner.sol
@@ -19,7 +19,7 @@ contract EVMScriptRunner {
         }
     }
 
-    function makeSingleScript(address to, bytes calldata) returns (bytes script) {
+    function makeSingleScript(address to, bytes calldata) internal returns (bytes script) {
         uint l = 20 + 32 + calldata.length;
 
         uint srcPointer; uint dstPointer;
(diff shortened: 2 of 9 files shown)


contracts/apps/voting/EVMScriptRunner.sol
```diff
@@ -0,0 +1,101 @@
+pragma solidity 0.4.15;
+
+// Inspired by https://github.com/reverendus/tx-manager
```
readevalprint: nice

contracts/apps/voting/VotingApp.sol
```diff
@@ -0,0 +1,190 @@
...
+        minAcceptQuorumPct = _minAcceptQuorumPct;
+    }
+
+    /**
+    * @notice Create new voting to execute `_executionScript`
+    * @param _executionScript EVM script to be executed on approval
+    * @return votingId id for newly created vote
+    */
+    function newVoting(bytes _executionScript) auth external returns (uint256 votingId) {
+        votingId = votings.length++;
```
readevalprint: Nice, I'll start using this method of determining the id

contracts/apps/voting/VotingApp.sol
```diff
@@ -0,0 +1,190 @@
...
+    function executeVote(uint256 _votingId) external {
+        require(canExecute(_votingId));
+        _executeVote(_votingId);
+    }
+
+    function canVote(uint256 _votingId, address _voter) constant returns (bool) {
+        Voting storage voting = votings[_votingId];
+
+        return voting.open &&
+               uint64(now) < voting.startDate + voteTime &&
```
readevalprint: Could you add parentheses to make sure the order of operations is explicit?

contracts/apps/voting/VotingApp.sol
```diff
@@ -0,0 +1,190 @@
...
+        if (uint64(now) >= voting.startDate + voteTime &&
+            voting.yea >= pct(totalVotes, supportRequiredPct) &&
+            voting.yea >= pct(voting.totalVoters, minAcceptQuorumPct))
+            return true;
+
+        return false;
+    }
+
+    function getVoting(uint256 _votingId) constant returns (bool open, bool executed, address creator, uint64 startDate, uint256 snapshotBlock, uint256 yea, uint256 nay, uint256 totalVoters, bytes32 scriptHash, uint256 scriptActionsCount) {
+        Voting storage voting = votings[_votingId];
```
readevalprint: :D literally 10 return values

onbjerg: jesus 😂

contracts/apps/voting/VotingApp.sol
izqui: 👍. In this case it is probably fine, but we need to be very careful the client handles chain reorgs well if they happen. We don't want users voting on an incorrect votingId.

contracts/apps/voting/VotingApp.sol
izqui: yep

contracts/apps/voting/VotingApp.sol
izqui: We should probably increase it by default. AFAIK the cost is 1 gas per 32 bytes of memory expansion

contracts/apps/App.sol
```diff
@@ -4,7 +4,7 @@ import "./AppStorage.sol";
 
 contract App is AppStorage {
     modifier auth {
-        require(kernel.canPerform(msg.sender, address(this), msg.sig));
+        require(address(kernel) == 0 || kernel.canPerform(msg.sender, address(this), msg.sig));
```
readevalprint: Good catch, helps with testing too

PR #96, opened 2017-09-07
Implement Token Manager + Abstract EVMScriptRunner

izqui:
Implements and tests the Token Manager app according to the spec.

Support both 'native tokens' (can be arbitrarily minted) and 'wrapped tokens' (need to stake 1 other token to mint 1).

### DIFF
--- a/test/helpers/evmScript.js
+++ b/test/helpers/evmScript.js
@@ -1,13 +1,15 @@
 const abi = require('ethereumjs-abi')
 
 module.exports = {
+    // Encodes an array of actions ({ to: address, calldata: bytes}) into the EVM script format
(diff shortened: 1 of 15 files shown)


test/helpers/evmScript.js
```diff
@@ -0,0 +1,13 @@
+const abi = require('ethereumjs-abi')
+
+module.exports = {
+    encodeScript: actions => {
```
onbjerg: This looks interesting, but I think for the sake of clarity we should make a habit out of documenting our test helpers as well

izqui: You are very right, fixed!

PR #99, opened 2017-09-13
Finance App

readevalprint:
- [x] update Accounting with recurring payments/invoices
- [x] deposit function should record transactions
- [x] Add natspec
- [ ] More tests


contracts/apps/accounting/FinanceApp.sol
```diff
@@ -0,0 +1,432 @@
...
+    @dev Internal function to start the next accounting period
+    */
+    function _startNextAccountingPeriod() internal {
+        if(accountingPeriods.length == 0 || accountingPeriods[getCurrentAccountingPeriodId()].endTimestamp < now){
+            AccountingPeriod memory ap = defaultAccountingPeriodSettings;
+            ap.startTimestamp = now;
+            uint endTimestamp = next(ap.ct_sec, ap.ct_min, ap.ct_hour, ap.ct_day, ap.ct_month, ap.ct_weekday, ap.ct_year, now);
+            ap.endTimestamp = endTimestamp;
+            // TODO: store endBlock of last accountingPeriod?
+            ap.startBlock = block.number;
```
izqui: Is startBlock used for anything? I would fire an event when the new Accounting Period starts and not store the block number

contracts/apps/accounting/FinanceApp.sol
```diff
@@ -0,0 +1,432 @@
...
+            accountingPeriods.push(ap);
+            //vault.requestAllowances(ap.budgetTokens, ap.budgetAmounts);
```
izqui: Why is this commented out?

contracts/apps/accounting/FinanceApp.sol
```diff
@@ -0,0 +1,432 @@
...
+        }
+    }
+
+    /**
+    @dev External authenticated function to start the next accounting period
+    */
+    function startNextAccountingPeriod() external auth {
```
izqui: startNewAccountingPeriod should not only not be protected behind the ACL but it should be tried to be executed in every write function, as if the accounting period has ended, to ensure transactions are in the correct accounting period.

contracts/apps/accounting/FinanceApp.sol
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    /**
+    @dev Function to actually transfer the tokens to external address
+    */
+    function _executeTransaction(uint transactionId) internal  {
+        Transaction memory t = transactions[transactionId];
+        var (state, r) = getTransactionState(transactionId);
+        require(state == TransactionState.Approved);
+        bool succeeded = ERC20(t.token).transfer(t.externalAddress, t.amount);
```
izqui: This needs to be a transferFrom. The FinanceApp doesn't have a balance in the tokens, just whatever it was allowed.

contracts/apps/accounting/FinanceApp.sol
```diff
@@ -0,0 +1,432 @@
...
+    @notice This will a create a new payment
+    @param token The token that will be paid
+    @param amount The amount to be paid 
+    @param repeat This is the number of times that the payment should be sent 1..n times
+    @param startTimestamp This is when the payments will begin
+    */
+    function newPayment(ERC20 token, uint amount, address to, uint repeat, uint startTimestamp, bytes2 ct_sec, bytes2 ct_min, bytes2 ct_hour, bytes2 ct_day, bytes2 ct_month, bytes2 ct_weekday, bytes2 ct_year) auth external {
+        require(repeat > 0);
+        uint pid = _newPayment(token, amount, to, startTimestamp);
+        _setPaymentSchedual(pid, repeat, ct_sec, ct_min, ct_hour, ct_day, ct_month, ct_weekday, ct_year);
```
izqui: If a payment can be withdrawn on creation, we should make it happen so only one transaction is needed for individual payments.

contracts/apps/accounting/FinanceApp.sol
```diff
@@ -0,0 +1,432 @@
...
+        bytes2 ct_sec;
+        bytes2 ct_min;
+        bytes2 ct_hour;
+        bytes2 ct_day;
+        bytes2 ct_month;
+        bytes2 ct_weekday;
+        bytes2 ct_year;
+        uint startBlock;
+        uint startTimestamp;
+        uint endTimestamp;
```
izqui: By not specifying the uint size the compiler will default to using `uint256` for timestamps and block numbers that could be stored in `uint64`.

contracts/apps/accounting/FinanceApp.sol
```diff
@@ -0,0 +1,432 @@
...
+    event WithdrawalFailed(uint transactionId);
+
+    function () {
+    }
+
+    function FinanceApp() {
+    }
+
+    /**
+    @param vaultAddress The vault app to use with this FinanceApp
```
izqui: Add a `*` before the comment line for conformance with the spec. 

```
/**
* @param ...
*/
```

contracts/apps/accounting/FinanceApp.sol
readevalprint: Ah good point

contracts/apps/accounting/FinanceApp.sol
```diff
@@ -0,0 +1,432 @@
...
+    // throttledFunctions[string key] = timstamp last run
+    mapping (string => uint) throttledFunctions;
+
+    // Reverse relation of a Transaction ID  -> TransactionsUpdatesIds[]
+    // transactionUpdatesRelation[tid] = [tuid_0..tuid_N]
+    mapping (uint => uint[]) public transactionUpdatesRelation;
+
+    event WithdrawalFailed(uint transactionId);
+
+    function () {
```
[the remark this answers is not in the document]
readevalprint: right EtherToken

contracts/apps/accounting/FinanceApp.sol
readevalprint: I hadn't pulled from your branch when I added this initially so the function didn't exist

contracts/apps/accounting/FinanceApp.sol
readevalprint: So you mean if `repeat=1` then do the `token.transfer()` immediately?

izqui: Actually if a transaction can happen at the moment of payment creation, we should execute the transfer.

izqui:
Closed in favor of #118

izqui:
Closed in favor of #118

PR #102, opened 2017-09-14
[Byzantium] Implement support for dynamic returns in delegatecall proxies

izqui:
**Cannot be merged until Byzantium EIPs are implemented in TestRPC as tests will fail. Tested and working on geth 1.7.0.** 

This PR abstracts the concept of a `DelegateProxy` that both `KernelProxy` and `AppProxy` now use for delegatecall'ing to the actual implementation. 

Also forwards will `revert` when attempting to forward a call to an address without code, as it would just fail silently before.


izqui:
Adding Homestead compatibility logic so we can get this merged

PR #105, opened 2017-09-15
Add basic migrations

onbjerg:
Still has room for improvements - i.e. you have to call initialize yourself and add permissions for all the apps - but it's better than no migrations 😄


izqui:
Cool! You can check [this test](https://github.com/aragon/aragon-core/blob/dev/test/kernel_apps.js#L16) where a Kernel instance is fully initialized. Unsure if we should do this as part of the migration.

PR #108, opened 2017-09-19
Delete function_analysis.js

DeveloppSoft:
As discussed on slack, remove the `function_analysis.js` script which isn't used anymore (still uses the old repo architecture).


izqui:
Thanks!

PR #110, opened 2017-09-19
Introduce app defined roles as the basis for ACL + script to generate artifacts for client

izqui:
This PR introduces some important changes, now instead of pure action signatures (`msg.sig`) being used as the basis for the ACL, roles defined by apps are used. This has two main advantages:

- Future-proofness, as apps updates can now change a function signature in an update (changing function name or params) without requiring manual intervention on the ACL to continue functioning correctly.
- Allow to group multiple similar actions under the same role without having to create multiple permissions.

Roles are `bytes32` and an app can define an unlimited number of roles (actually `2^256`). Example:

```
bytes32 constant public MINT_ROLE = bytes32(1);
function mint(address _receiver, uint256 _amount) auth(MINT_ROLE) {
```

The auth modifier will check with the Kernel's ACL whether the sender has the needed role permissions. 

A new metadata format is also included, `metadata/[contractName].json` specifies basic metadata for apps that is then used to generate rich artifacts for client consumption. Running `npm run generate:artifacts-dev` deploys an instance of the contracts and generates a JSON file in `artifacts/[contractName].json` with:

- `id`: ENS compatible name where the package will be located.
- `version`: Semantic version of the package.
- `roles`: Array of role objects (name and role id used in the code).
- `functions`: Array of function objects. Each function objects has its signature, natspec documentation (@notice and @param) and the role needed for calling the function.
- `deployedNetwork`: Network id where the contract is deployed.
- `deployedAddress`: Address of contract in network.
- `abi`: Full contract ABI.
- `bytecode`: Initcode used to deploy contract to network.

Example artifacts file: https://gist.github.com/izqui/e3fd95d990b6139a4f4fd7920ddc9259


izqui:
Metadata is the directory that gets saved to the repo where we enter info that cannot be derived from the source code (appId, version, etc.)

From the metadata file we can then generate the artifact that is the important file (which effectively is metadata + some extra stuff that gets generated on the fly).

scripts/generate_artifacts.js
```diff
@@ -0,0 +1,71 @@
...
+const inspector = require('solidity-inspector')
+
+const getContract = x => artifacts.require(x)
+const flatten = x => [].concat.apply([], x)
+const contractName = p => path.basename(p).split('.')[0]
+
+const metadataDir = './metadata'
+const artifactsDir = './artifacts'
+
+const analizeContract = (f) => {
```
onbjerg: `analyze`

contracts/kernel/Kernel.sol
```diff
@@ -30,60 +34,59 @@ contract Kernel is KernelStorage, Initializable {
...
     /**
     * @dev Creates a permission that wasn't previously set. Access is limited by the ACL.
     *      if a created permission is removed it is possible to reset it with createPermission.
-    * @notice Create a new permission granting `_entity` the ability to perform `_action` on `_app` (setting `_parent` as parent)
-    * @param _entity Address of the whitelisted entity that will be able to perform the action
-    * @param _app Address of the app in which the action will be allowed (requires app to depend on kernel for ACL)
-    * @param _action Function signature of the action (requires action to be protected with the `auth` modifier)
+    * @notice Create a new permission granting `_entity` the ability to perform `_role` on `_app` (setting `_parent` as parent)
+    * @param _entity Address of the whitelisted entity that will be able to perform the role
+    * @param _app Address of the app in which the role will be allowed (requires app to depend on kernel for ACL)
```
onbjerg: A bit strange wording

contracts/kernel/Kernel.sol
```diff
@@ -30,60 +34,59 @@ contract Kernel is KernelStorage, Initializable {
...
     }
 
     /**
     * @dev Creates a permission that wasn't previously set. Access is limited by the ACL.
     *      if a created permission is removed it is possible to reset it with createPermission.
-    * @notice Create a new permission granting `_entity` the ability to perform `_action` on `_app` (setting `_parent` as parent)
-    * @param _entity Address of the whitelisted entity that will be able to perform the action
-    * @param _app Address of the app in which the action will be allowed (requires app to depend on kernel for ACL)
-    * @param _action Function signature of the action (requires action to be protected with the `auth` modifier)
+    * @notice Create a new permission granting `_entity` the ability to perform `_role` on `_app` (setting `_parent` as parent)
```
onbjerg: Strange wording here (sig was just replaced with role)

contracts/kernel/Kernel.sol
```diff
@@ -30,60 +34,59 @@ contract Kernel is KernelStorage, Initializable {
...
+    * @param _entity Address of the whitelisted entity that will be able to perform the role
+    * @param _app Address of the app in which the role will be allowed (requires app to depend on kernel for ACL)
+    * @param _role Function signature of the role (requires role to be protected with the `auth` modifier)
```
onbjerg: Not actually a role signature, just the role identifier bytes, also *requires role to be protected with the `auth` modifier* makes no sense.

contracts/kernel/Kernel.sol
```diff
@@ -30,60 +34,59 @@ contract Kernel is KernelStorage, Initializable {
...
+        _createPermission(_entity, _app, _role, _parent);
     }
 
     /**
     * @dev Grants a permission if allowed. This requires `msg.sender` to have the permission and be its own parent
-    * @notice Grants `_entity` the ability to perform `_action` on `_app` (setting `_parent` as parent)
-    * @param _entity Address of the whitelisted entity that will be able to perform the action
-    * @param _app Address of the app in which the action will be allowed (requires app to depend on kernel for ACL)
-    * @param _action Function signature of the action (requires action to be protected with the `auth` modifier)
+    * @notice Grants `_entity` the ability to perform `_role` on `_app` (setting `_parent` as parent)
```
onbjerg: Wording

contracts/kernel/Kernel.sol
```diff
@@ -30,60 +34,59 @@ contract Kernel is KernelStorage, Initializable {
...
+    * @param _entity Address of the whitelisted entity that will be able to perform the role
```
onbjerg: Wording

contracts/kernel/Kernel.sol
```diff
@@ -30,60 +34,59 @@ contract Kernel is KernelStorage, Initializable {
...
+    * @param _app Address of the app in which the role will be allowed (requires app to depend on kernel for ACL)
```
onbjerg: Wording

contracts/kernel/Kernel.sol
```diff
@@ -30,60 +34,59 @@ contract Kernel is KernelStorage, Initializable {
...
+    * @param _role Function signature of the role (requires role to be protected with the `auth` modifier)
```
onbjerg: Wording

contracts/kernel/Kernel.sol
```diff
@@ -30,60 +34,59 @@ contract Kernel is KernelStorage, Initializable {
...
+        require(permissions[_entity][_app][_role].allowed == false);
 
-        _setPermission(_entity, _app, _action, _parent, true);
+        _setPermission(_entity, _app, _role, _parent, true);
     }
 
     /**
     * @dev Revokes permission if allowed. This requires `msg.sender` to be the parent of the permission
-    * @notice Revokes `_entity` the ability to perform `_action` on `_app`
+    * @notice Revokes `_entity` the ability to perform `_role` on `_app`
```
onbjerg: Wording

contracts/kernel/Kernel.sol
```diff
@@ -30,60 +34,59 @@ contract Kernel is KernelStorage, Initializable {
...
     * @param _entity Address of the whitelisted entity that will be revoked access
-    * @param _app Address of the app in which the action will no longer be allowed
-    * @param _action Function signature of the action being revoked access
+    * @param _app Address of the app in which the role will no longer be allowed
+    * @param _role Function signature of the role being revoked access
```
onbjerg: Wording

contracts/kernel/Kernel.sol
```diff
@@ -30,60 +34,59 @@ contract Kernel is KernelStorage, Initializable {
...
     }
 
     /**
     * @dev Revokes permission if allowed. This requires `msg.sender` to be the parent of the permission
-    * @notice Revokes `_entity` the ability to perform `_action` on `_app`
+    * @notice Revokes `_entity` the ability to perform `_role` on `_app`
     * @param _entity Address of the whitelisted entity that will be revoked access
-    * @param _app Address of the app in which the action will no longer be allowed
-    * @param _action Function signature of the action being revoked access
+    * @param _app Address of the app in which the role will no longer be allowed
```
onbjerg: Wording

contracts/kernel/Kernel.sol
```diff
@@ -102,11 +105,11 @@ contract Kernel is KernelStorage, Initializable {
     * @dev Function called by apps to check ACL on kernel
     * @param _entity Sender of the original call
     * @param _app Address of the app
-    * @param _action Function signature of the action checked
-    * @return boolean indicating whether the ACL allows the action or not
+    * @param _role Function signature of the role checked
```
onbjerg: Wording

scripts/generate_artifacts.js
onbjerg: Also, avoid mental mapping (`f` should be `file`)

scripts/generate_artifacts.js
```diff
@@ -0,0 +1,71 @@
...
+    const filePath = path.join(process.cwd(), 'contracts', f)
+    const functions = inspector.parseFile(filePath).toJSON().functions
+
+    return Object.keys(functions)
+        .map(k => functions[k])
```
onbjerg: Mental mapping

scripts/generate_artifacts.js
```diff
@@ -0,0 +1,71 @@
...
+        .filter(f => f.accessModifier != 'internal' && f.accessModifier != 'private')
+}
+
+const processFile = name => {
```
onbjerg: Can we write a comment or two about what processing a file in this context means, or give the function a more meaningful name?

test/kernel_acl.js
```diff
@@ -11,14 +11,15 @@ contract('Kernel ACL', accounts => {
     const granted = accounts[1]
     const child = accounts[2]
 
-    const action = getSig('upgradeKernel(address)')
+    let role = ''
```
onbjerg: Is empty string required or could it be null?

izqui: Could be null actually

izqui:
Addresses your concerns with the latest commit.

Now json files in artifacts are exported by `index.js`. Should we generate them on `prepublish`?

onbjerg:
Yep, generate them on prepublish if possible :)

PR #112, opened 2017-09-21
Expose artifacts and contracts

onbjerg:
The published package will now expose two objects:

- ``artifacts`` which is an object of generated artifacts (defined in #110), where the key is the artifact ID
- ``contracts`` which is the built truffle contracts, where the key is the contract name


onbjerg:
Never mind, it is possible to deploy the contracts w/ Web3 using `binary` from the artifacts.

PR #115, opened 2017-09-22
Implement group app

izqui:
Group app is a very simple app that keeps a list of addresses that are allowed to forward actions as members of the group.

Having the concept of groups of identities as a Forwarder (app layer) instead of inside the ACL allows to keep a simpler ACL while at the same time allowing to grant and revoke many permissions to one entity just by adding it or removing it from a group.


onbjerg:
Is it possible to have multiple instances of the same app in this architecture? If not, this needs to handle multiple groups somehow.

izqui:
Yes, new architecture supports multiple instances of the same app (updates are centralized on the Kernel so all instances get updated all at once). This is allowing us to have simpler apps.

contracts/apps/group/GroupApp.sol
```diff
@@ -0,0 +1,74 @@
...
+
+import "../App.sol";
+
+import "../../common/Initializable.sol";
+import "../../common/IForwarder.sol";
+import "../../common/EVMCallScript.sol";
+
+contract GroupApp is App, Initializable, IForwarder, EVMCallScriptRunner {
+    mapping (address => bool) members;
+    string context;
```
onbjerg: Any reason you went with `context` instead of `name`?

contracts/apps/group/GroupApp.sol
```diff
@@ -0,0 +1,74 @@
...
+
+    event AddMember(address indexed entity);
+    event RemoveMember(address indexed entity);
+
+    bytes32 constant public ADD_MEMBER_ROLE = bytes32(1);
+    bytes32 constant public REMOVE_MEMBER_ROLE = bytes32(2);
+
+    modifier onlyMember {
```
onbjerg: I see this modifier is only used once on L58, is there a reason it's a modifier?

contracts/apps/group/GroupApp.sol
```diff
@@ -0,0 +1,74 @@
...
+    function forward(bytes _evmCallScript) onlyMember external {
+        runScript(_evmCallScript);
+    }
+
+    function isMember(address _entity) constant returns (bool) {
+        return members[_entity];
+    }
+
+    function canForward(address _sender, bytes _evmCallScript) constant returns (bool) {
+        _evmCallScript;
```
onbjerg: What does this line do?

izqui: Just silences the Solidity warning about the variable not being used

contracts/apps/group/GroupApp.sol
izqui: Not anymore. There used to be more functions that used it.

contracts/apps/group/GroupApp.sol
onbjerg: Oh, okay.

onbjerg:
Coveralls is saying it doesn't know because this is branched off of pre-coveralls fix.

PR #118, opened 2017-09-25
v0.5.0 Finance app

### LINKED ISSUE #13: New Accounting app on top of TokenVault

izqui:
Still working on documentation and integrating EtherToken.

Closes #13


contracts/apps/finance/FinanceApp.sol
```diff
@@ -0,0 +1,428 @@
...
+        address _receiver,
+        uint256 _amount,
+        uint64 _initialPaymentTime,
+        uint64 _interval,
+        uint64 _maxRepeats,
+        string _reference
+    ) auth(PAYMENT_CREATOR_ROLE) transitionsPeriod external returns (uint256 paymentId)
+    {
+
+        require(settings.budgets[_token] > 0); // Token must have been added to budget
```
onbjerg: What if you have no budgets?

test/apps_finance.js
```diff
@@ -0,0 +1,265 @@
...
+    let app, vault, token1, token2, executionTarget = {}
+
+    const n = '0x00'
+    const periodDuration = 100
+    const maxUint = 0
+
+    beforeEach(async () => {
+        vault = await Vault.new()
+
+        token1 = await MiniMeToken.new(n, n, 0, 'n', 0, 'n', true)
```
onbjerg: What's this? It's a lot of mental mapping, so it's hard to understand. Can we be clearer here? The ability to read tests is very very crucial since many contributors might look here instead of documentation (I usually do).

test/apps_finance.js
```diff
@@ -0,0 +1,265 @@
...
+        await token1.generateTokens(vault.address, 100)
+        await token1.generateTokens(accounts[0], 10)
+
+        token2 = await MiniMeToken.new(n, n, 0, 'n', 0, 'n', true)
```
onbjerg: Same as above

contracts/apps/finance/FinanceApp.sol
izqui: You should have a budget in a token before creating a payment that withdraws from that.

There could still be the case in which a token budget that a recurring payment depends on is set to 0, which will make payment executions fail.

test/apps_finance.js
izqui: I agree that test legibility could be very much improved, but that case it is just initializing the token with empty parameters :)

contracts/apps/finance/FinanceApp.sol
onbjerg: The question is whether or not we want to enforce users to create budgets. Is it a technical limitation, or is it a UX one? If it's a UX one, then I don't know if it's a good idea.

I mean, budgetting is good, but not everyone does it, as it introduces some overhead.

izqui: It is a purely UX decision.

I like the idea of allowing not to have budgets and just allow unlimited spending in a given token.  The question is what should be the default behavior. I think it would be a good idea requiring users to explicitly set an 'unlimited' budget in a token for allowing that.

@luisivan thoughts?

izqui: Talked to luis about this and decided that unless a budget is explicitly set, unlimited spending can happen on that token.

We need to add a new function to free a token from having a budget, as it is a different behavior than setting it to 0.

onbjerg: Cool!

izqui:
Ready to merge @onbjerg?

izqui:
I agree, but let's do it as one big PR that enforces the convention to all apps and Kernel :)

PR #123, opened 2017-09-27
Split Token Manager initializers based on mode

izqui:
Creates two different initializers for the two modes a Token Manager can be in.


onbjerg:
Can you cover L70 in this PR?

izqui:
Good catch @onbjerg, done

onbjerg:
L141 is a validation line too, cover that and we're good to go ✅

izqui:
True, covered

PR #124, opened 2017-09-27
Added tests for vault contract

### LINKED ISSUE #119: Tests for Vault
Coverage on the Vault is pretty low. Check Coveralls for uncovered lines.

JGcarv:
Closes #119

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,13 +19,6 @@
   "author": "Aragon",
   "license": "GPLv3",
   "devDependencies": {
-    "eth-ens-namehash": "^2.0.0",
-    "ethereumjs-abi": "^0.6.4",
-    "ethereumjs-util": "^5.1.1",
-    "solium": "^0.5.5",
-    "truffle-config": "^1.0.2",
-    "truffle-expect": "0.0.3",
-    "truffle-hdwallet-provider": "0.0.3",
     "babel-helpers": "^6.24.1",
     "babel-polyfill": "^6.26.0",
     "babel-preset-es2015": "^6.18.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -38,14 +31,17 @@
...
-    "truffle": "^3.4.9",
+    "truffle": "^3.4.11",
     "truffle-config": "^1.0.2",
     "truffle-expect": "0.0.3",
     "truffle-hdwallet-provider": "0.0.3",
-    "zeppelin-solidity": "^1.2.0"
+    "zeppelin-solidity": "^1.3.0"
+  },
+  "dependencies": {
+    "os-tmpdir": "^1.0.2"
```
onbjerg: Why was this added?

onbjerg:
Thanks! 😊 There's still a few uncovered lines - can you cover them?

package.json
JGcarv: Oh! That got there when I was testing npm problem. I wasn't supposed to be saved. I'll remove

onbjerg: Also there's a few deps that we do require that you've removed by accident, e.g. `solium` 😊

JGcarv:
I added two more cases and that increased the branch percentage up to 75.

I believe that the uncovered bits are for the case that the modifier `auth` fails, but those are covered on the kernel app tests cases.

Should I aim for maximum coverage, and therefore, increase redundancy or the other way around?

onbjerg:
No need for that, but L21 is uncovered (which is not a part of the auth modifier) and also there are some external calls made to tokens that should be checked - e.g. L49 that makes sure the Vault throws if the transfer fails. More [here](https://coveralls.io/jobs/29748850/source_files/1768945932).

No need to aim for maximum coverage if it's redundant 😊

JGcarv:
I guess what's is missing now is a failing case in L35, but to get that(without failing before on L32) I would need to create an elaborate test case(running out of gas between the two tx, or mocking a token that only accepts one `approve` or something like that), and I'm not sure if it's worth it.

izqui:
Testing that will be a PITA given that MiniMe doesn't return false on `approve` at all https://github.com/aragon/aragon-core/blob/dev/contracts/common/MiniMeToken.sol#L231

Also the side-effects of it returning false are clear enough (revert state) that I'm comfortable not testing that particular bit.

onbjerg:
Makes sense 👍 Great work, thank you for your contribution @JGcarv!

@izqui merge when ready ⛓

JGcarv:
Awesome. Thanks for the help @onbjerg  and @izqui

PR #126, opened 2017-09-29
Implement BaseFactory for atomic DAO deployments

izqui:
The idea behind having a Factory deploying organizations is only needing 1 on-chain transaction to completely set up a default organization with sane permission defaults.

It performs a deploy of a `KernelProxy`, one `MiniMeToken` and one instance of every default aragon-core apps. 

The default permissions give too much power to the Group called 'Founders Group' in which the creator of the organization is added. For very destructive actions (such as upgrading the Kernel), a voting is required. The voting app uses the created token as its governance token, the creator of the organization is given 1 token so the voting app is usable.

When launch approaches we should create multiple types of factories (kind of like organization templates).

Pending improvements: 
- Instead of receiving all app code contract addresses in the constructor, it should fetch them from the [apm](https://github.com/aragon/apm-contracts) when deployed.

### DIFF
--- a/contracts/factories/BaseFactory.sol
+++ b/contracts/factories/BaseFactory.sol
@@ -25,6 +25,13 @@ contract BaseFactory {
 
     event DAODeploy(address dao, address token);
 
+    /**
+    * @param _kernelRef Reference to deployed kernel
+    * @param _appIds Application ids, order: [VotingApp, TokenManager, GroupApp, FundraisingApp, Vault, FinanceApp]
+    * @param _appCode Reference app code for apps. Same order as appIds
+    * @param _etherToken Deployed ether token to be used in deployed DAOs
+    * @param _minimeFactory Generic minime factory for organization token clones
+    */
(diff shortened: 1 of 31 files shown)


izqui:
Investigating why compilation is failing on Travis when running coverage, it is working locally.

contracts/factories/BaseFactory.sol
```diff
@@ -0,0 +1,192 @@
...
+        etherToken = _etherToken;
+        minimeFactory = _minimeFactory;
+    }
+
+    /**
+    * @notice Deploy Aragon organization
+    * @param _tokenName name for organization token
+    * @param _tokenSymbol symbol for organization token
+    */
+    function deploy(string _tokenName, string _tokenSymbol) {
```
onbjerg: nit: Add some comments here since there's quite a bit going on

test/base_factory.js
```diff
@@ -0,0 +1,16 @@
+const { assertInvalidOpcode } = require('./helpers/assertThrow')
+const BaseFactory = artifacts.require('BaseFactory')
+
+contract('BaseFactory', accounts => {
```
onbjerg: Can we add some tests to check that apps and permissions have been set up and deployed correctly?

PR #142, opened 2017-10-26
Minime token testing

### LINKED ISSUE #120: Tests for MiniMeToken
Check Coverall for uncovered lines.

cpurta:
Added tests for the MiniMeToken contract. Will resolve #120


izqui:
Very good contribution! 

It would be great to use the opportunity to test that transfers in cloned tokens don't affect the other token balance. This will also test the different scenarios for either checking the token's own checkpoints or the parent token balance

PR #147, opened 2017-11-14
Linter upgrade

cpurta:
Upgrade solium to most recent version and update the configuration file to use updated config syntax. Also added some of the fixes suggested by the linter.

Addresses #147

### DIFF
--- a/contracts/common/MiniMeToken.sol
+++ b/contracts/common/MiniMeToken.sol
@@ -37,11 +37,11 @@ contract Controlled {
 
     address public controller;
 
-    function Controlled() { controller = msg.sender;}
+    function Controlled()  public { controller = msg.sender;}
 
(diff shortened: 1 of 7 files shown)


contracts/common/MiniMeToken.sol
```diff
@@ -131,7 +131,7 @@ contract MiniMeToken is Controlled {
         uint8 _decimalUnits,
         string _tokenSymbol,
         bool _transfersEnabled
-    ) {
+    )  public {
```
izqui: Linter is complaining here in the CI:  ERROR: Function 'MiniMeToken': Opening brace must be on a new line immediately after the last modifier.

cpurta: Fixed that error all CI tests are now passing.

PR #148, opened 2017-11-14
Add ability to execute initialization payload on AppProxy deployment

### LINKED ISSUE #144: App initialization front-running
When apps are deployed without using a Factory (after initial DAO setup), they require two steps for creation: first Proxy deploy, then initialization. This can result in an attacker introducing a transaction between both and initializing the app with different parameters (which may go unnoticed).

Possible solutions:
- Make https://github.com/aragon/aragon-core 's `Initializable` inherit from `Ownable` and then make `Initializable` a part of AppProxy default contract, so after a proxy is created it can only be initialized by the same account.

- Add a new optional parameter to AppProxy's initializer for data to be executed on the target contract on initialization. This approach requires the app code to be set before creating the app proxy. This would look like this:

```
    function AppProxy(IKernel _kernel, bytes32 _appId, bytes _initializeData) {
        kernel = _kernel;
        appId = _appId;
        if (_initializeData.length > 0)
             this.call(_initializeData);
    }
```

izqui:
Closes #144 

Adds the ability to provide a payload to be executed when an AppProxy is deployed. This allows to perform an atomic deployment and initialization of any app without the need to have a specific factory.

It shouldn't introduce any security vulnerabilities as the proxy will forward any call in the same way after it being deployed.

**TODO:** Needs documentation update on the wiki

### DIFF
--- a/contracts/apps/AppProxy.sol
+++ b/contracts/apps/AppProxy.sol
@@ -4,9 +4,25 @@ import "./AppStorage.sol";
 import "../common/DelegateProxy.sol";
 
 contract AppProxy is AppStorage, DelegateProxy {
-    function AppProxy(IKernel _kernel, bytes32 _appId) {
+    /**
+    * @dev Initialize AppProxy
+    * @param _kernel Reference to organization kernel for the app
+    * @param _appId Identifier for app
+    * @param _initializePayload Payload for call to be made after setup to initialize
+    */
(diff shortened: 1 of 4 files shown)


onbjerg:
The payload is just generic calldata? What happens if the calldata is malformed?

izqui:
Yes. If it is malformed and the delegatecall fails, then it will revert.

I'm not too worried about the calldata not being an initialization payload, as right after the contract is deployed, it starts forwarding any data to the target contract with a delegatecall anyway.

PR #154, opened 2017-11-19
Remove babel

### LINKED ISSUE #153: Remove Babel
As Node 9 and 8.9.x was released, we no longer have unsupported features in this repository. I think we can safely remove Babel and use Node 8.9.x or Node 9, since we only use Node for tests.

Continuation of old issue #33

onbjerg:
Closes #153

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,0 @@
-{
-  "presets": ["es2015", "stage-2", "stage-3"]
-}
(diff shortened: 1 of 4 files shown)


izqui:
Awesome!

onbjerg:
Merge when the time is right 👌

sohkai:
So it turns out that `import` and `export` don't work transparently yet in Node 9 (not sure if `.mjs` is necessary for modules); just keep this in mind when writing tests.

PR #156, opened 2017-11-27
ACL: permission instance parent -> permission owner

### LINKED ISSUE #155: Simplify permission parent model to a permission owner model
In the current iteration of the Kernel ACL, every permission instance has its own 'parent'. A more detailed explanation can be found in the [ACL section](https://wiki.aragon.one/dev/AragonOS/#1-kernel-and-the-access-control-list) of the AragonOS doc. In this model the parent of a permission instance means two different things:

- Parent can always revoke a permission instance.
- If an entity has a permission and it is also its own parent, it can regrant the permission.

The problem with this permission model is that in case a permission has been granted to an entity that cannot do arbitrary calls (i.e. Finance or Fundraising), the process for granting the permission to another entity (i.e. wanting to install another app that needs vault access) involves revoking the permission in a complex workflow like this:

```
- Finance app has permission X, with parent Y.
- Y revokes Finance app permission X.
- An entity allowed to do `kernel.createPermission` creates the permission X to a temporary entity Z, with parent Z.
- Z grants the new entity the permission.
- Z grants Finance app permission X, with parent Y (reset initial state)
- Z revokes its own permission X.
```

Problems:
- If parent `Y` doesn't have permission to do `kernel.createPermission`, two different entities need to coordinate their actions. The permission cannot be created again while the Finance app has it.
- If the permission has been granted to more entities (and these entities don't have the ability to re-grant), all these entities need to be revoked the permission. If they need to be revoked by different entities, the coordination problem becomes even worse.
- `Z` needs to be trusted to perform all three actions (easily solvable making `Z` a contract that executes a `EVMCallScript`)

## Proposed new model

Introduce a permission `owner` at the permission level and remove the permission `parent` from the permission instance level. The `owner` is able to `grant` or `revoke` an arbitrary number of permission instances of that permission.

- `createPermission(address entity, address app, bytes32 role, address newOwner)`: protected behind the ACL. Creates a new permission setting an owner. Permission is granted to `entity`. If permission has already been created, it fails.
- `grantPermission(address entity, address app, bytes32 role)`: Grants a permission instance for entity. Only callable by permission owner.
- `revokePermission(address entity, address app, bytes32 role)`: Revokes a permission instance for entity. Only callable by permission owner.
- `transferOwnership(address app, bytes32 role, address newOwner)`: only callable by permission owner, new owner will be able to revoke all previous permission instances too (impossible under current system).

This model still allows every permission to have a different `owner`, which is a very important security feature since some permissions need to be more protected than others.

The only disadvantage compared to the current model is that it won't be possible to have two different permission instances anymore (i.e. token transfers can be done by entity A and B) that could be revoked by different entities (entity A can be revoked its permission by C and B by D).

izqui:
Closes #155 (see issue for explanation and motivation).

If merged, requires changes in AragonOS documentation in the wiki.

### DIFF
--- a/contracts/kernel/Kernel.sol
+++ b/contracts/kernel/Kernel.sol
@@ -137,18 +137,6 @@ contract Kernel is IKernel, KernelStorage, Initializable {
     }
 
     /**
-    * @dev Get permission status
-    * @param _entity Entity address checked
-    * @param _app Address of the app
-    * @param _role Identifier for a group of actions in app
-    * @return allowed boolean indicating whether entity has permissions over role
-    * @return owner address that can revoke or grant the permission
-    */
-    function getPermission(address _entity, address _app, bytes32 _role) constant public returns (bool) {
-        return permissions[_entity][_app][_role];
-    }
-
-    /**
     * @dev Get owner address for permission
     * @param _app Address of the app
     * @param _role Identifier for a group of actions in app
@@ -159,14 +147,14 @@ contract Kernel is IKernel, KernelStorage, Initializable {
     }
 
     /**
-    * @dev Function called by apps to check ACL on kernel
+    * @dev Function called by apps to check ACL on kernel or to check permission statu
     * @param _entity Sender of the original call
     * @param _app Address of the app
     * @param _role Identifier for a group of actions in app
     * @return boolean indicating whether the ACL allows the role or not
     */
-    function canPerform(address _entity, address _app, bytes32 _role) constant public returns (bool) {
-        return getPermission(_entity, _app, _role);
+    function hasPermission(address _entity, address _app, bytes32 _role) constant public returns (bool) {
+        return permissions[_entity][_app][_role];
     }
 
     /**
@@ -233,7 +221,7 @@ contract Kernel is IKernel, KernelStorage, Initializable {
     }
 
     modifier auth(bytes32 _role) {
-        require(canPerform(msg.sender, address(this), _role));
+        require(hasPermission(msg.sender, address(this), _role));
         _;
     }
 }
--- a/contracts/kernel/Kernel.sol
+++ b/contracts/kernel/Kernel.sol
@@ -12,13 +12,13 @@ contract Kernel is IKernel, KernelStorage, Initializable {
 
     // whether a certain entity has a permission
     mapping (address => mapping (address => mapping (bytes32 => bool))) permissions;
-    // who is the owner of a permission
-    mapping (address => mapping (bytes32 => address)) permissionOwner;
+    // who is the manager of a permission
+    mapping (address => mapping (bytes32 => address)) permissionManager;
     // appId -> implementation
     mapping (bytes32 => address) appCode;
 
-    modifier onlyPermissionOwner(address app, bytes32 role) {
-        require(msg.sender == getPermissionOwner(app, role));
+    modifier onlyPermissionManager(address app, bytes32 role) {
+        require(msg.sender == getPermissionManager(app, role));
         _;
     }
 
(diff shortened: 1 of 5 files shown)


onbjerg:
Can you add tests to make sure that events (specifically `ChangePermissionOwner`) are emitted when they should be?

contracts/kernel/Kernel.sol
```diff
@@ -47,52 +41,44 @@ contract Kernel is IKernel, KernelStorage, Initializable {
...
     */
     function createPermission(
         address _entity,
         address _app,
         bytes32 _role,
-        address _parent
-    ) auth(CREATE_PERMISSIONS_ROLE) external
+        address _owner
+    )   auth(CREATE_PERMISSIONS_ROLE)
+        external
```
sohkai: This slightly deviates from the [official style guide](https://solidity.readthedocs.io/en/develop/style-guide.html#function-declaration); do you like aligning the modifiers like this?

The official way would be:

```
function createPermission(
    address _entity,
    address _app,
    bytes32 _role,
    address _parent
)
    auth(CREATE_PERMISSIONS_ROLE)
    external
{ ... }
```

Some other places ([L75](https://github.com/aragon/aragon-core/pull/156/files#diff-14a0803379fe8527d47915e3727fe7d0R75), [L95](https://github.com/aragon/aragon-core/pull/156/files#diff-14a0803379fe8527d47915e3727fe7d0R95), and [L113](https://github.com/aragon/aragon-core/pull/156/files#diff-14a0803379fe8527d47915e3727fe7d0R113)) are similar.

contracts/kernel/Kernel.sol
```diff
@@ -47,52 +41,44 @@ contract Kernel is IKernel, KernelStorage, Initializable {
...
+    function grantPermission(address _entity, address _app, bytes32 _role)
+             onlyPermissionOwner(_app, _role)
+             external
     {
-        // Implicitely check parent has permission and its parent of it (if it didn't have permission, it wouldn't have a parent)
-        require(permissions[msg.sender][_app][_role].parent == msg.sender);
-        // Permission can only can be set if entity doesn't already have it
-        require(permissions[_entity][_app][_role].allowed == false);
-
+        // Only permission owner can grant permission
```
sohkai: Maybe move this comment up to the line with the `onlyPermissionOwner` modifier?

contracts/kernel/Kernel.sol
```diff
@@ -143,21 +142,20 @@ contract Kernel is IKernel, KernelStorage, Initializable {
     * @param _app Address of the app
     * @param _role Identifier for a group of actions in app
     * @return allowed boolean indicating whether entity has permissions over role
-    * @return parent address of the entity which can revoke the permission
+    * @return owner address that can revoke or grant the permission
```
sohkai: This line should be removed; `owner` isn't returned.

contracts/kernel/Kernel.sol
```diff
@@ -168,8 +166,7 @@ contract Kernel is IKernel, KernelStorage, Initializable {
     * @return boolean indicating whether the ACL allows the role or not
     */
     function canPerform(address _entity, address _app, bytes32 _role) constant public returns (bool) {
-        var (isAllowed,) = getPermission(_entity, _app, _role);
-        return isAllowed;
+        return getPermission(_entity, _app, _role);
```
sohkai: Maybe we can rename `getPermission()` to `hasPermission()` and remove `canPerform()` altogether?

contracts/kernel/Kernel.sol
```diff
@@ -188,18 +185,19 @@ contract Kernel is IKernel, KernelStorage, Initializable {
...
         address _app,
         bytes32 _role,
-        address _parent
+        address _owner
     ) internal
     {
-        // only allow permission creation when there is no instance of the permission
-        require(getPermissionInstances(_app, _role) == 0);
+        // only allow permission creation when it has no owner (hasn't been created before)
+        require(permissionOwner[_app][_role] == 0);
```
sohkai: 💯:tada: Nice one, this is way more elegant than before.

test/kernel_acl.js
```diff
@@ -88,28 +91,62 @@ contract('Kernel ACL', accounts => {
...
                 assert.isFalse(await kernel.canPerform(granted, app, role))
             })
 
-            it('permissions root can re-create', async () => {
-                await kernel.createPermission(granted, app, role, granted, { from: permissionsRoot })
-                assert.isTrue(await kernel.canPerform(granted, app, role))
+            it('permissions root cannot re-create', async () => {
+                return assertRevert(async () => {
+                    await kernel.createPermission(granted, app, role, granted, { from: permissionsRoot })
+                })
```
sohkai: I'd add a test here to make sure that `granted` could still re-grant itself the permission it owns.

contracts/kernel/Kernel.sol
izqui: Removing it all together as the modifier is self explanatory

contracts/kernel/Kernel.sol
izqui: Good idea

test/kernel_acl.js
izqui: Added https://github.com/aragon/aragon-core/pull/156/commits/8865a1a16776bed9090a2c2e1224a742cd52a005

onbjerg:
🎉

PR #160, opened 2017-12-05
Allow loose solidity versions for interfaces and files imported by apps

### LINKED ISSUE #158: Required solidity version
By requiring a fixed solidity version in all `aragon-core` files, we are forcing all app developers to use the same solidity version, which right now is set to `0.4.15`.

At least we should allow `0.4` versions greater than `15` by changing it to `^0.4.15`.

izqui:
Closes #158 

- Requires a specific solc version (`0.4.15`) in critical core contracts (Kernel, AppProxy, etc.)
- Allows a loose solc version (`^0.4.15`) in contracts that are meant to be imported by app developers, so they can use other versions of solc.

### DIFF
--- a/contracts/apps/App.sol
+++ b/contracts/apps/App.sol
@@ -1,4 +1,4 @@
-pragma solidity 0.4.15;
+pragma solidity ^0.4.15;
 
 import "./AppStorage.sol";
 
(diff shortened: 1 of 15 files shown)


izqui:
`EVMCallScript` is frequently imported by apps that need forwarding functionality.

Regarding `EtherToken` and `Minime`, they will be imported frequently, but only because apps need their interface. I think we can create loosely pinned interfaces for both and pin the actual implementation.

PR #169, opened 2018-01-08
solc update to 0.4.18 (use view and pure 🎉)

### LINKED ISSUE #158: Required solidity version

### LINKED ISSUE #167: Upgrade truffle and solc and use 'view' and 'pure'
Pretty self explanatory

izqui:
Closes #167 

Also locks critical contracts to 0.4.18 and 'interfaces' to `^0.4.18`, so it closes #158 and closes #160 (cc @sohkai please review this)

Didn't bother to update anything ERC677 or MiniMe as those are getting replace soon.

### DIFF
--- a/contracts/apps/AppProxy.sol
+++ b/contracts/apps/AppProxy.sol
@@ -1,4 +1,4 @@
-pragma solidity 0.4.15;
+pragma solidity 0.4.18;
 
 import "./AppStorage.sol";
 import "../common/DelegateProxy.sol";
@@ -22,7 +22,6 @@ contract AppProxy is AppStorage, DelegateProxy {
             // returns ending execution context and halts contract deployment
             require(appCode.delegatecall(_initializePayload));
         }
-
     }
 
     function () payable public {
--- a/contracts/apps/AppProxy.sol
+++ b/contracts/apps/AppProxy.sol
@@ -3,6 +3,7 @@ pragma solidity 0.4.18;
 import "./AppStorage.sol";
 import "../common/DelegateProxy.sol";
 
+
 contract AppProxy is AppStorage, DelegateProxy {
     /**
     * @dev Initialize AppProxy
--- a/contracts/common/DelegateProxy.sol
+++ b/contracts/common/DelegateProxy.sol
@@ -22,7 +22,7 @@ contract DelegateProxy {
         }
     }
 
-    function isContract(address _target) constant internal returns (bool) {
+    function isContract(address _target) internal view returns (bool) {
         uint256 size;
         assembly { size := extcodesize(_target) }
         return size > 0;
--- a/contracts/common/DelegateProxy.sol
+++ b/contracts/common/DelegateProxy.sol
@@ -1,4 +1,4 @@
-pragma solidity 0.4.15;
+pragma solidity 0.4.18;
 
 contract DelegateProxy {
     /**
--- a/contracts/common/DelegateProxy.sol
+++ b/contracts/common/DelegateProxy.sol
@@ -1,5 +1,6 @@
 pragma solidity 0.4.18;
 
+
 contract DelegateProxy {
     /**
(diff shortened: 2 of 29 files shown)


izqui:
Ready for review

contracts/common/DelegateProxy.sol
```diff
@@ -22,7 +23,7 @@ contract DelegateProxy {
         }
     }
 
-    function isContract(address _target) constant internal returns (bool) {
+    function isContract(address _target) internal view returns (bool) {
```
sohkai: @izqui Do you know if there's a list of assembly opcodes that force a function to be `view` instead of `pure` somewhere? The [solidity docs are ambiguous](https://solidity.readthedocs.io/en/develop/contracts.html#pure-functions) as to which ones might do, but I noticed this is marked `view` (due to `extcodesize`) while the ones that deal purely with memory are marked `pure`.

contracts/common/MiniMeToken.sol
```diff
@@ -359,11 +361,11 @@ contract MiniMeToken is Controlled {
         bool _transfersEnabled
     ) public returns(address)
     {
-        if (_snapshotBlock == 0)
-            _snapshotBlock = block.number;
+        uint256 snapshot = _snapshotBlock == 0 ? block.number - 1 : _snapshotBlock;
```
sohkai: Was there a reason for changing this to be `block.number - 1` rather than `block.number` when `_snapshotBlock == 0`?

contracts/common/MiniMeToken.sol
```diff
@@ -1,5 +1,7 @@
 pragma solidity ^0.4.6;
```
sohkai: Should we update this to `^0.4.18` as well?

contracts/common/DelegateProxy.sol
onbjerg: I think it is checked at compile-time @sohkai, so anything that calls `SSTORE` or `SLOAD` will error out. No runtime enforcement

contracts/common/MiniMeToken.sol
izqui: Didn't bother to update anything ERC677 or MiniMe as those are getting replace soon.

contracts/common/MiniMeToken.sol
izqui: Yes, there is a small race condition that can happen if using the same block number. We don't use it anywhere but fixed it jic

contracts/common/DelegateProxy.sol
sohkai: AFAIK they're not enforced at compile-time yet (see warnings in the docs).

I'm guessing any instruction that deals with on-chain storage will force `view`, so:

- `sstore`
- `sload`
- [every instruction in the "State queries" category](https://solidity.readthedocs.io/en/develop/julia.html#low-level-functions)

izqui: Pure is for functions that dont access state (storage nor memory outside function scope)
View for functions that only read state (sload)
Functions that can modify state dont have special modifiers.

contracts/apps/AppProxy.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity 0.4.15;
+pragma solidity 0.4.18;
```
renexdev: should't be  +pragma solidity ^0.4.18; ?

izqui: Yes. All files that are used as interfaces have the `^`, files that will be deployed have a compiler version fixed for security reasons. In the case of aragon apps all have to be fixed

renexdev: https://gist.github.com/renexdev/a6325c0d1ec90a7bbd81898e05dbaf50

PR #179, opened 2018-01-09
Add comments to Initializable

sohkai:
~Not sure if there was a plan to make more use of these functions in `Initializable`, but as they are it doesn't really make sense to keep them.~

Adds a few comments to `Initializable` for implementation details.

### DIFF
--- a/contracts/common/Initializable.sol
+++ b/contracts/common/Initializable.sol
@@ -2,7 +2,7 @@ pragma solidity ^0.4.18;
 
 
 contract Initializable {
-    uint256 private initializationBlock;
+    uint256 public initializationBlock;
 
     modifier onlyInit {
         require(initializationBlock == 0);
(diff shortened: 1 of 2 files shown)


onbjerg:
`getInitializationBlock` is used in aragon.js

contracts/common/Initializable.sol
```diff
@@ -2,28 +2,17 @@ pragma solidity ^0.4.18;
...
     /**
     * @dev Function to be called by top level contract after initialization has finished.
     */
     function initialized() internal onlyInit {
-        initializationBlock = getBlockNumber();
-    }
-
-    function getBlockNumber() internal view returns (uint256) {
-        return block.number;
+        initializationBlock = block.number;
```
izqui: getBlockNumber() is used on purpose so it can be easily mocked

contracts/common/Initializable.sol
```diff
@@ -2,28 +2,17 @@ pragma solidity ^0.4.18;
 
 
 contract Initializable {
-    uint256 private initializationBlock;
+    uint256 public initializationBlock;
```
izqui: We keep it private since there is no need for Apps to read that value or modify it

sohkai:
Ok, looks like we need some comments instead of simplification :).

PR #182, opened 2018-01-11
Add new EVMScript types (DelegateScript and DeployDelegateScript)

### LINKED ISSUE #166: Add conditionals and variable assignment to EVMCallScripts
Might seem crazy, but it could unleash a huge amount of interesting stuff. It could be similar to [Bitcoin script](https://github.com/bitcoin/bitcoin/tree/master/src/script).

Opening a repo under the labs org for experimentation: https://github.com/aragonlabs/callscript-plus

Side idea, we already have a virtual machine to do stuff on, so there is no need to build a VM on a VM. We could add a prefix to current callscripts that would deploy the rest of the payload as a contract, and `delegatecall` into it, effectively running a script written in EVM. `ds-proxy` does a pretty similar thing: https://github.com/dapphub/ds-proxy/blob/master/src/proxy.sol#L39

I would still keep regular call scripts as they are cheaper to run than deploying a contract, but for complex logic I think that's the way to go.

izqui:
Close #166 (as it allows turing complete EVMScripts, by using the EVM by delegatecall'ing)

This PR introduces the notion of EVMScript specs, which will allow us to have more than just one type of EVM scripts when forwarding actions. This will also allow for gracefully upgrading script types. There is a limit of 4,294,967,296 script types, which I think will be fine for a couple centuries.

EVMScripts are an arbitrary byte array that gets run depending on its specId. They can take a byte array as input and return a byte array as output. What this input and output represents can vary depending on the spec. For example `CallScript` ignores input and always returns and empty array, but DelegateScript input is the calldata made to the contract and the output is the forwarded return data from the call. Scripts are responsible for reverting to indicate failure.

Good old EVMCallScripts still work as they did, but they need to be prefixed with 4 bytes indicating that it is the spec ID 1: `0x00000001`. 

Two new types of scripts are introduced:

- **DelegateScript** (spec ID 2): which provides an address of a contract that will be called with delegatecall to its `exec()` function. This allows to execute arbitrary logic, as this contract will be run in the context of the script execution.
- **DeployDelegateScript** (spec ID 3): it builds on top of DelegateScript, but allows providing the init code of a contract that will be deployed at execution time and then a delegatecall will be done to the `exec()` function as well.

I used the opportunity to add the concept of 'banned addresses' which are addresses that the host application prohibits from being called as part of a script. This actually fixes a **pretty bad vulnerability** in which as a token holder, you could use the Token Manager to forward any call to the managed token and mint infinite tokens or burn any tokens (whoops! 😅).
In the case of the newly introduced script types, as it is impossible to know what addresses they will interact with, if any address is banned, the script will fail.

A problem with DelegateScripts is that they can literally modify the entire storage of the app they are being run from. I added a standard protection to the function that protects against changing kernel address and appId, which should never be changed. Apps should check what they need to protect.

Pending review and updating documentation.


izqui:
This is ready for review @sohkai @onbjerg. Would love to merge today or tomorrow.

contracts/evmscript/EVMScript.sol
```diff
@@ -0,0 +1,51 @@
...
+    */
+    function execScript(bytes memory script, bytes memory input, address[] bannedAddrs) protectState internal returns (bytes output) {
+        uint32 spec = script.getSpecId();
+        require(spec > 0 && spec <= LAST_SPEC_ID);
+
+
+        ExecuteScript(msg.sender, spec, keccak256(script));
+
+        // Exec in spec
+        if (spec == CallsScript.SPEC_ID) return CallsScript.execScript(script, input, bannedAddrs); // solium-disable-line lbrace
```
onbjerg: Can we use a switch here provided we'll probably (maybe (eventually)) have more script IDs?

izqui:
The banned addresses blacklist is to limit what addresses can scripts interact with. For example to avoid scripts run in token manager to mint tokens directly. 

IMO a whitelist would make it really clunky, but could be added at the app level, depending on how critical is the situation.

contracts/evmscript/EVMScript.sol
izqui: > Most of the control structures from JavaScript are available in Solidity except for switch and goto

Only way to use a switch in solidity is using assembly, and I don't think it is worth it in this situation

onbjerg: Oh ok

sohkai:
🎊 I think this has the potential to do lots of crazy stuff (in a good way)! When Ara-REPL?

Some thoughts:

-------------------

Architecturally, the implementation doesn't sit right with me. It feels odd that the script runner is so tightly coupled with any app using it due to the inheritance mechanisms (each app would also be redeploying the same implementation).

Maybe after making APM an Aragon governed app, I'm thinking in this pattern too much (aka going ara-crazy), but I see the potential to make this a single-app org that governs a registry of script types.

We could have a `EVMScriptExecutorRegistry` that holds a list of available script types (controlled via ACL), and makes `getScriptExecutor(bytes script) returns (address)` publicly available. Any app wanting to run a script calls `getScriptExecutor(...)`, and then casts the returned address into an `IScriptExecutor` before `delegatecall`ing into its `execScript(script, input)`.

So something like:

```
contract EVMScriptExecutorRegistry is AragonApp {
    struct ExecutorEntry {
        IScriptExecutor executor;
        bool enabled;
    }

    ExecutorEntry[] executors;

    function addScriptExecutor(IScriptExecutor _executor) auth(MANAGE_REGISTRY) {
        executors.push(ExecutorEntry(_executor, true));
    }

    function disableScriptExecutor(uint256 _executorId) auth(MANAGE_REGISTRY) {
        executors[_executorId].enabled = false;
    }

    function getScriptExecutor(bytes _script) returns (IScriptExecutor executor) {
        ExecutorEntry storage entry = executors[_script.getSpecId()];
        if (entry.enabled) {
            executor = entry.executor;
        }
    }
}

contract IScriptExecutor {
    uint256 constant internal SCRIPT_START_LOCATION = 4;

    function execScript(bytes _script, bytes _input) external returns (bytes);

    // Maybe add the decoder functionality here too?
    function getScriptActionsCount(bytes _script) external returns (uint256);
    function getScriptAction(bytes _script, uint256 position) external returns (address, bytes);
}

contract EVMScriptRunner is AppStorage {
    EVMScriptExecutorRegistry registry;

    function runScript(bytes _script, bytes _input) protectState returns (bytes output) {
        IScriptExecutor executor = registry.getScriptExecutor(_script);
        if (address(executor) != 0) {
            output = executor.delegatecall(..., _script, _input);
        }
    }

    modifier protectState {
        address preKernel = kernel;
        bytes32 preAppId = appId;
        _; // exec
        require(kernel == preKernel);
        require(appId == preAppId);
    }
}

contract DemoApp is AragonApp, EVMScriptRunner {
    function fnThatExecutesScript(bytes _script, bytes _input) {
        ...
        runScript(_script, _input);
    }
}
```

Instances of `IScriptExecutor`, e.g. `CallScript`, `DelegateScript`, etc. would be separately deployed and then individually added to `EVMScriptExecutorRegistry` (alternatively, the registry could deploy those instances itself if given the bytecodes).

The missing link is how to give each app an instance of `EVMScriptExecutorRegistry` (see `EVMScriptRunner`). It wouldn't be ideal to set it up each time for every app, so I was thinking about putting it into the kernel...

-------------------

> you could use the Token Manager to forward any call to the managed token and mint infinite tokens or burn any tokens

This is a really good catch :heart:. Forwarding is a general attack vector in aragonOS, especially when the apps _manage_ something, like a token (or other funds, in e.g. Vault, if we build forwarding into apps controlling them).

However, I'm not convinced it's solved best with a generic blacklist when attempting to run a script through `EVMScript`: it feels like it's coupling the separate concerns of checking user input and execution. We have `canForward()` in the forwarding interface to check if an action can be forwarded, which is where I'd be more comfortable implementing something like this if an app required it. Adding it at that level should also allow the client to better detect if an app is available as a forwarding path or not for a specific action.

Another benefit to this pattern would be allowing an app to control whitelists, which may fit `DelegateScript` better. While you can't control what goes on in the delegated script, you can at least control which deployed scripts you allow (e.g. only those that have come through `DeployDelegateScript` first, or those that you've deployed yourself as part of the app). Adding to this whitelist could be controlled via the ACL, as normal.

izqui:
First of all, thanks for the really thorough comment.

> (each app would also be redeploying the same implementation)

To some extent that is the case, given that used apps are behind proxies, it doesn't impact how expensive it is to create app instances, but it does impact how expensive it is to deploy the original app code.

> `contract EVMScriptExecutorRegistry is AragonApp`

What worries me about this is that even if the registry logic doesn't allow for a script spec id from changing under you (spec id 1 will always be executed by the same script logic), if the registry app is upgradeable, then we can no longer trust that assumption. I think this would be a security hole and this app would need to be non-upgradeable, this is bigger than this particular issue, but I really think there is value in pinning an app code forever, making it non-upgradeable (even if it still uses a proxy).

> `if (entry.enabled)`

This would allow whoever the registry manager is to censor a script by blocking a specID. One the other hand it allows to shutdown a script type if a problem is found. Not sure about the right tradeoff here.

> `contract EVMScriptRunner is AppStorage {`

I really like this better than what we are doing right now :)

> It wouldn't be ideal to set it up each time for every app

I wouldn't mind linking it similar to a Solidity library. Have a placeholder in the bytecode and substitute it with the library address at deploy time. Also wouldn't mind exploring an app lookup table at the kernel level.

Have to give a further thought to your ending notes.

sohkai:
> I think this would be a security hole and this app would need to be non-upgradeable, this is bigger than this particular issue, but I really think there is value in pinning an app code forever, making it non-upgradeable (even if it still uses a proxy).

This is true, but would it be less of an issue if upgrading is controlled via permissions, e.g. from APM or AN? Alternatively, we could also throw away the upgrade permission from DAOs holding pinned apps.

Perhaps we could add a flag to APM Repos to block contract updates, although regardless of anything we do there, the DAO could still "upgrade" an app code to anywhere else if the permission is misused.

> I wouldn't mind linking it similar to a Solidity library. Have a placeholder in the bytecode and substitute it with the library address at deploy time. Also wouldn't mind exploring an app lookup table at the kernel level.

I really like putting a link at the kernel level because it gives orgs the ability to choose which registry they'd like to use for all their apps. Allowing apps to choose their own registry would violate the kernel's inversion of control; it'd be a mess if one app depended on one and another on another, but the org itself had no power when it wanted to add its own script type or disallow one.

izqui:
I don't think it is crazy to have the script runner be a dependency at the kernel level tbh. I'm going to play around with the idea and see how it would look like :)

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    function getExecutor(bytes script) public view returns (IEVMScriptExecutor) {
+        return IEVMScriptExecutor(getExecutorRegistry().getScriptExecutor(script));
+    }
+
+    // TODO: Internal
+    function getExecutorRegistry() internal view returns (IEVMScriptRegistry) {
+        address registryAddr = kernel.getApp(EVMSCRIPT_REGISTRY_APP);
+        return IEVMScriptRegistry(registryAddr);
```
sohkai: We should add a `require(registryAddr != address(0))` check here.

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+    * @dev copies and returns last's call data. Needs to ABI decode first
+    */
+    function returnedDataDecoded() internal view returns (bytes ret) {
```
sohkai: We have a few implementations of returning data from a `delegatecall` (see `DelegateProxy`, this, and `DelegateScript`), but each looks somewhat different. Are there different cases for handling each or can we put this functionality into a base contract?

contracts/evmscript/executors/DelegateScript.sol
```diff
@@ -0,0 +1,64 @@
...
+    /**
+    * @dev Delegatecall to contract with input data
+    */
+    function delegate(address addr, bytes memory input) internal returns (bytes memory output) {
+        require(isContract(addr));
+        require(addr.delegatecall(input.length > 0 ? input : defaultInput()));
+        return returnedData();
+    }
+
+    function isContract(address _target) internal view returns (bool) {
```
sohkai: We should make a `ChecksContract` base contract with this and have `DelegateProxy` and this contract inherit from it.

contracts/evmscript/executors/DelegateScript.sol
```diff
@@ -0,0 +1,64 @@
...
+        uint256 size;
+        assembly { size := extcodesize(_target) }
+        return size > 0;
+    }
+
+    function defaultInput() internal pure returns (bytes) {
+        return DelegateScriptTarget(0).exec.selector.toBytes();
```
sohkai: Does this need to be turned into `bytes`? `bytes4` won't work?

contracts/factory/EVMScriptRegistryFactory.sol
```diff
@@ -0,0 +1,45 @@
...
+import "../acl/ACL.sol";
+
+
+contract EVMScriptRegistryFactory is AppProxyFactory, EVMScriptRegistryConstants {
+    address public baseReg;
+    address public baseCalls;
+    address public baseDel;
+    address public baseDeployDel;
+
+    function EVMScriptRegistryFactory() public {
```
sohkai: In general, it might be nice for factories to take in the addresses of any base contracts they need to avoid recreating the same base contracts in multiple deployments. If `0x0` is given, they should deploy a new version.

contracts/kernel/Kernel.sol
```diff
@@ -5,227 +5,52 @@ import "./KernelStorage.sol";
...
-    * @dev Internal function that sets management
-    */
-    function _setPermissionManager(address _newManager, address _app, bytes32 _role) internal {
-        require(_newManager != address(0));
-
-        permissionManager[_app][_role] = _newManager;
-        ChangePermissionManager(_app, _role, _newManager);
+    function hasPermission(address who, address where, bytes32 what, bytes how) view public returns (bool) {
+        IACL _acl = acl();
+        return _acl == address(0) ? true : _acl.hasPermission(who, where, what, how);
```
sohkai: I don't like this `_acl == address(0)` business.

We should make it a rule that you shouldn't use a kernel until it's initialized, otherwise "undefined behaviour" happens. `initialize()` should call an internal `_setApp()` to get past any access controls otherwise set on public functions.

contracts/kernel/KernelProxy.sol
```diff
@@ -11,14 +11,15 @@ contract KernelProxy is KernelStorage, DelegateProxy {
     * @param _kernelImpl Address of the contract used as implementation for kernel
     */
     function KernelProxy(address _kernelImpl) public {
-        kernelImpl = _kernelImpl;
+        apps[keccak256(CORE_NAMESPACE, KERNEL_APP_ID)] = _kernelImpl;
+        assert(apps[KERNEL_APP] == _kernelImpl);
```
sohkai: Seems like a weird place to check that `keccak256(CORE_NAMESPACE, KERNEL_APP_ID) == KERNEL_APP`. The tests might be better?

contracts/kernel/Kernel.sol
```diff
@@ -5,227 +5,52 @@ import "./KernelStorage.sol";
...
-    /**
-    * @dev Changes appCode reference for `_appId`. This role is required before an app with a certain appId working properly
-    * @notice Upgrade app code of `_appId` to new implementation at address `_code` (CRITICAL!)
-    * @param _appId Namehash of the app name
-    * @param _code Address of new implementation for app
-    */
-    function setAppCode(bytes32 _appId, address _code) auth(UPGRADE_APPS_ROLE) external {
-        appCode[_appId] = _code;
-        SetAppCode(_appId, _code);
+    function setApp(bytes32 namespace, bytes32 name, address app) auth(APP_MANAGER) public returns (bytes32 id) {
```
sohkai: I think `setApp()` is trying to do do many things at once.

From what I gather, it:
- Locates the kernel implementation
- Locates base implementations of apps
- Locates "special" installed apps like the EVMScriptRegistry or ACL

Maybe it would make sense to put the last one in another mapping or give it a separate API?

(I also like `setAppCode` for clarity)

contracts/kernel/KernelStorage.sol
```diff
@@ -1,6 +1,19 @@
...
+contract KernelConstants {
+    bytes32 constant public CORE_NAMESPACE = keccak256("core");
+    bytes32 constant public APP_BASES_NAMESPACE = keccak256("base");
+    bytes32 constant public APP_ADDR_NAMESPACE = keccak256("app");
+
+    bytes32 constant public KERNEL_APP_ID = keccak256("kernel.aragonpm.eth");
+    bytes32 constant public KERNEL_APP = keccak256(CORE_NAMESPACE, KERNEL_APP_ID);
+
+    bytes32 constant public ACL_APP_ID = keccak256("acl.aragonpm.eth");
+    bytes32 constant public ACL_APP = keccak256(APP_ADDR_NAMESPACE, ACL_APP_ID);
```
sohkai: I'm a little confused by the naming, since `setApp()` takes in an ID, but we always use the non-ID version to actually locate something (e.g. we always use `setApp(KERNEL_APP)`).

Maybe `<name>_APP_NAME` and `<name>_APP_ID` would work better?

contracts/apps/AppProxyBase.sol
```diff
@@ -1,33 +1,41 @@
 pragma solidity 0.4.18;
 
+import "./IAppProxy.sol";
 import "./AppStorage.sol";
 import "../common/DelegateProxy.sol";
+import "../kernel/KernelStorage.sol";
 
 
-contract AppProxy is AppStorage, DelegateProxy {
+contract AppProxyBase is AppStorage, DelegateProxy, IProxy, KernelConstants {
```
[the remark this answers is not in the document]
izqui: Yup, made some changes in your scripts-suggestions PR for this

contracts/apps/AppProxyPinned.sol
```diff
@@ -0,0 +1,35 @@
+pragma solidity 0.4.18;
+
+import "./AppProxyBase.sol";
+
+
+contract AppProxyPinned is AppProxyBase {
+    address public pinnedCode;
```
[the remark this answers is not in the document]
izqui: Yup, made some changes in your scripts-suggestions PR for this

contracts/apps/AppStorage.sol
```diff
@@ -6,4 +6,6 @@ import "../kernel/IKernel.sol";
 contract AppStorage {
     IKernel public kernel;
     bytes32 public appId;
+    uint256[97] private storageOffset; // forces App storage to start at after 100 slots
+    uint256 private offset;
```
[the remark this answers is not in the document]
izqui: yes

contracts/evmscript/EVMScriptRunner.sol
izqui: Why? Call will fail regardless when trying to call a `IEVMScriptRegistry(0)`

PR #184, opened 2018-01-12
Implement ACL permission interpretation

### LINKED ISSUE #164: Permission parameters
Right now all apps in an organization require the same permission level to upgrade them, `UPGRADE_APPS_ROLE`.

We should decide whether to allow to create special permissions for upgrading certain apps, as some apps are could be more critical than others.

### LINKED ISSUE #168: 100% test coverage

### LINKED ISSUE #172: Add 'anyone can perform this action' notion to ACL
If the last address in Ethereum's address space (`0xffff...ffff` or `address(uint256(-1))`) is given a permission (can perform an action), that means that any address is allowed to perform that action, effectively having an open bylaw. 

This is needed for `apm` governance.

### LINKED ISSUE #197: APMRegistry: `initialize()`'s `require(kernel != address(0));` check can be removed

izqui:
Builds on top of #182

Closes #164 closes #172 closes #197 closes #197.

**This PR achieves 100% test coverage so closes #168 too.**

*there is a weird problem with the NOT op, that stack overflows on testing (that's the only missing code path from being tested.

This PR extends the ACL to allow for permission parameters. These parameters are evaluated when the ACL is checked for whether an entity has a permission to perform an action.

This will allow us to reduce the number of permissions, for example the kernel will stop having two separate `setAppCode(bytes32 appId, address code)` and `upgradeKernel(address newKernel)`, which are doing the same action with different importance.

The syntax will look like this:

```
contract Kernel {
  function setCode(bytes32 id, address code) auth(SET_CODE_ROLE, id) { ... }
}
```
See that the id is passed as a parameter with the role. When evaluating the permission, different entities can have different parameters for the same permission.

For example, the CTO can have the `SET_CODE_ROLE` with the parameter `(param0, not equal, kernel id)` and the voting app `(param0, equal, kernel id)`. Just with this we can remove this redundant permission in the Kernel, while at the same time allowing for cooler configurations. For example, in the Finance app, the `CREATE_PAYMENT_ROLE` can take amount and token parameters, allowing for different permissions to be granted to different entities to limit what tokens can use and the amounts they can send. 

Also major changes in the documentation.


contracts/acl/ACL.sol
```diff
@@ -0,0 +1,348 @@
...
+        permissionManager[roleHash(_app, _role)] = _newManager;
+        ChangePermissionManager(_app, _role, _newManager);
+    }
+
+    function roleHash(address where, bytes32 what) pure internal returns (bytes32) {
+        return keccak256(uint256(1), where, what);
+    }
+
+    function permissionHash(address who, address where, bytes32 what) pure internal returns (bytes32) {
+        return keccak256(uint256(2), who, where, what);
```
sohkai: What do you think about using a string, e.g. "permission" and "role", to prefix each hash instead?

PR #185, opened 2018-01-12
Gas reporter

cpurta:
Add gas reporting for tests. Will be used with https://github.com/aragon/aragon-apps/pull/71 to get gas-reporting testing in all aragon app smart contracts.


izqui:
This is pretty cool, but I'm now unsure we want to run this every time, at least locally, as it makes running the tests way slower.

PR #186, opened 2018-01-14
Implement testing in Geth and Parity

### LINKED ISSUE #171: Run tests in geth and parity
Right now the test suite only runs on ganache-cli (former testrpc). Gearing up to mainnet deployment, we need to be running our tests in real nodes where code will eventually be deployed. 

Great inspiration for how to do this right can be taken from the https://github.com/JoinColony/colonySale repo

cpurta:
Add script to allow for dockerized and local testing using a private development networks. Allows for specific clients to be used. 

Will resolve #171 

Currently only the tests using the Geth clients are working. The parity tests are hanging on send_transaction due to the client not running with the `--geth` flag. In order to use that flag an account will have to be unlocked. 

I am looking into how accounts can be created and unlocked with the dev chain in the docker image provided by parity.


package.json
```diff
@@ -3,15 +3,20 @@
   "version": "2.0.1",
   "description": "Core contracts for Aragon",
   "scripts": {
-    "test": "npm run ganache-cli:dev && truffle test --network rpc",
```
izqui: Why are these renamed back to testrpc? I intentionally renamed them after their rename when updating the dependencies on #169

izqui: I think I now understand the rationale behind it. IMO we should rename the script and references to something other than `testrpc` as it is easy to think only the testrpc client is run. My suggestions would be `test-nodes.sh`, `start-nodes.sh` or `manage-nodes.sh`

.gitignore
```diff
@@ -11,3 +11,7 @@ scTopics
 functions.csv
 artifacts/
 allFiredEvents
+
+# geth and parity specific files
+password
```
izqui: Aren't these files needed to run?

izqui:
This looks amazing!

I had to modify the truffle config gas limit for the rpc network so it would be below the block gas limit in geth. Anyway, blocks are not sealing for some reason and tests are not running locally.

Travis build is failing as well, but exiting without an error code: https://travis-ci.org/aragon/aragon-core/jobs/328620235

package.json
cpurta: I agree. Wasn't quite sure what to name that back to. Seems like the same file name should be applied in the aragon apps as to avoid confusion there as well.

.gitignore
cpurta: No, not needed I was just using that file to store passwords for accounts for geth/parity. I just didn't want to be committing any password files.

package.json
izqui: I agree

cpurta:
Found that for geth if I reduce the gas in the truffle config I no longer the the `Error: exceeds block gas limit` but it appears that truffle does not like to play nicely with the dev chain of geth. Found that in this issue: https://github.com/trufflesuite/truffle/issues/624.

I believe that this is the same issue with parity as well. When running `truffle test --verbose-rpc --network rpc` it is constantly making the `eth_getFilterChanges` and waiting for changes. Found another issue for that: https://github.com/trufflesuite/truffle/issues/389. 

So that presents two options. 
1. Attempt to fix truffle to avoid the data race issue mentioned in https://github.com/trufflesuite/truffle/issues/624
2. Tests will have to use a private development network both locally and in docker. 

I am not opposed to attempting option 1 but that requires learning the truffle codebase and having to go through the PR process (unsure on timeline). If the issue is not being worked on already (does not look like it). It would be nice for truffle to work on the dev chain though. 

Having a private network for all testing does not seem like a bad route to go and it allows for mocking test and main networks at some point without actually deploying to them. Setting up a private network for local test is a bit tedious to get all the nuances correct. Dockerizing a private network setup would more than likely mean having to manage and update the image(s) when new versions come out.

Thoughts on what is the best route to go would be helpful.

izqui:
IMO it is not worth it to try and hack around truffle. I would say let's go for route 2, either by configuring a testnet ourselves (tried before and never got the correct mining and gas settings to work, but definitely doable). 

We can also try to modify geth slightly as @karalabe suggests in the truffle issue and slightly delay block sealing on dev mode. Unsure if there is a CLI setting we can tweak, or we can just modify the source code in geth and make that the docker image we depend on. I feel way more comfortable hacking around geth than truffle ;)

izqui:
> Would it seem more feasible to only have tests (CI and local) use dockerized clients?

Totally, doesn't seem a crazy dependency at all to me

cpurta:
Have geth test running via docker with a dockerized private chain. Some of the issues with the geth dev chain are occurring with the parity dev chain. Looks like it was an issue with truffle and web3. The issue was with web3 and was addressed. One of the guys at truffle have stated that they have their sights set on updating truffle soon to use web3 1.0 which contains the fix. 

Running a dockerized private chain in parity is quite tedious in that it requires setting up a couple nodes and linking them together. I have been able to get it to run locally but setting it up in docker has been difficult and unsuccessful.

izqui:
Thanks for the update and your effort.

izqui:
👍 looks like there are only issues related to the gas limit now, very close!

cpurta:
Addressed the gas issues in the tests. Seeing `Invalid number of arguments to Solidity function` which is odd because checking the files in the build folder they look to be correct and have the same number of parameters as what is being sent to them. 

Looks like revert assertions are not working correctly. Could be a difference in how geth and ganache do reverts. 

Also seeing that some contracts are exceeding the maximum size enforced by geth. So at some point it looks like those will have to be slimmed down a bit in order to deploy to the mainnet.

cgewecke:
@cpurta Sorry to look over your shoulder here but the `revert` issue is known and should be fixed in a Truffle patch release scheduled for this Wednesday. Would really like to understand source of `Invalid number of arguments` error if you figure that out.

cpurta:
@cgewecke, no worries. I appreciate the heads up. One less issue I have to track down the source of. I keep you updated if I find anything.

cpurta:
Based on some initial findings relating to the `Invalid number of arguments` errors looks to be related to loading variables from storage. All functions that fail with that error eventually make a call to  `sload`. Example of this is in `Repo.sol`

```
function getByVersionId(uint _versionId) public view returns (uint16[3] semanticVersion, address contractAddress, bytes contentURI) {
        require(_versionId > 0);
        Version storage version = versions[_versionId];
        return (version.semanticVersion, version.contractAddress, version.contentURI);
}
```

The `view` modifier should allow it to load variables from storage but not store them. I am not sure if truffle could perhaps be compiling the contracts incorrectly, or perhaps that view functions cannot access storage at all. 

Playing around with some tests to confirm.

izqui:
Is there any update on this @cpurta?

cpurta:
@izqui I have not had much time during past week to focus on this. Will be able to spend some time this week on this though. 

No longer seeing the issue with large contract sizes with the most recent changes. One thing that I have narrowed down is the assertion issues. It is coming from Geth and Parity not returning any errors when a revert occurs. Confirmed that testrpc returns an error message stating a revert occurred and this is checked using the test helpers. So tackling that issue will require a different testing procedure for failures. 

Still seeing the `Invalid number of arguments` errors occur and confirmed that it is not from any modifier. I have a feeling that it may be the way that truffle is compiling the contracts. Will dig into that issue a bit more and will provide an update if I find anything.

cpurta:
Filed an issue regarding the `Invalid number of args` issue: https://github.com/trufflesuite/truffle/issues/794

cpurta:
Looks like the geth test is exceeding the maximum time limit for jobs. Not sure if there is a way to speed it up with the current PoA network that it is using. Have looked into trying to use an instant seal network with multiple unlocked accounts but am unable to find anything.

Running locally all tests have passed with an exception of some the minime token tests. I believe that is from the PoA putting transactions in the same block so it not getting the correct balance on some cloning tests (randomly fails).

The parity test is now running successfully. I had to determine what was causing some errors and had to do some custom error handling in the tests since they are specific to web3 handling the response from geth and parity. 

Also had to modify the `assertRevert` functions being used to a try-catch since the `await codeBlock()` would not return the transaction receipt that I need to check if the transaction was successful. There was not a great way to do that with the `assertRevert` function (open to suggestions).

cpurta:
@izqui mind reviewing some of the most recent changes?

karalabe:
Why don't you run Geth in instant sealing mode? (`--dev`)

cpurta:
@karalabe I have done that and I run into the issue of there only being one account which then causes many test to fail due to `Invalid number of arguments`. I am looking into adding accounts on the dev chain and just transferring ether and unlocking them so that they can perform transactions. Do you happen to know of an easy way to do that?

karalabe:
Something like?

```js
acc := personal.newAccoun("");
personal.unlockAccount(acc, "");
eth.sendTransaction({from: eth.accounts[0], to: acc, value: web3.toWei(1000, "ether")});
```

cpurta:
@karalabe that worked. Have all tests passing now.

izqui:
That's amazing, thanks so much @cpurta and @karalabe

scripts/manage-nodes.sh
```diff
@@ -0,0 +1,77 @@
...
+
+    sleep 5
+}
+
+start_geth() {
+    check_docker
+    # pull the latest image using the dev test network
+    docker pull purta/geth-devnet:$GETH_VERSION
+    # run the geth dev network container
+    docker run -d -p 8545:8545 purta/geth-devnet:$GETH_VERSION
```
izqui: Why did you have to create your own images? I think we should eventually move them under aragon's docker account

truffle-config.js
```diff
@@ -23,10 +23,10 @@ const mocha = process.env.GAS_REPORTER ? mochaGasSettings : {}
 module.exports = {
   networks: {
     rpc: {
-      network_id: 15,
+      network_id: 454545,
```
izqui: why this network id?

izqui: it is still 15 on `manage-nodes.sh`

test/apm_registry.js
```diff
@@ -61,21 +60,30 @@ contract('APMRegistry', accounts => {
     })
 
     it('fails to create empty repo name', async () => {
-        return assertRevert(async () => {
-            await registry.newRepo('', repoDev, { from: apmOwner })
-        })
+        try {
```
izqui: Can we still run all of this inside a function like `assertRevert(...)` does? 

We can expect that the function provided returns a Promise to a receipt, and after that we can check it:

```
async function assertThrows(codeBlock, message, errorCode) {
    try {
        const result = await codeBlock()
        assert.equal(result.receipt.status, 0, 'should have failed status')
    } catch (e) {
        return assertError(e, errorCode, message)
    }
    assert.fail('should have thrown before')
}
```

scripts/manage-nodes.sh
cpurta: I had to make my own images so that all the accounts needed for testing could be set up correctly. For geth I ended up having to have a way to build images for a PoA network and the instant seal dev network with accounts being set up on the fly. 

I agree that they should be moved eventually. The code for the geth images are [here](https://github.com/cpurta/geth-devnet) and the parity images are [here](https://github.com/cpurta/parity-instantseal).

truffle-config.js
cpurta: I think that was left over from me making a private PoA network in geth. I think that it can just stay at 15. Will revert that change.

test/apm_registry.js
cpurta: I tried that at first but the `result` that was returned was always undefined. That's why I went to using the try-catch blocks whenever the function was called in the tests. I will mess around with trying to see if I can get that logic reduced into a function.

cpurta:
Addressed the comments, looks like CI failed the coverage test due to testrpc dying in the middle of the test.

izqui:
This looks great.

scripts/manage-nodes.sh
izqui: Perfect. We should create a separate issue for this, as IMO is not a blocker at this point

izqui: Just published it to docker.io/aragon/geth-devnet

cpurta: Happy to use the new image using the aragon namespace, however I see that there is only a `latest` tag on docker hub. Do you want to use that tag or versioned tag like the one that is already being used (`v1.7.3`)?

test/apm_registry.js
```diff
@@ -62,19 +61,19 @@ contract('APMRegistry', accounts => {
 
     it('fails to create empty repo name', async () => {
         return assertRevert(async () => {
-            await registry.newRepo('', repoDev, { from: apmOwner })
+            return await registry.newRepo('', repoDev, { from: apmOwner })
```
sohkai: Since all of our `assert*` helpers are async themselves, these inner functions can just return the promise, e.g.

```js
return assertRevert(() => {
  return registry.newRepo('', repoDev, { from: apmOwner })
})
```

test/apm_repo.js
```diff
@@ -76,20 +75,28 @@ contract('Repo', accounts => {
...
+                let result = await repo.getByVersionId(0)
+            } catch (e) {
+                if (e.message.search('base 16 number') > -1) {
+                    // geth and parity will return values for the repo version which web3 chokes on
+                    // and determines that they cannot be converted to uint16 so we need to catch this error
+                    return true
+                } else {
+                    return assert.isAbove(e.message.search('revert'), -1, 'should have failed with revert')
+                }
+            }
```
sohkai: Somewhat confused by this—`geth` and `parity` don't fail on this but rather, return values? The error that occurs (`e.message.search('base 16 number')`) is from web3 failing to convert the return value?

Sounds nuts. What repo version do they return?

In any case, it might be nice to wrap this so we can still use the generic `assertRevert`:

```js
return assertRevert(async () => {
  try {
    await repo.getByVersionId(0)
  } catch (e) {
    if (e.message.search('base 16 number') > -1) {
      // geth and parity will return values for the repo version which web3 chokes on
      // and determines that they cannot be converted to uint16 so we need to catch this error
        throw new Error('mock revert')
      } else {
        throw e
      }
    }
  }
})
```

test/kernel_acl.js
```diff
@@ -65,14 +65,21 @@ contract('Kernel ACL', accounts => {
...
         const newKernel = Kernel.at(newKernelProxy.address)
-        return assertRevert(async () => {
-          const result = await newKernel.hasPermission(permissionsRoot, app, role, '0x00')
-        })
+        try {
+            let result = await newKernel.hasPermission(permissionsRoot, app, role, '0x00')
+            if (result.receipt) {
+                assert.equal(result.receipt.status, 0, 'should have failed status')
+            } else {
+                assert.isFalse(result, 'should not have permission')
```
sohkai: This also looks nuts—`geth` or `parity` return a result here rather than reverting?

If so, it might also be nice to have a generic wrapper for this (I notice there's another instance of this later), using a mock revert error:

```js
const assertRevertOrFalse = (code, falseMsg) => {
  return assertRevert(async () => {
    const result = await code()
    if (!result) {
      throw new Error('mock revert')
    }
  })
}
```

test/kernel_apps.js
```diff
@@ -46,16 +52,22 @@ contract('Kernel apps', accounts => {
...
 
-        return assertRevert(async () => {
-            await AppProxyUpgradeable.new(kernel.address, appId, initializationPayload)
-        })
+        try {
+            let result = await AppProxyUpgradeable.new(kernel.address, appId, initializationPayload, { gas: 5e6 })
+        } catch (e) {
+            if (codeDeploymentError(e)) { return true }
+            assert.isAbove(e.message.search('revert'), -1, 'should have failed with revert')
+        }
```
sohkai: It'd be nice to hoist this up as a `assertCodeDeploymentErrorOrRevert`:

```js
const assertCodeDeploymentErrorOrRevert = code => {
  return assertRevert(async () => {
    try {
      await code()
    } catch (e) {
      if (codeDeploymentError(e)) { throw new Error('mock revert') }
      else { throw e }
    }
  })
}
```

test/helpers/assertThrow.js
```diff
@@ -4,7 +4,8 @@ function assertError(error, s, message) {
 
 async function assertThrows(codeBlock, message, errorCode) {
     try {
-        await codeBlock()
+        const result = await codeBlock()
+        return assert.equal(result.receipt.status, 0, 'should have failed status')
```
sohkai: Hmm, not sure I like this; the `assert.fail('should have thrown before')` at the end should fail any test that doesn't throw an error.

Not having this bit also lets us not worry about what's being returned from the `codeBlock()`.

test/kernel_apps.js
cpurta: By "hoist" do you mean add this as function in `helpers/assertThrow.js`?

test/apm_registry.js
cpurta: Seems like if we go this route then we will have to add in the logic to handle the 0 status everywhere that we call `assertRevert`. So all test that do not have to deal will odd truffle/geth test differences (`base 16 number`). 

It seems like these tests would then have to use a wrapper to check that the receipt status is 0 and throw the error:

```js
const checkRecieptStatus = code => {
    let result = await code()
    if (result.recipt.status == 0) { throw new Error('mock revert') }
}
```

Which makes the test look something like:
```js
return assertRevert(async () => {
    checkRecieptStatus(() => {
        return await registry.newRepo('', repoDev, { from: apmOwner })
    })
})
```

cpurta:
@sohkai, replied to some of your comments regarding the error checking. Some feedback on that would be great to get this wrapped up.

sohkai:
Hey @cpurta, I've been meaning to get back to looking at this in the last week but have a few other tasks prioritized. I'm really sorry that I've left you hanging (and missed the comments two weeks ago); but know that this is nagging me too (and I hope to get back to it once those other things are done!).

cpurta:
@sohkai, would appreciate some feedback on the comments. Will have to do quite a bit of updating to this PR due to it being a bit outdated but looking to get this one closed soon.

sohkai:
@cpurta Really, really sorry about leaving you hanging. We've continually deprioritized this so I haven't had a chance to really dig back in :(.

Given the current priorities, I doubt we'll be able to look at this until at least the end of July... There's a lot of context required for this and I need some time to get up to speed with all the differences. Something I noticed with ganache-cli a while back was the ability to disable its "easy" error reporting mode to be more similar to real clients ([see the "Important note about contract runtime error reporting" section](https://github.com/trufflesuite/ganache-cli/releases/tag/v6.1.0-beta.0)). Maybe if we enforce this in the testing script, we don't have to use separate logic between the different clients?

PR #199, opened 2018-01-24
Scripts suggestions

sohkai:
Best way to review is probably to go through each commit and check them; they can all be individually reverted.


izqui:
Ready to merge to scripts!

PR #205, opened 2018-01-26
Improvements for App support: add authP modifiers in AragonApp base and add more syntax sugar

izqui:
Looks like we might need to release aOS 3.0.1 earlier than expected 😅

### DIFF
--- a/contracts/apps/AragonApp.sol
+++ b/contracts/apps/AragonApp.sol
@@ -3,15 +3,29 @@ pragma solidity ^0.4.18;
 import "./AppStorage.sol";
 import "../common/Initializable.sol";
 import "../evmscript/EVMScriptRunner.sol";
+import "../acl/ACLSyntaxSugar.sol";
 
 
-contract AragonApp is AppStorage, Initializable, EVMScriptRunner {
+contract AragonApp is AppStorage, Initializable, ACLSyntaxSugar, EVMScriptRunner {
     modifier auth(bytes32 _role) {
-        require(canPerform(msg.sender, _role));
+        require(canPerform(msg.sender, _role, new uint256[](0)));
         _;
     }
 
-    function canPerform(address _sender, bytes32 _role) public view returns (bool) {
-        return address(kernel) == 0 || kernel.hasPermission(_sender, address(this), _role, new bytes(0));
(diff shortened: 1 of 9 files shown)


contracts/apps/AragonApp.sol
```diff
@@ -3,15 +3,29 @@ pragma solidity ^0.4.18;
...
 
-    function canPerform(address _sender, bytes32 _role) public view returns (bool) {
-        return address(kernel) == 0 || kernel.hasPermission(_sender, address(this), _role, new bytes(0));
+    modifier authP(bytes32 _role, uint256[] params) {
+        require(canPerform(msg.sender, _role, params));
+        _;
+    }
+
+    function canPerform(address _sender, bytes32 _role, uint256[] params) public view returns (bool) {
+        bytes memory how; // no need to init memory as it is never used
```
sohkai: We now have two slightly different ways of doing this `uint[]` to `bytes` conversion (see `canPerform()` in `Kernel`), so I'm wondering if it'd be nice to add this as `toBytes()` somewhere. Maybe we can move some of the more general helpers in `ScriptHelpers` (e.g. `uint256At`, `addressAt`, `uint32At`, `locationOf`) to another library like `BytesHelpers`?

contracts/apps/AragonApp.sol
```diff
@@ -3,15 +3,29 @@ pragma solidity ^0.4.18;
...
+        if (params.length > 0) {
+            uint256 byteLength = params.length * 32;
+            assembly {
+                how := params // forced casting
+                mstore(how, byteLength)
+            }
+        }
+        return address(kernel) == 0 || kernel.hasPermission(_sender, address(this), _role, how);
```
sohkai: This should probably be `address(kernel) != 0 && ...`?

`getCode()` will fail if `address(kernel) == 0` so you won't be able to perform the action anyway.

izqui: this is for the case when you are using the app without a proxy

contracts/apps/AragonApp.sol
izqui: Good idea, making this its own issue

PR #215, opened 2018-02-01
Add template support

izqui:
- Rename roles to new scheme (`keccka256('X_ROLE')`)
- Publish APM apps to APM
- Prepare migration files for being imported from other repos (needed for `aragon-apps/templates`)

Requires publishing NPM patch release on merge

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -11,7 +11,7 @@ interface ACLOracle {
 
 
 contract ACL is IACL, AragonApp, ACLHelpers {
-    bytes32 constant public CREATE_PERMISSIONS_ROLE = bytes32(1);
+    bytes32 constant public CREATE_PERMISSIONS_ROLE = keccak256('CREATE_PERMISSIONS_ROLE');
 
     // whether a certain entity has a permission
     mapping (bytes32 => bytes32) permissions; // 0 for no permission, or parameters id
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -11,7 +11,7 @@ interface ACLOracle {
 
 
 contract ACL is IACL, AragonApp, ACLHelpers {
-    bytes32 constant public CREATE_PERMISSIONS_ROLE = keccak256('CREATE_PERMISSIONS_ROLE');
+    bytes32 constant public CREATE_PERMISSIONS_ROLE = keccak256("CREATE_PERMISSIONS_ROLE");
 
     // whether a certain entity has a permission
     mapping (bytes32 => bytes32) permissions; // 0 for no permission, or parameters id
--- a/contracts/factory/APMRegistryFactory.sol
+++ b/contracts/factory/APMRegistryFactory.sol
@@ -53,11 +53,11 @@ contract APMRegistryFactory is DAOFactory, AppProxyFactory, APMRegistryConstants
         bytes32 namespace = dao.APP_BASES_NAMESPACE();
 
         // Deploy app proxies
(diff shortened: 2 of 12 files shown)


migrations/2_apm.js
```diff
@@ -1,12 +1,14 @@
...
-const getContract = name => artifacts.require(name)
+module.exports = async (deployer, network, accounts, arts = null) => {
+  if (arts != null) artifacts = arts // allow running outside
 
-const bases = ['APMRegistry', 'Repo', 'ENSSubdomainRegistrar']
-const baseContracts = bases.map(getContract)
+  const getContract = name => artifacts.require(name)
+
+  const bases = ['APMRegistry', 'Repo', 'ENSSubdomainRegistrar']
+  const baseContracts = bases.map(getContract)
```
sohkai: Assuming `artifacts.require()` only takes in one argument, this could just be `bases.map(artifacts.require)`, which might be clearer :).

contracts/acl/ACL.sol
```diff
@@ -11,7 +11,7 @@ interface ACLOracle {
 
 
 contract ACL is IACL, AragonApp, ACLHelpers {
-    bytes32 constant public CREATE_PERMISSIONS_ROLE = bytes32(1);
+    bytes32 constant public CREATE_PERMISSIONS_ROLE = keccak256("CREATE_PERMISSIONS_ROLE");
```
sohkai: I guess this is more of a cosmetic change, since the `bytes32` value doesn't really matter (as long as it doesn't collide)?

contracts/factory/APMRegistryFactory.sol
```diff
@@ -53,41 +53,55 @@ contract APMRegistryFactory is DAOFactory, AppProxyFactory, APMRegistryConstants
         bytes32 namespace = dao.APP_BASES_NAMESPACE();
 
         // Deploy app proxies
-        ENSSubdomainRegistrar ensSub = ENSSubdomainRegistrar(dao.newAppInstance(ENS_SUB_APP_ID, ensSubdomainRegistrarBase));
-        APMRegistry apm = APMRegistry(dao.newAppInstance(APM_APP_ID, registryBase));
+        ENSSubdomainRegistrar ensSub = ENSSubdomainRegistrar(dao.newAppInstance(keccak256(node, keccak256(ENS_SUB_APP_NAME)), ensSubdomainRegistrarBase));
+        APMRegistry apm = APMRegistry(dao.newAppInstance(keccak256(node, keccak256(APM_APP_NAME)), registryBase));
```
sohkai: From what I can tell, these changes are to make the `ENSSubdomainRegistrar`, `APMRegistry`, and `Repo` appIds match their namehash values. Is this mostly cosmetic or is the plan to use this with ENS in some way?

contracts/acl/ACL.sol
izqui: it was made to make artifact generation easier for contracts that follow this pattern

contracts/factory/APMRegistryFactory.sol
izqui: It allows node-aragon to recognize APM apps when traversing an APM org

contracts/factory/APMRegistryFactory.sol
```diff
@@ -53,41 +53,55 @@ contract APMRegistryFactory is DAOFactory, AppProxyFactory, APMRegistryConstants
...
 
-        // Initialize
-        ens.setOwner(node, ensSub);
-        ensSub.initialize(ens, node);
-        apm.initialize(ensSub);
-
         return apm;
     }
 
+    function b(string memory x) internal pure returns (bytes memory y) {
```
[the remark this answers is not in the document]
izqui: It doesn't work because it thinks it is bytes storage by default.... it sucks

migrations/2_apm.js
izqui: i think artifacts.require takes more than one param

PR #216, opened 2018-02-03
Fix minime tests

DeveloppSoft:
Not wanting to throw out accusations or whatever but the minime tests are a mess, let me give you samples:
  - there are [unused imports](https://github.com/aragon/aragonOS/blob/dev/test/mini_me_token.js#L1)
  - at least one test is failing because of a [forgotten await](https://github.com/aragon/aragonOS/blob/dev/test/mini_me_token.js#L88)
  - there is at least one [contradictory statements](https://github.com/aragon/aragonOS/blob/dev/test/mini_me_token.js#L101)
  - [this test](https://github.com/aragon/aragonOS/blob/dev/test/mini_me_token.js#L94) is actually totally wrong since we are using `accounts[0]` as the source of ethereum transactions: based on [the minime code](https://github.com/aragon/aragonOS/blob/dev/contracts/lib/minime/MiniMeToken.sol#L172) the controller doesn't need any allowance...

Having figured out that, I thought you might be interested in a little update 😏

### DIFF
--- a/test/mini_me_token.js
+++ b/test/mini_me_token.js
@@ -1,14 +1,12 @@
-const { getBlockNumber, getBlock, getBalance } = require('./helpers/web3')
+const { getBlockNumber } = require('./helpers/web3')
 const { assertRevert } = require('./helpers/assertThrow')
 const MiniMeToken = artifacts.require('MiniMeToken')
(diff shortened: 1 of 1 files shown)


izqui:
Those tests were inherited from another codebase and were needing some clean up indeed! 

Thanks for taking the time :)

PR #220, opened 2018-02-12
Add changes to support deployment factories (requires publishing to npm)

izqui:
- Most of the changes are due to gas optimizations.

- Real code changes are in Zeppelin contracts in order to add MintableToken


test/apm_repo.js
```diff
@@ -1,6 +1,7 @@
 const { assertRevert } = require('./helpers/assertThrow')
 
 const Repo = artifacts.require('Repo')
+const getContract = artifacts.require
```
sohkai: Seems unnecessary?

contracts/lib/zeppelin/token/MintableToken.sol
```diff
@@ -0,0 +1,86 @@
+pragma solidity 0.4.18;
```
sohkai: Rather than concating `Ownable` and `MintableToken`, could we split these two out? I assume you just copy pasta'd these contracts from zeppelin?

Also, maybe it'd be useful if we put these somewhere else—e.g. `contracts/staging` or etc to signal that they're only meant for test networks?

contracts/lib/zeppelin/token/BasicToken.sol
```diff
@@ -11,9 +11,11 @@ import '../math/SafeMath.sol';
  */
 contract BasicToken is ERC20Basic {
   using SafeMath for uint256;
+  uint256 totalSupply_;
 
   mapping(address => uint256) balances;
 
+  function totalSupply() public view returns (uint256) { return totalSupply_; }
```
sohkai: These `totalSupply` changes are copy pasta'd from zeppelin, yeah? (I forget where they had this implemented before, but I guess not here?)

izqui: copy pasta indeed

test/apm_repo.js
izqui: In order to support copying code from migrations lol

contracts/lib/zeppelin/token/MintableToken.sol
izqui: which other contracts would fall under `contracts/staging`?

I put it inside the same file on purpose because I didn't want aOS to have an `Ownable.sol` ;)

sohkai: Offline discussion: we're moving this to aragon-apps instead.

izqui:
Moved `MintableToken` to `aragon-apps`

Issue #223, opened 2018-02-12
Rename ACL functions to reduce verbosity


izqui:
Proposed changes: 

```
acl.createPermission(...) -> acl.create(...)
acl.grantPermission(...) -> acl.grant(...)
...
```

IMO it made sense when the ACL was inside the kernel to use more descriptive names, but now that it is its own component, it isn't needed IMO.

Thoughts?

sohkai:
💓

onbjerg:
reported for making too much sense

NFhbar:
https://github.com/aragon/aragonOS/pull/301

sohkai:
Closing as this didn't make it to `aragonOS@4` and would mean another breaking interface change.

PR #226, opened 2018-02-15
FIX: #223 - Rename ACL functions to reduce verbosity

pmespresso:
- Problem:

 when the ACL was inside the kernel to use more descriptive names, but now that it is its own component, it isn't needed.

- Changes:

1. createPermission(...) -> create(...)
2. _createPermission(...) -> _create(...)
3. grantPermission(...) -> grant(...)
4. revokePermission(...) -> revoke(...)
5. _setPermission(...) -> _set(...)


izqui (commented):
This looks amazing, thanks so much for taking the time.

What do you think about also changing:
- `acl.setPermissionManager(...)` -> `acl.setManager(...)`? (Also getPermissionManager)
- `acl.grantPermissionP(...)` -> `acl.grantP(...)` or `acl.grantWithParams` (@sohkai which one do you prefer, i think i like `grantP` more)

renexdev:
Great, its clear that in the context of ACL, the permission/access word is redundant... =)
acl.grantP +1

sohkai:
Agree with `grantP`; we're using the short form everywhere so it's best to go along with it (even though I would advocate that `[name]` and `[name]P` are really easy to mistype/read 😉).

sohkai:
> Should we merge it to an aragonOS 4 branch when we do?

Let's do it!

onbjerg:
I strongly dislike `authP` and `grantP`. It's not clear what they do at first glance from a new developers standpoint, and since we're removing redundancy in this PR, I also think we could add additional context where it makes sense.

Code should be as readable as possible without comments 😊

yjkimjunior:
On second thought, I agree with @onbjerg that grantPermissionP ought to be changed to grantWithParams rather than grantP, to make it clear even at first glance that there is a difference.

sohkai:
Also strongly agree, but then we will literally be breaking everything in our world in 4.0. Migration docs 🤞

onbjerg:
@sohkai We need to do those anyway, this is literally breaking everything in our world, with or without the `grantP` change

sohkai:
Given the time since this was made, and that this would mean another breaking change (to an API that is used in quite a lot of places now), we won't be modifying these interfaces.

Thanks for doing this though @yjkimjunior, and sorry this didn't make it to `aragonOS@4`!

PR #233, opened 2018-02-22
Make pragmas consisitent

verdverm:
pin solidity to `0.4.18`  where contracts lack the `^` and interfaces / libraries posses the `^`

also cleans up some default modifier warnings

---
I attempted to fix the `throw` warnings, but I ended up with failing tests... will create another issue/pr for that

### DIFF
--- a/contracts/lib/ens/ENS.sol
+++ b/contracts/lib/ens/ENS.sol
@@ -17,7 +17,8 @@ contract ENS is AbstractENS {
 
     // Permits modifications only by the owner of the specified node.
     modifier only_owner(bytes32 node) {
-        require(records[node].owner != msg.sender);
+        if (records[node].owner != msg.sender) throw;
+        // require(records[node].owner != msg.sender);
         _;
     }
 
--- a/contracts/lib/ens/PublicResolver.sol
+++ b/contracts/lib/ens/PublicResolver.sol
@@ -40,7 +40,8 @@ contract PublicResolver {
     mapping(bytes32=>Record) records;
 
     modifier only_owner(bytes32 node) {
-        require(ens.owner(node) != msg.sender);
+        if (ens.owner(node) != msg.sender) throw;
+        // require(ens.owner(node) != msg.sender);
         _;
     }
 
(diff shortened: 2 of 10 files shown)


contracts/apps/AragonApp.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.18;
+pragma solidity 0.4.18;
```
izqui: The `AragonApp` contract is not pinned given that it might be imported by developers running a future (non-breaking) compiler version.

contracts/lib/ens/AbstractENS.sol
```diff
@@ -1,14 +1,14 @@
-pragma solidity ^0.4.15;
```
izqui: Contracts in the `lib` directory are just copied from their source because they went through code review already. Please revert changes in lib to their original version!

contracts/apm/Repo.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.15;
+pragma solidity 0.4.18;
```
izqui: 👍

contracts/apps/AppStorage.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.18;
+pragma solidity 0.4.18;
```
izqui: Needs to be loose as it is used from `AragonApp` which needs to be loose as well

contracts/apps/IAppProxy.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity 0.4.18;
+pragma solidity ^0.4.18;
```
izqui: 👍

contracts/common/Initializable.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.18;
+pragma solidity 0.4.18;
```
izqui: Same as `AppStorage`

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.18;
+pragma solidity 0.4.18;
```
izqui: Same as `AppStorage`

contracts/evmscript/IEVMScriptRegistry.sol
```diff
@@ -1,6 +1,8 @@
 pragma solidity 0.4.18;
 
 
+// TODO should this be in a separate file?
```
izqui: I agree!

izqui:
Thanks for the contribution! As a rule of thumb: everything that will be imported by 3rd party devs to make an app should have a loose version, everything else is locked.

verdverm:
I think I'm going to pull a fresh copy and separate the pragmas and view warnings into two PRs

verdverm:
See #235 and #236

PR #235, opened 2018-02-23
cleans up a number of visibility warnings

verdverm:


contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -36,7 +36,7 @@ contract EVMScriptRunner is AppStorage, EVMScriptRegistryConstants {
     /**
     * @dev copies and returns last's call data. Needs to ABI decode first
     */
-    function returnedDataDecoded() internal view returns (bytes ret) {
+    function returnedDataDecoded() internal pure returns (bytes ret) {
```
sohkai: I'm not 100% sure, but I think this should be a `view` function since it's relying on manipulating memory. Correct me if I'm wrong though; my understanding was that `pure` is for functions that only use their params.

contracts/evmscript/executors/DelegateScript.sol
```diff
@@ -51,7 +51,7 @@ contract DelegateScript is IEVMScriptExecutor {
     /**
     * @dev copies and returns last's call data
     */
-    function returnedData() internal view returns (bytes ret) {
+    function returnedData() internal pure returns (bytes ret) {
```
sohkai: See above.

contracts/evmscript/EVMScriptRunner.sol
sohkai: Note to self: clearly I was wrong. `view` is only necessary for `SLOAD` and `SSTORE`.

.travis.yml
```diff
@@ -8,6 +8,7 @@ node_js:
   - '8'
 env:
   - TASK=lint
+  - TASK=test
```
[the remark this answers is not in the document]
sohkai: I'll remove.

sohkai:
So it turns out we can't really use `pure` or `view` yet in our tests cause we're on the old solidity-coverage (it's kind of funky, not 100% sure but I'm guessing their code mangling didn't understand those new modifiers at that point).

PR #236, opened 2018-02-23
updates some pragma versions

verdverm:


contracts/ens/ENSConstants.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.18;
+pragma solidity 0.4.18;
```
sohkai: I think this should be left unpinned; it's subclassed by `AragonApp`.

contracts/common/Initializable.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.18;
+pragma solidity 0.4.18;
```
sohkai: This should be left unpinned; it's subclassed by `AragonApp`.

contracts/evmscript/EVMScriptRegistry.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity 0.4.18;
+pragma solidity ^0.4.18;
```
sohkai: I think this should be kept pinned; it's a pretty complicated contract.

`EVMScriptRunner` was left unpinned as it's subclassed by `AragonApp`.

verdverm:
I noticed some of these inconsistencies, and I am still getting up to speed with Solidity dev and such. What do you think about ...?

- getting a second opinion on which one's need to be fixed to get everything straight
- getting the current contracts in the appropriate state
- documenting the guidelines for future development so things stay in line

sohkai:
Agreed on all :).

> documenting the guidelines for future development so things stay in line

This would be awesome in a dev guide or similar! I've made https://github.com/aragon/aragonOS/issues/240 for this, feel free to add any more to the list!

sohkai:
@izqui please re-review.

[Unpins](https://github.com/aragon/aragonOS/pull/236/commits/0ee4dd283c0ce243cee4759d8f955d2092357ff0) all the contracts in `AragonApp`'s dependency graph.

PR #239, opened 2018-02-23
ITokenController: remove constant modifier

sohkai:
`onTransfer()` and `onApprove()` are arbitrary and should be allowed to modify state (e.g. keeping a list of transferees).

Needed for https://github.com/aragon/aragon-apps/pull/104.

### DIFF
--- a/contracts/lib/minime/ITokenController.sol
+++ b/contracts/lib/minime/ITokenController.sol
@@ -15,7 +15,7 @@ interface ITokenController {
     /// @param _to The destination of the transfer
     /// @param _amount The amount of the transfer
     /// @return False if the controller does not authorize the transfer
-    function onTransfer(address _from, address _to, uint _amount) public constant returns(bool);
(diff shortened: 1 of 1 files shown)


verdverm:
I was informed recently that the `lib` directory is a special code base and had to revert some changes.  See: https://github.com/aragon/aragonOS/pull/233

@izqui What is the change process for code in the lib directory? Where is it copied from?

izqui:
@verdverm is right. @sohkai is there any reason we need to be changing this, is anything not working properly?

sohkai:
See https://github.com/aragon/aragon-apps/pull/104/files#r170335211.

Unfortunately, our `TokenManager` app may write to state during `onTransfer` so without this change, it'll fail to compile.

In general though, I find it restricting for callbacks to forced `constant` from their interfaces, since you'll never know what someone might want to do in that callback (e.g. log holders in our case).

PR #242, opened 2018-03-02
Add SafeMath for uint64

bingen:
Needed for Voting and Finance apps

### DIFF
--- a/contracts/lib/zeppelin/math/SafeMath64.sol
+++ b/contracts/lib/zeppelin/math/SafeMath64.sol
@@ -1,0 +1,32 @@
+pragma solidity ^0.4.11;
+
+
+/**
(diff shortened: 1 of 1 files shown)


bingen:
I'll do.

sohkai:
Mmm, since we don't use this in aragonOS, shouldn't the apps independently introduce this?

bingen:
Oops, a little bit late sorry, I even did the release! ;-)
Well, I don't know but I think having both versions (256 and 64 bits) together is a good thing. Besides, aragonOS (or any other repo that has it as a dependency) may use it eventually, and aragonOS is the most basic repo.
Anyway, I'm fine with any option, if you want me to move it I can do it.

PR #243, opened 2018-03-05
Minor changes for deployment with docker 

izqui:


package.json
```diff
@@ -13,6 +14,8 @@
     "migrate:dev": "npm run ganache-cli:dev && npm run migrate:dev:contracts",
     "migrate:dev:contracts": "truffle migrate --all --network rpc;",
     "generate:artifacts-dev": "npm run migrate:dev; truffle exec --network rpc scripts/generate_artifacts.js",
+    "deploy:devnet:ens": "truffle compile; truffle exec --network devnet scripts/deploy-beta-ens.js",
+    "deploy:devnet:apm": "truffle compile; truffle exec --network devnet scripts/deploy-beta-apm.js",
```
bingen: Did you add `compile` before in order to use it here?

scripts/deploy-beta-apm.js
```diff
@@ -52,6 +52,7 @@ module.exports = async callback => {
   console.log('=========')
   const apmAddr = receipt.logs.filter(l => l.event == 'DeployAPM')[0].args.apm
   console.log('deployed APM:', apmAddr)
+  console.log(apmAddr)
```
sohkai: What is the point of logging this again (and in [the ENS script](https://github.com/aragon/aragonOS/pull/243/files#diff-1fb706e498588d59d326e44987cc814dR15))?

Should we return the value instead so users of the script can have access to the deployed address?

izqui: So we can easily pipe the value in a bash script https://github.com/aragon/aragon-apps/pull/115/files#diff-273eaaae9a8f6fc623375ffa23d33531R6

PR #246, opened 2018-03-13
Bye WETH

### LINKED ISSUE #245: Deprecate EtherToken
I was reviewing BCC's chief evangelist (@decanus) code for Dexy and I was really surprised at how clean a Vault that handles ETH and tokens can be made, making the UX actually better than wrapping and unwrapping tokens. From an architectural design point of view it is kind of elegant to just handle tokens, but given how clean this looks I'm really having second thoughts.

Pasting some of Dean's code here:
```solidity
address constant public ETH = 0x0;

// ETH and erc20
function deposit(address token, uint amount) external payable {
        require(token == ETH || msg.value == 0);

        uint value = amount;
        if (token == ETH) {
            value = msg.value;
        }

        if (token != ETH) {
            require(ERC20(token).transferFrom(msg.sender, address(this), value));
        }

        depositFor(msg.sender, token, value); // pretty much just accounting
}

// erc777
function tokenFallback(address from, uint value, bytes data) public {
        depositFor(from, msg.sender, value);
}

function withdraw(address token, uint amount) external {
        require(balanceOf(token, msg.sender) >= amount);

        balances[token][msg.sender] = balances[token][msg.sender].sub(amount);
        accounted[token] = accounted[token].sub(amount);

        if (token == ETH) {
            msg.sender.transfer(amount);
        } else if (isERC777[token]) {
            ERC777(token).send(msg.sender, amount);
        } else {
            ERC20(token).transfer(msg.sender, amount);
        }

        Withdrawn(msg.sender, token, amount);
}

```

The Vault is not something we should be updating frequently, so I would like to push this for the new audit and have WHG review it too.

@aragon/aragon-core please give your opinion ASAP.

izqui:
> It is not you, it's us.

Close #245


package.json
```diff
@@ -47,6 +46,7 @@
     "truffle-hdwallet-provider": "0.0.3"
   },
   "dependencies": {
+    "homedir": "^0.6.0",
```
sohkai: wat

onbjerg: 🤔

contracts/common/EtherToken.sol
```diff
@@ -1,51 +0,0 @@
-pragma solidity 0.4.18;
-
-import "../lib/erc677/ERC677Token.sol";
```
bingen: Is this used anywhere else? Can't we remove `ERC677Token` too?

package.json
izqui: idk

izqui: wish i remembered

izqui: wish i remembered

izqui: l think related to being able to use it from aragon-apps, whose truffle.js just requires aOS truffle.js

decanus:
RIP Weth

onbjerg:
the weth is not anymore like it used to be

izqui:
he lived a happy life, now in a happier place

package.json
sohkai: Maybe we should move the truffle config out into a package if we're going to reuse it like this.

PR #248, opened 2018-03-14
delegatecall safety: Implement min return size in DelegateProxy

izqui:
Adds a new param to `delegateFwd(...)` to pass the minimum amount of bytes the underlying call must return, otherwise, it reverts. I kept the old function as a convenince method that just passes `0` as the minimum return data size.

This allows things such as always forcing to return something when calling a specific type of contract to avoid a `selfdesctruct`. Given that when a `selfdestruct` occurs the context is exited returning `1` (success) but always returns 0 bytes of data. Ideally this would be done at the protocol level, returning a different error code when it `selfdestruct`s but we can't wait.

This only works if you force the functions you forward to return something (setting the min to 32 and returning a boolean is enough).

### DIFF
--- a/test/TestDelegateProxy.sol
+++ b/test/TestDelegateProxy.sol
@@ -1,0 +1,75 @@
+pragma solidity 0.4.18;
+
+import "truffle/Assert.sol";
+import "./helpers/ThrowProxy.sol";
+
+import "../contracts/common/DelegateProxy.sol";
+import "../contracts/evmscript/ScriptHelpers.sol";
+
+
+contract Target {
+    function returnSomething() public pure returns (bool) { return true; }
+    function dontReturn() {}
+    function fail() { revert(); }
+    function die() { selfdestruct(0); }
+}
+
+
+contract TestDelegateProxy is DelegateProxy {
+    using ScriptHelpers for *;
+
+    Target target;
+    ThrowProxy throwProxy;
+
+    function beforeAll() {
+        target = new Target();
+    }
+
+    function beforeEach() {
+        throwProxy = new ThrowProxy(address(this));
+    }
+
+    function testMinReturn0WithoutReturn() {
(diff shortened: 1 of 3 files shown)


test/TestDelegateProxy.sol
```diff
@@ -0,0 +1,75 @@
...
+        TestDelegateProxy(throwProxy).revertIfReturnLessThanMinAndDie();
+        throwProxy.assertThrows("should have reverted to stop selfdestruct");
+    }
+
+    function revertIfReturnLessThanMinAndDie() {
+        delegateFwd(target, target.die.selector.toBytes(), 32);
+    }
+
+    function testFailIfReverts() {
+        TestDelegateProxy(throwProxy).revertIfReturnLessThanMinAndDie();
```
bingen: Shouldn't this be `revertCall`?

test/TestDelegateProxy.sol
```diff
@@ -0,0 +1,75 @@
...
+    function beforeEach() {
+        throwProxy = new ThrowProxy(address(this));
+    }
+
+    function testMinReturn0WithoutReturn() {
+        delegateFwd(target, target.dontReturn.selector.toBytes(), 0);
+    }
+
+    function testMinReturn0WithReturn() {
+        delegateFwd(target, target.returnSomething.selector.toBytes(), 0);
```
sohkai: For these that don't fail, it'd be great if we could assert that their returns are the same as what `Target` should've returned (or nothing). But doing this would probably be easier in a separate JS-based test for the `DelegateProxy`.

test/TestDelegateProxy.sol
izqui: Precisely

test/TestDelegateProxy.sol
izqui: yep

PR #249, opened 2018-03-20
Remove docker deploy image for beta templates in apps

bingen:


bingen:
Needed for
https://github.com/aragon/aragon-apps/pull/160

PR #255, opened 2018-03-25
Require minimum return data from EVMScript executors to prevent selfdestructs

### LINKED ISSUE #250: Prevent EVMScript runners from self destructing apps

izqui:
Close #250


izqui:
CI doing stupid stuff. Tests are passing locally

bingen:
Yes, I had the exact same problem a while ago with one of my PRs. Following your advice and just restarting it did the trick ;-)

test/evm_script.js
```diff
@@ -279,5 +290,29 @@ contract('EVM Script', accounts => {
...
+                dyingExecutor = await MockDyingScriptExecutor.new()
+                await reg.addScriptExecutor(dyingExecutor.address, { from: boss })
+            })
+
+            it('registers new executor', async () => {
+                assert.equal(await reg.getScriptExecutor('0x00000004'), dyingExecutor.address)
+            })
+
+            it('executes mock executor', async () => {
+                await executor.execute('0x00000004')
```
sohkai: Maybe this could check that the result is actually `bytes(1)`?

izqui: Executor.execute doesn't return anything, it would require some changes

izqui: Executor.execute doesn't return anything, it would require some changes

sohkai: Ah damn, ok. Let's not worry about it then.

PR #257, opened 2018-03-26
Fix storage shadowing in DeployDelegateScript cache

### LINKED ISSUE #254: DeployDelegateScript's cache can shadow AppStorage
Because the executor ends up being delegate called from the app, the cache for the executor needs to be properly offset, right now it is at storage slot 0. This is not a big problem given that it is a mapping, but it is quite dirty.

izqui:
Close #254


sohkai:
Wondering if it'd be better to use one of the reserved slots in `AppStorage` for this instead, so users of `AragonApp` won't collide with it (however improbable using slot 2^254 + 1 may be).

PR #258, opened 2018-03-26
Add modifier to check for already initialized apps

bingen:
Required by https://github.com/aragon/aragon-apps/issues/188

### DIFF
--- a/test/kernel_apps.js
+++ b/test/kernel_apps.js
@@ -132,10 +132,22 @@ contract('Kernel apps', accounts => {
                     await kernel.setApp(APP_BASE_NAMESPACE, appId, appCode1.address)
                 })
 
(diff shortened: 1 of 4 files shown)


test/kernel_apps.js
```diff
@@ -132,12 +132,24 @@ contract('Kernel apps', accounts => {
                     await kernel.setApp(APP_BASE_NAMESPACE, appId, appCode1.address)
                 })
 
+                it('fails calling function with isInitialized (if it\'s not)', async () => {
```
sohkai: Useful to note, you can use double quotes `"` to avoid the `\'` :).

bingen: Haha, yes I know, I was just trying to be consistent with the rest of the file (and unsure of if there is any guideline about the use of single vs double quotes)

PR #261, opened 2018-03-26
Fix ACLOracle calls not to fail if underlying call reverts

### LINKED ISSUE #252: Ensure ACL Oracle param returns false rather than reverting if call reverts
This will require making a raw call to the address rather than a solidity call

izqui:
Close #252 

I really do not understand why it isn't working and not copying the return data correctly 🤷‍♀️


izqui:
This is ready for review now @sohkai

contracts/acl/ACL.sol
```diff
@@ -320,6 +320,34 @@ contract ACL is IACL, AragonApp, ACLHelpers {
...
+        if (size != 32) {
+            return false;
+        }
+
+        bool result;
+        assembly {
+            let ptr := mload(0x40)
+            returndatacopy(ptr, 0, size)
+            result := mload(ptr)
+            mstore(ptr, 0)
```
sohkai: Might be useful to add a few comments here to explain exactly what we're doing (copying the return into memory, loading the memory, and then clearing the memory).

contracts/acl/ACL.sol
```diff
@@ -320,6 +320,34 @@ contract ACL is IACL, AragonApp, ACLHelpers {
         return false;
     }
 
+    function checkOracle(address _oracleAddr, address _who, address _where, bytes32 _what, uint256[] _how) internal view returns (bool) {
+        bytes4 sig = ACLOracle(_oracleAddr).canPerform.selector;
+
+        // a raw call is required so we can return false if the call reverts, rather than reverting
+        bool ok = _oracleAddr.call(sig, bytes32(_who), bytes32(_where), _what, 0x80, _how.length, _how);
```
sohkai: I think the `.call()` pads the addresses for you, but maybe they pad them in the wrong direction?

izqui: Seems to be working fine, removing the explicit casting :)

PR #264, opened 2018-03-26
acl: Add better getters for extracting ACL info

### LINKED ISSUE #251: Add better getters for extracting ACL info

bingen:
Closes #251.

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -137,6 +137,32 @@ contract ACL is IACL, AragonApp, ACLHelpers {
     }
 
     /**
+     * @notice Get parameters for permission array length
+     * @param _entity Address of the whitelisted entity that will be able to perform the role
+     * @param _app Address of the app
+     * @param _role Identifier for a group of actions in app
+     * @return Length of the array
+     */
+    function getPermissionParamsLength(address _entity, address _app, bytes32 _role) external view returns (uint) {
+        return permissionParams[permissions[permissionHash(_entity, _app, _role)]].length;
+    }
+
+    /**
+    * @notice Get parameters for permission
+    * @param _entity Address of the whitelisted entity that will be able to perform the role
+    * @param _app Address of the app
+    * @param _role Identifier for a group of actions in app
+    * @param _index Index of parameter in the array
+    * @return Parameter (id, op, value)
+    */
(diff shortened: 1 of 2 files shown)


test/kernel_acl.js
```diff
@@ -103,6 +103,15 @@ contract('Kernel ACL', accounts => {
...
 
             const r1 = await acl.grantPermissionP(accounts[3], app, role, [param], { from: granted })
+
+            // retrieve the params back with the getters
+            const numParams = await acl.getPermissionParamsLength(accounts[3], app, role)
+            assert.equal(numParams, 1, 'There should be just 1 param')
+            const returnedParam = await acl.getPermissionParams(accounts[3], app, role, 0)
+            assert.equal(returnedParam[0].valueOf(), parseInt(argId), 'param id should match')
+            assert.equal(returnedParam[1].valueOf(), parseInt(op), 'param op should match')
+            assert.equal(returnedParam[2].valueOf(), parseInt(value), 'param value should match')
```
sohkai: We should be careful with the uses of `parseInt`, based on the string representations for each variable:

- `argId` should use `parseInt(argId, 16)`
- `op` should use `parseInt(op, 10)`
- `value` should use `parseInt(value, 10)`

contracts/acl/ACL.sol
```diff
@@ -136,6 +136,32 @@ contract ACL is IACL, AragonApp, ACLHelpers {
...
+
+    /**
+    * @notice Get parameters for permission
+    * @param _entity Address of the whitelisted entity that will be able to perform the role
+    * @param _app Address of the app
+    * @param _role Identifier for a group of actions in app
+    * @param _index Index of parameter in the array
+    * @return Parameter (id, op, value)
+    */
+    function getPermissionParams(address _entity, address _app, bytes32 _role, uint _index) external view returns (uint8 id, uint8 op, uint240 value) {
```
sohkai: Since we can only get a single param at a time, might make more sense to drop the plurals (e.g. `getPermissionParam`).

bingen: Makes sense.

test/kernel_acl.js
bingen: I don't think it would work with different encodings but it doesn't hurt (and it's probably clearer for hex case), so changing it.

PR #267, opened 2018-03-26
Make AppProxies and KernelProxy ERC897 compatible

### LINKED ISSUE #238: Make AppProxies and KernelProxy ERC897 compatible
https://github.com/ethereum/EIPs/pull/897

bingen:
Closes #238.

### DIFF
--- a/contracts/apps/AppProxyBase.sol
+++ b/contracts/apps/AppProxyBase.sol
@@ -1,13 +1,12 @@
 pragma solidity 0.4.18;
 
-import "./IAppProxy.sol";
 import "./AppStorage.sol";
 import "../common/DelegateProxy.sol";
 import "../kernel/KernelStorage.sol";
 import "../lib/misc/ERCProxy.sol";
 
 
-contract AppProxyBase is IAppProxy, AppStorage, DelegateProxy, KernelConstants, ERCProxy {
+contract AppProxyBase is ERCProxy, AppStorage, DelegateProxy, KernelConstants {
     /**
     * @dev Initialize AppProxy
     * @param _kernel Reference to organization kernel for the app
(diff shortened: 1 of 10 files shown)


contracts/lib/misc/ERCProxy.sol
```diff
@@ -0,0 +1,7 @@
+pragma solidity ^0.4.18;
+
+
+interface ERCProxy {
```
sohkai: Although ideally this would be an `interface`, I think having `FORWARDING` and `UPGRADEABLE` being in this contract is more valuable than trying to keep this as a pure `interface`.

test/kernel_upgrade.js
```diff
@@ -30,6 +30,14 @@ contract('Kernel Upgrade', accounts => {
         kernelId = await kernel.KERNEL_APP_ID()
     })
 
+    it('checks ERC897 functions', async () => {
+        const kernelProxy = KernelProxy.at(app)
+        const implementation = await kernelProxy.implementation()
+        assert.equal(implementation, kernelBase.address, "App address should match")
+        const proxyType = await kernelProxy.proxyType()
+        assert.equal(proxyType, 2, "Proxy type should be 2 (upgradeable)")
```
sohkai: Rather than using the constant `1` or `2` here and elsewhere, maybe we could use `ERCProxy.FORWARDING()` and `ERCProxy.UPGRADABLE()`?

contracts/apps/AppProxyBase.sol
```diff
@@ -32,6 +33,20 @@ contract AppProxyBase is IAppProxy, AppStorage, DelegateProxy, KernelConstants {
...
+     * @dev ERC897, whether it is a forwarding (1) or an upgradeable (2) proxy
+     */
+    function proxyType() public pure returns (uint256 proxyTypeId) {
+        return isUpgradeable() ? UPGRADEABLE : FORWARDING;
+    }
+
+    /**
+     * @dev ERC897, the address the proxy would delegate calls to
+     */
+    function implementation() public view returns (address) {
```
izqui: We could save having one extra function if `APPinned` and `APUpgradeable` just implement `implementation()` rather than `getCode()` directly!

bingen: Makes sense. We can actually do the same with `proxyType` and even remove `IAppProxy` interface, as it's essentially the same as the ERC 897 one.

sohkai: > We can actually do the same with proxyType and even remove IAppProxy interface, as it's essentially the same as the ERC 897 one.

👌 very nice!

PR #271, opened 2018-03-26
DAOFactory: remove unnecessary ternary

sohkai:
> Currently there is a superfluous ternary being used in the DAOFactory.sol function newDAO when only 1 conditinal is necessary.

### DIFF
--- a/contracts/factory/DAOFactory.sol
+++ b/contracts/factory/DAOFactory.sol
@@ -32,7 +32,9 @@ contract DAOFactory {
     function newDAO(address _root) public returns (Kernel dao) {
         dao = Kernel(new KernelProxy(baseKernel));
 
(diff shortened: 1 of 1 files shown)


contracts/factory/DAOFactory.sol
```diff
@@ -55,6 +53,8 @@ contract DAOFactory {
 
             acl.setPermissionManager(address(0), dao, appManagerRole);
             acl.setPermissionManager(_root, acl, permRole);
+        } else {
+            dao.initialize(baseACL, _root);
```
izqui: I like it! Given the difference between the size of the code blocks, I would swap them: 
```solidity
if (address(regFactory) == address(0)) {
  dao.initialize(baseACL, _root);
} else {
  ...
}
```

PR #272, opened 2018-03-26
AppProxyBase: always check that appCode is a contract

sohkai:
> In AppProxyBase.sol require is contract should come before if _initializePayload.length > 0

------------

~~With the changes I had to make (making sure `setApp` is done before `newAppProxy()`), I'm not 100% sure about keeping this.~~

~~On the one hand, it is safer to know that all app proxy instances are connected to real app code in their kernels, on the other, it forces you to always have that app code before instantiating the proxy instances (but I think we probably want this).~~

Edit: reverted original 2e7dfad change and instead made `AppProxyPinned` check if the initial address is a contract. From the tests, it looks like `AppProxyUpgradeable` was, by design, allowed to not have a real contract backing it initially.


sohkai:
Supercede by #278; we now check if given app addresses are contracts when we set them, so there's no need to do it again when we create app instances.

PR #276, opened 2018-03-26
ACL: add removePermissionManager()

### LINKED ISSUE #266: Add removePermissionManager
We currently use `acl.setPermissionManager(address(0), <app>, <role>);` but this is non-obvious and leaky (users would have to know that creating permissions is only possible if the current permission manager is `address(0)` rather than there being no permission manager).

sohkai:
Closes #266.

### DIFF
--- a/test/kernel_acl.js
+++ b/test/kernel_acl.js
@@ -180,6 +180,7 @@ contract('Kernel ACL', accounts => {
 
         context('transferring managership', () => {
             const newManager = accounts[8]
+            assert.notEqual(newManager, granted, 'newManager should not be the same as granted')
 
             beforeEach(async () => {
                 const receipt = await acl.setPermissionManager(newManager, app, role, { from: granted })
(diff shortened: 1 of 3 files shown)


test/kernel_acl.js
```diff
@@ -194,6 +195,36 @@ contract('Kernel ACL', accounts => {
...
+                assertEvent(createReceipt, 'SetPermission')
+                assertEvent(createReceipt, 'ChangePermissionManager')
+
+                const grantReceipt = await acl.grantPermission(granted, app, role, { from: newManager })
+                assertEvent(grantReceipt, 'SetPermission')
+            })
+
+            it('old manager lost power', async () => {
+                return assertRevert(async () => {
+                    await acl.grantPermission(newManager, app, role, { from: granted })
```
bingen: Wouldn't it be better to use another account rather than `newManager` here? Otherwise there are actually 2 reasons to fail here: `granted` not being the manager and `newManager` already having the permission, right?

sohkai: Good point!

sohkai:
@izqui More explicit and decoupling details from the user (it takes _a lot_ of digging to figure out we use `address(0)` to reset the manager and reallow the permission to be created if you're new).

PR #278, opened 2018-03-27
Make sure given address is a contract in `kernel.setApp()`

### LINKED ISSUE #277: Make sure given address is a contract in `kernel.setApp()`
`setApp` and `setAppIfNew()` currently just sets the given address without doing any sanity checks.

It would be safer if they did check the given address was actually a contract before allowing it to be set as an app.

bingen:
Closes #277.


bingen:
I think the problem is here:
https://coveralls.io/builds/16199567/source?filename=contracts%2Fapps%2FAppProxyPinned.sol#L17
Now `pinnedCode` won't be 0 any more. I'm going to check if there's a way to force it to 0. Otherwise, should we remove that check?

test/kernel_apps.js
```diff
@@ -314,8 +313,9 @@ contract('Kernel apps', accounts => {
             await kernel.setApp(APP_BASE_NAMESPACE, appId, appCode1.address)
             assert.equal(appCode1.address, await kernel.getApp(appSetId))
 
-            const appProxy = await kernel.newPinnedAppInstance(appId, '0x0')
-            assert.equal(appCode1.address, await kernel.getApp(appSetId))
+            return assertRevert(async () => {
+                const appProxy = await kernel.newPinnedAppInstance(appId, '0x0')
+            })
```
sohkai: For these new revert cases here and above for `.newAppInstance()`, it might be worth changing the name of the test to better reflect the test :).

sohkai:
> Otherwise, should we remove that check?

We need to keep it, in case you try to pin an app when its code doesn't exist yet. That should be the way to trigger the `== address(0)` check; try to make a pinned app on an id that doesn't have code set to it yet.

contracts/kernel/Kernel.sol
```diff
@@ -114,6 +114,10 @@ contract Kernel is IKernel, KernelStorage, Initializable, AppProxyFactory, ACLSy
     }
 
     function isContract(address _target) internal view returns (bool) {
+        if (_target == address(0)) {
```
sohkai: I think `extcodesize(0)` always produces 0 so this might not be necessary.

izqui: `extcodesize` costs 700 gas and that check is almost free :)

izqui:
Pls fix the conflict @bingen!

PR #279, opened 2018-03-27
Add address 0 check to isContract

bingen:
To save one extcodesize in that case.


bingen:
I'll check the coverage decrease here too.

PR #281, opened 2018-03-28
[WIP] Make send and transfer work across proxies

### LINKED ISSUE #275: Native `send()` and `transfer()` don't work across proxies
> Send or transfer only permits 2300 gas but delegate proxy requires 10000 + execution costs.

Some ideas:

- Override the native methods with our own, that increase the gas used
- Don't delegate in `DelegateProxy` under certain conditions, and log an event (I like this one better):
    - `if (msg.gas < 10k) return require(msg.value > 0 && msg.data.length == 0)`

bingen:
Besides, add function to recover funds.

Closes #275.

### DIFF
--- a/contracts/kernel/Kernel.sol
+++ b/contracts/kernel/Kernel.sol
@@ -73,6 +73,25 @@ contract Kernel is IKernel, KernelStorage, Initializable, AppProxyFactory, ACLSy
     }
 
     /**
+    * @dev Get the address of the default Vault instance (to recover funds)
+    * @return Address of the Vault
+    */
+    function getDefaultVault() public view returns (address) {
+        return apps[defaultVaultId];
+    }
+
+    /**
+    * @dev Set the resolving address of an app instance or base implementation
+    * @param _namespace App namespace to use
+    * @param _name Name of the app
+    * @return ID of app
+    */
+    function setDefaultVaultId(bytes32 _namespace, bytes32 _name) auth(APP_MANAGER_ROLE, arr(_namespace, _name)) kernelIntegrity public returns (bytes32) {
+        defaultVaultId = keccak256(_namespace, _name);
+        return defaultVaultId;
+    }
+
+    /**
     * @dev Get the installed ACL app
     * @return ACL app
     */
--- a/contracts/kernel/KernelStorage.sol
+++ b/contracts/kernel/KernelStorage.sol
@@ -11,9 +11,12 @@ contract KernelConstants {
 
     bytes32 constant public ACL_APP_ID = keccak256("acl.aragonpm.eth");
     bytes32 constant public ACL_APP = keccak256(APP_ADDR_NAMESPACE, ACL_APP_ID);
+
+    address constant public ETH = 0x0;
 }
 
 
 contract KernelStorage is KernelConstants {
     mapping (bytes32 => address) public apps;
+    bytes32 public defaultVaultId;
 }
--- a/test/proxy_funds.js
+++ b/test/proxy_funds.js
@@ -1,0 +1,88 @@
+const { assertRevert } = require('./helpers/assertThrow')
+const { getBalance } = require('./helpers/web3')
+const { hash } = require('eth-ens-namehash')
+
+const Kernel = artifacts.require('Kernel')
+const AppProxyUpgradeable = artifacts.require('AppProxyUpgradeable')
+const AppStub = artifacts.require('AppStub')
+const DAOFactory = artifacts.require('DAOFactory')
+const ACL = artifacts.require('ACL')
+
+const getContract = artifacts.require
+
+const keccak256 = require('js-sha3').keccak_256
+const APP_BASE_NAMESPACE = '0x'+keccak256('base')
+
+contract('Proxy funds', accounts => {
+  let factory, acl, kernel, app, appCode, appProxy, vault
+
+  const permissionsRoot = accounts[0]
+  const appId = hash('stub.aragonpm.test')
+  const zeroAddr = '0x0000000000000000000000000000000000000000'
+  let ETH
+
+  before(async () => {
+    const kernelBase = await getContract('Kernel').new()
+    ETH = await kernelBase.ETH()
+    const aclBase = await getContract('ACL').new()
+    factory = await DAOFactory.new(kernelBase.address, aclBase.address, '0x00')
+
+    appCode = await AppStub.new()
+
+    const receipt = await factory.newDAO(permissionsRoot)
+    const kernelAddress = receipt.logs.filter(l => l.event == 'DeployDAO')[0].args.dao
+
+    kernel = Kernel.at(kernelAddress)
+    acl = ACL.at(await kernel.acl())
+
+    const r = await kernel.APP_MANAGER_ROLE()
+    await acl.createPermission(permissionsRoot, kernel.address, r, permissionsRoot)
+
+    // app
+    await kernel.setApp(APP_BASE_NAMESPACE, appId, appCode.address)
+    const initializationPayload = appCode.contract.initialize.getData()
+    appProxy = await AppProxyUpgradeable.new(kernel.address, appId, initializationPayload, { gas: 5e6 })
+    app = AppStub.at(appProxy.address)
+    // vault
+    vault = await getContract('VaultMock').new()
+    const vaultId = hash('vault.aragonpm.test')
+    await kernel.setApp(APP_BASE_NAMESPACE, vaultId, vault.address)
+    await kernel.setDefaultVaultId(APP_BASE_NAMESPACE, vaultId)
+  })
+
+  it('recovers ETH', async() => {
+    const amount = 1
+    const initialAppBalance = await getBalance(appProxy.address)
+    const initialVaultBalance = await getBalance(vault.address)
+    const r = await appProxy.sendTransaction({ value: 1, gas: 25000 })
+    assert.equal((await getBalance(appProxy.address)).valueOf(), initialAppBalance.plus(amount))
+    await appProxy.transferToVault(ETH)
+    assert.equal((await getBalance(appProxy.address)).valueOf(), 0)
+    assert.equal((await getBalance(vault.address)).valueOf(), initialVaultBalance.plus(initialAppBalance).plus(amount).valueOf())
+  })
+
+  it('recovers tokens', async () => {
+    const amount = 1
+    const token = await getContract('StandardTokenMock').new(accounts[0], 1000)
+    const initialAppBalance = await token.balanceOf(appProxy.address)
+    const initialVaultBalance = await token.balanceOf(vault.address)
+    await token.transfer(appProxy.address, amount)
+    assert.equal((await token.balanceOf(appProxy.address)).valueOf(), initialAppBalance.plus(amount))
+    await appProxy.transferToVault(token.address)
+    assert.equal((await token.balanceOf(appProxy.address)).valueOf(), 0)
+    assert.equal((await token.balanceOf(vault.address)).valueOf(), initialVaultBalance.plus(initialAppBalance).plus(amount).valueOf())
+  })
+
+  it('fails if vault is not contract', async() => {
+    const amount = 1
+    const vaultId = hash('vault.aragonpm.test')
+    const initialAppBalance = await getBalance(appProxy.address)
+    await kernel.setApp(APP_BASE_NAMESPACE, vaultId, '0x0')
+    const r = await appProxy.sendTransaction({ value: 1, gas: 25000 })
+    assert.equal((await getBalance(appProxy.address)).valueOf(), initialAppBalance.plus(amount))
+    return assertRevert(async () => {
+      await appProxy.transferToVault(ETH)
+    })
+  })
+
+})
--- a/contracts/apps/AppProxyBase.sol
+++ b/contracts/apps/AppProxyBase.sol
@@ -4,9 +4,29 @@ import "./IAppProxy.sol";
 import "./AppStorage.sol";
 import "../common/DelegateProxy.sol";
 import "../kernel/KernelStorage.sol";
+import "../lib/zeppelin/token/ERC20.sol";
 
 
 contract AppProxyBase is IAppProxy, AppStorage, DelegateProxy, KernelConstants {
+    event ProxyDeposit(address sender, uint256 value);
+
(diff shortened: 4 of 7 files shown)


contracts/kernel/Kernel.sol
```diff
@@ -72,6 +72,25 @@ contract Kernel is IKernel, KernelStorage, Initializable, AppProxyFactory, ACLSy
...
+        return apps[defaultVaultId];
+    }
+
+    /**
+    * @dev Set the resolving address of an app instance or base implementation
+    * @param _namespace App namespace to use
+    * @param _name Name of the app
+    * @return ID of app
+    */
+    function setDefaultVaultId(bytes32 _namespace, bytes32 _name) auth(APP_MANAGER_ROLE, arr(_namespace, _name)) kernelIntegrity public returns (bytes32) {
```
izqui: We should hardcode `namespace` to be the `APPS_NAMESPACE`

izqui: This will also remove the need to check `kernelIntegrity`

contracts/kernel/KernelStorage.sol
```diff
@@ -16,4 +16,5 @@ contract KernelConstants {
 
 contract KernelStorage is KernelConstants {
     mapping (bytes32 => address) public apps;
+    bytes32 public defaultVaultId;
```
izqui: Why do we need this extra storage?

bingen: To recover it from here:
https://github.com/aragon/aragonOS/pull/281/files#diff-14a0803379fe8527d47915e3727fe7d0R79
and here:
https://github.com/aragon/aragonOS/pull/281/files#diff-807af7e85f274f9138c01132b86256b6R19

contracts/kernel/Kernel.sol
```diff
@@ -72,6 +72,23 @@ contract Kernel is IKernel, KernelStorage, Initializable, AppProxyFactory, ACLSy
...
+        return apps[defaultVaultId];
+    }
+
+    /**
+    * @dev Set the resolving address of an app instance or base implementation
+    * @param _name Name of the app
+    * @return ID of app
+    */
+    function setDefaultVaultId(bytes32 _name) auth(APP_MANAGER_ROLE, arr(APP_BASES_NAMESPACE, _name)) public {
+        defaultVaultId = keccak256(APP_BASES_NAMESPACE, _name);
```
izqui: I think we should make this part of the initialization of the kernel. Set this to `defaultVaultId = keccak256(APP_BASES_NAMESPACE, namehash(`vault.aragonpm.eth`)`

contracts/common/FundsProxy.sol
```diff
@@ -0,0 +1,44 @@
...
+     * @notice Send funds to default Vault. This contract should never receive funds,
+     *         but in case it does, this function allows to recover them.
+     * @param _token Token balance to be sent to Vault. ETH(0x0) for ether.
+     */
+    function transferToVault(address _token) external {
+        address vault = getDefaultVault();
+        require(isContract(vault));
+
+        if (_token == ETH) {
+            vault.transfer(address(this).balance);
```
izqui: Maybe we could make this a proper vault transfer so it gets logged. We just need to convert it to `require(vault.call.value(address(this).balance)())`

contracts/common/FundsProxy.sol
```diff
@@ -0,0 +1,44 @@
...
+        } else {
+            uint256 amount = ERC20(_token).balanceOf(this);
+            ERC20(_token).transfer(vault, amount);
```
izqui: Also this should be a proper deposit, need to give it an extra thought about how we can support different deposit logic to the vault depending on the token standard. I would say lets leave it like this for now, but lets give it an extra thought

contracts/kernel/Kernel.sol
izqui: See #282, as the name should be the appId for the vault

izqui:
Looking good. @sohkai pls review

contracts/kernel/Kernel.sol
```diff
@@ -26,6 +26,8 @@ contract Kernel is IKernel, KernelStorage, Initializable, AppProxyFactory, ACLSy
         _setApp(APP_ADDR_NAMESPACE, ACL_APP_ID, acl);
 
         acl.initialize(_permissionsCreator);
+
+        defaultVaultId = keccak256(APP_BASES_NAMESPACE, apmNamehash('vault'));
```
sohkai: We should change this to use `APP_ADDR_NAMESPACE`.

test/TestDelegateProxy.sol
```diff
@@ -79,6 +76,17 @@ contract TestDelegateProxy is DelegateProxy {
         Assert.isFalse(result, "should return false");
     }
 
+    /* TODO
```
sohkai: Maybe add a comment saying that ganache doesn't like this bit :(.

test/proxy_funds.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  context('App Proxy', async () => {
+    let vault
+    before(async () => {
+      // vault
+      vault = await getContract('VaultMock').new()
+      const vaultId = hash('vault.aragonpm.test')
+      await kernel.setApp(APP_BASE_NAMESPACE, vaultId, vault.address)
+      await kernel.setDefaultVaultId(vaultId)
```
sohkai: These tests should make and use a proxy for the vault rather than the vault's base contract :).

contracts/common/FundsProxy.sol
```diff
@@ -0,0 +1,45 @@
+pragma solidity 0.4.18;
+
+import "./DelegateProxy.sol";
+import "../lib/zeppelin/token/ERC20.sol";
+import "../lib/misc/ERCProxy.sol";
+
+
+contract FundsProxy is DelegateProxy, ERCProxy {
+    address constant public ETH = 0x0;
```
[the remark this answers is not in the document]
izqui: also we should use `address(0)` everywhere as it is more descriptive.

contracts/kernel/Kernel.sol
izqui: That should have been `APP_ADDR_NAMESPACE`

izqui:
I got confused with the namespace for the `defaultVaultId` during my review.

By default on Kernel initialization we can set `vault.aragonid.eth` as the default Vault appId. Then when a vault is created and wants to be set as the default Vault it needs to be assigned to the `APP_ADDR_NAMESPACE`. 

We could add a `setDefault` boolean param to `newAppInstance` to standardize this:

```solidity
function newAppInstance(bytes32 _name, address _appBase, bool _setDefault) auth(APP_MANAGER_ROLE, arr(APP_BASES_NAMESPACE, _name)) public returns (ERCProxy appProxy) {
  _setAppIfNew(APP_BASES_NAMESPACE, _name, _appBase);
  appProxy = newAppProxy(this, _name);
  if (_setDefault) setApp(APP_ADDR_NAMESPACE, _name, appProxy);
}
```

By calling `setApp` directly and not the internal functions, we make sure the params are checked and it will only succeed if sender has permissions to set something to the namespace.

sohkai:
> We could add a setDefault boolean param to newAppInstance to standardize this

I like this.

bingen:
Yes, this is done in the last commit.

izqui:
Why is 'EtherTokenConstant` in `lib/`?

bingen:
Where do you want me to put it? IT's going to be used from aragon-apps too, (and how know where else eventually), so I thought a generic place would be good. But let me know which you think is the best place and I'll move it.

sohkai:
I'll move `'EtherTokenConstant` to `common/` since it'll get used by users.

contracts/kernel/Kernel.sol
```diff
@@ -26,30 +28,44 @@ contract Kernel is IKernel, KernelStorage, Initializable, AppProxyFactory, ACLSy
...
     *      its base implementation if it was not already set
     * @param _name Name of the app
     * @param _appBase Address of the app's base implementation
+    * @param _setDefault Whether the app proxy app is the default one.
+    *        Useful when the Kernel needs to know of an instance of a particular app,
+    *        like Vault for escape hatch mechanism.
     * @return AppProxy instance
     */
-    function newPinnedAppInstance(bytes32 _name, address _appBase) auth(APP_MANAGER_ROLE, arr(APP_BASES_NAMESPACE, _name)) public returns (ERCProxy appProxy) {
+    function newPinnedAppInstance(bytes32 _name, address _appBase, bool _setDefault) auth(APP_MANAGER_ROLE, arr(APP_BASES_NAMESPACE, _name)) public returns (ERCProxy appProxy) {
```
sohkai: @bingen I'm thinking it might be nice to have an overloaded version of `newAppInstance()` and `newPinnedAppInstance()` that just passes false to `setDefault` to avoid making this a breaking change.

contracts/common/VaultRecoverable.sol
```diff
@@ -0,0 +1,27 @@
...
+    function transferToVault(address _token) external {
+        address vault = getRecoveryVault();
+        require(isContract(vault));
+
+        if (_token == ETH) {
+            // solium-disable-next-line security/no-call-value
+            require(vault.call.value(address(this).balance)());
+        } else {
+            uint256 amount = ERC20(_token).balanceOf(this);
+            ERC20(_token).transfer(vault, amount);
```
izqui: This is fine in a ERC20 world, I'm working on how the vault can provide multi-standard deposit logic for contracts. 

Since this is no longer logic at the proxy level but upgradeable (right?) we can update in the future.

contracts/evmscript/executors/DelegateScript.sol
```diff
@@ -2,14 +2,15 @@ pragma solidity 0.4.18;
...
+import "../../common/IsContract.sol";
 
 
 interface DelegateScriptTarget {
     function exec() public returns (bool);
 }
 
 
-contract DelegateScript is IEVMScriptExecutor {
+contract DelegateScript is IEVMScriptExecutor, IsContract {
```
izqui: thx

contracts/kernel/Kernel.sol
```diff
@@ -5,13 +5,16 @@ import "./KernelStorage.sol";
...
+import "../common/VaultRecoverable.sol";
 import "../factory/AppProxyFactory.sol";
 
 
-contract Kernel is IKernel, KernelStorage, Initializable, AppProxyFactory, ACLSyntaxSugar {
+contract Kernel is IKernel, KernelStorage, Initializable, IsContract, AppProxyFactory, ACLSyntaxSugar, VaultRecoverable {
     // Hardocde constant to save gas
     //bytes32 constant public APP_MANAGER_ROLE = keccak256("APP_MANAGER_ROLE");
     bytes32 constant public APP_MANAGER_ROLE = 0xb6d92708f3d4817afc106147d969e229ced5c46e65e0a5002a0d391287762bd0;
+    bytes32 constant public DEFAULT_VAULT_ID = 0x4214e5fd6d0170d69ea641b5614f5093ebecc9928af51e95685c87617489800e;
```
izqui: Please lets add a comment here

contracts/kernel/Kernel.sol
```diff
@@ -28,30 +31,54 @@ contract Kernel is IKernel, KernelStorage, Initializable, AppProxyFactory, ACLSy
...
+    *        like Vault for escape hatch mechanism.
     * @return AppProxy instance
     */
-    function newAppInstance(bytes32 _name, address _appBase) auth(APP_MANAGER_ROLE, arr(APP_BASES_NAMESPACE, _name)) public returns (ERCProxy appProxy) {
+    function newAppInstance(bytes32 _name, address _appBase, bool _setDefault) auth(APP_MANAGER_ROLE, arr(APP_BASES_NAMESPACE, _name)) public returns (ERCProxy appProxy) {
         _setAppIfNew(APP_BASES_NAMESPACE, _name, _appBase);
         appProxy = newAppProxy(this, _name);
+        // By calling setApp directly and not the internal functions, we make sure the params are checked
+        // and it will only succeed if sender has permissions to set something to the namespace.
+        if (_setDefault) setApp(APP_ADDR_NAMESPACE, _name, appProxy);
```
izqui: I love how this looks but we need to be consistent with the agreed style:

```solidity
if (x) {
  y()
}

File: contracts/kernel/Kernel.sol [hunk 15]
```diff
@@ -28,30 +31,54 @@ contract Kernel is IKernel, KernelStorage, Initializable, AppProxyFactory, ACLSy
...
+    }
+
+    function newPinnedAppInstance(bytes32 _name, address _appBase) auth(APP_MANAGER_ROLE, arr(APP_BASES_NAMESPACE, _name)) public returns (ERCProxy appProxy) {
```
izqui: This function needs proper comments :)

test/proxy_funds.js
```diff
@@ -0,0 +1,165 @@
+const { assertRevert } = require('./helpers/assertThrow')
```
izqui: Can we rename the test file to something a bit more descriptive like `proxy_recover_funds`?

contracts/common/VaultRecoverable.sol
```diff
@@ -0,0 +1,27 @@
...
+     *         but in case it does, this function allows one to recover them.
+     * @param _token Token balance to be sent to recovery vault.
+     */
+    function transferToVault(address _token) external {
+        address vault = getRecoveryVault();
+        require(isContract(vault));
+
+        if (_token == ETH) {
+            // solium-disable-next-line security/no-call-value
+            require(vault.call.value(address(this).balance)());
```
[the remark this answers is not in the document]
bingen: Ok, changing it, but better `transfer` than `send`, right?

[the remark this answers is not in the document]
bingen: Ok, changing it, but should I rather use `transfer`?

contracts/common/VaultRecoverable.sol
bingen: `VaultRecoverable` is inherited in `AragonApp` and `Kernel`, so yes, it should be upgradeable.

contracts/kernel/Kernel.sol
bingen: I'm fine with any standard you like, but I think it should be documented somewhere

PR #284, opened 2018-04-09
Fixes #221 : Change indentation of '.sol' files from 4 to 2 spaces

### LINKED ISSUE #221: Change indentation from 4 to 2 spaces

smitrajput:
Indentation of size 4 spaces for a tab, occurred in almost all the '.sol' files and in no '.js' files.

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -6,403 +6,403 @@ import "./IACL.sol";
 
 
 interface ACLOracle {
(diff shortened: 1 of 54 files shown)


sohkai:
We ended up deciding to stay with 4 spaces, as that is the solidity convention and there would be a lot of merge conflicts resulting from trying to apply these changes.

Sorry for being so late with responding to this (we were always waiting to be in a "good state" to merge a change like this), and for not merging it in the end!

PR #286, opened 2018-04-13
Fix gas issues

bingen:
Fixes `npm run test`.

### DIFF
--- a/contracts/factory/APMRegistryFactory.sol
+++ b/contracts/factory/APMRegistryFactory.sol
@@ -16,7 +16,7 @@ contract APMRegistryFactory is APMRegistryConstants {
     ENSSubdomainRegistrar public ensSubdomainRegistrarBase;
     ENS public ens;
 
-    event DeployAPM(bytes32 indexed node, address apm);
+    event DeployAPM(bytes32 indexed node, address apm, address acl);
 
     // Needs either one ENS or ENSFactory
     function APMRegistryFactory(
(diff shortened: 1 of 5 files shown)


izqui:
I understand that for gas reasons we are now requiring 2 txs to the factory to completely set up an APMRegistry? 

Can you explain a bit what the rationale behind the changes is and how it impacts the entire system? Will this have repercussions with the scripts in this repo or aragon-apps?

izqui:
Also how much gas does this need? Is it above 7m? If that's not the case, can we workaround that by explicitly specifying the gas amount in the problematic function?

bingen:
Yes, indeed I split `newAPM` into 2 transactions to avoid gas limit issues, and changed all affected calls everywhere in the repo. I was planning to adapt templates in aragon-apps repo in case/when this was merged.
About the gas, yes, it's about 7.7m. I see in [ethstats](https://ethstats.net/) that now is at about 8m so I guess it's safe to increase the limit above 7.7m. I'll open a new PR and close this one.

bingen:
Replaced by #288

sohkai:
I'm starting to think this is still necessary. With the changes in https://github.com/aragon/aragonOS/pull/294, it looks like we're pushing > 8m gas already.

It's probably best to split this out so that we're safe later on :).

bingen:
Ok, I'm reopening it and closing #288. Maybe we should close #292 too, or keep gas under Rinkeby limit at least.
Review then please. The main problem I see here is that if the second transaction fails, the APM would be created but with the wrong permissions. We should be careful when deploying.

bingen:
Oh, I did a git force push to amend commit message and now it doesn't allow me to reopen, so I had to recreate. New one here: #295

PR #288, opened 2018-04-15
Fix gas issues for newAPM

bingen:


bingen:
Closed in favour of #295

PR #291, opened 2018-04-15
Tests: restart ganache every test

sohkai:
It's not as necessary in aragonOS, but I thought we mind as well restart a clean instance of ganache on every test.

Was there a reason we didn't do this here though, @bingen (I think you added this to aragon-apps?)


bingen:
Btw, I think it would be a good idea to kill ganache after the test, as we are always starting it, to avoid leaving a hanging process in th background (this is done in aragon-apps too, although it could be improved for the case the test fails)

PR #292, opened 2018-04-15
Tests: increase gas limit in truffle config

sohkai:
So I found this fixes `npm run test`, but not the coverage locally :(.

Related to #286, #288.


sohkai:
Closing for #297.

PR #293, opened 2018-04-16
Fix compile warnings

sohkai:


contracts/evmscript/executors/CallsScript.sol
```diff
@@ -49,8 +49,8 @@ contract CallsScript is IEVMScriptExecutor {
         }
 
         bytes memory ret = new bytes(1);
-        ret[0] = 1;
-        
+        ret[0] = bytes1(1);
```
sohkai: @izqui Please sanity check this for me :).

The compiler was complaining:
> Warning: Decimal literal assigned to bytesXX variable will be left-aligned. Use an explicit conversion to silence this warning.

izqui: can we do `0x01` directly? anyway this looks good :)

PR #294, opened 2018-04-16
Issue275 vault recoverable

sohkai:
Refactors the `FundProxy` into `DepositableDelegateProxy` and `VaultRecoverable`.

Also moves the `ERCProxy` inheritance to `DelegateProxy`; was there a reason this was done in `FundsProxy` instead before?


contracts/apps/AragonApp.sol
```diff
@@ -30,4 +31,10 @@ contract AragonApp is AppStorage, Initializable, ACLSyntaxSugar, EVMScriptRunner
         }
         return address(kernel) == 0 || kernel.hasPermission(_sender, address(this), _role, how);
     }
+
+    function getRecoveryVault() public view returns (address) {
+        // Funds recovery via a vault is only available when used with a kernel
+        require(address(kernel) != 0);
```
sohkai: I've just disabled this functionality for apps that don't have a kernel, as the easiest way out.

If users use `AragonApp` by itself and want an escape hatch, they can implement their own.

test/proxy_funds.js
```diff
@@ -38,82 +38,125 @@ contract('Proxy funds', accounts => {
...
+    const r = await target.sendTransaction({ value: 1, gas: 31000 })
+    assert.equal((await getBalance(target.address)).valueOf(), initialBalance.plus(amount))
     return assertRevert(async () => {
-      await proxy.transferToVault(ETH)
+      await target.transferToVault(ETH)
     })
   }
 
-  context('App Proxy', async () => {
+  context('App without proxy is not recoverable', async () => {
```
bingen: Maybe the whole picture is still not clear enough for me, but do we really want this? At least is kind of confusing that `AragonApp is VaultRecoverable`.
Shouldn't we have a test for app without kernel?

bingen:
To be more explicit, I think it's way better now. A lot of great improvements: the FundsProxy split, the ERCProxy behind DelegateProxy, the IsContract one... really good job!!!!

test/proxy_funds.js
sohkai: Oops, should've redone this part as "App without kernel".

PR #295, opened 2018-04-16
Split APM creation transaction.

bingen:
To fix gas issues and on deployment and `npm run test`, `newAPM` is
now split into 2 transactions to avoid passing the transaction gas
limit.

On deployment of new APM we must be careful because if only the second
transaction fails the APM would be created but would have wrong
permissions.


sohkai:
Hmm, I really dislike how the permissions are in a weird state at the end of the first transaction.

It'd be really nice if we could figure this out instead by reducing the gas costs of the proxies somehow :).

bingen:
Replaced by #296.

PR #296, opened 2018-04-18
Split APM creation transaction

bingen:
To fix gas issues and on deployment and `npm run test`, `newAPM` is
now split into 2 transactions to avoid passing the transaction gas
limit. The first one just creates the DAO, which currently consumes
~1.5M gas.

Replaces #295


sohkai:
@bingen I think this is superseded by #297?

PR #298, opened 2018-04-19
[WIP] Switch namespace order - DO NOT merge!!

bingen:
Temporary PR with namespace change proposal, to remember later on an eventual discussion.

Before we were using, e.g. for kernel app:

`keccak256(keccak256(keccak256(keccak256(keccak256('core'), bytes32(0)),
'eth'), 'aragonpm'), 'kernel')`

With this change we would use:

`keccak256(keccak256(keccak256(keccak256(keccak256(bytes32(0)), 'eth'),
'aragonpm'), 'kernel'), 'core')`

Which is equivalent to `namehash("core.kernel.aragonpm.eth")`

TODO:
- Remove APMNamehash
- Change constants in IEVMScriptRegistry.sol
- Change constants in ACL.sol
- Change order everywhere namespaces are used
- Add bytes32(0) at the other end (after namespace)??

### DIFF
--- a/contracts/ens/ENSConstants.sol
+++ b/contracts/ens/ENSConstants.sol
@@ -3,8 +3,14 @@ pragma solidity ^0.4.18;
 
 contract ENSConstants {
     bytes32 constant public ENS_ROOT = bytes32(0);
+    /* Constants are hardcoded to save gas
     bytes32 constant public ETH_TLD_LABEL = keccak256("eth");
     bytes32 constant public ETH_TLD_NODE = keccak256(ENS_ROOT, ETH_TLD_LABEL);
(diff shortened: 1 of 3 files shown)


izqui:
@bingen can this be closed given the refactor to the new namehashes? https://github.com/aragon/aragonOS/pull/310

bingen:
I don't think so, that refactor is independent to the proposed change here. Maybe we can discuss it briefly next week and close if we decide so.

sohkai:
After a conversation with @bingen, we decided to close this for now.

The main reason was that using a proper namehash would make the internal namespacing of installed apps confusing—the namespace would come as the second argument in the hash (usually unintuitive) and the hash being a proper namehash could mean inadvertent collisions / confusion with existing ENS names.

PR #299, opened 2018-04-21
Allow newer versions of Solidity

jvluso:


sohkai:
Hey @jvluso, we haven't made it externally obvious why some of those contracts are pinned, but it's a general security practice to avoid compiler issues. We've consciously made `AragonApp` unpinned so users can choose their own solc versions for their own app.

See https://github.com/aragon/aragonOS/pull/236 for related info.

PR #301, opened 2018-05-04
renamed ACL functions to simpler names

NFhbar:
Based on this issue:

https://github.com/aragon/aragonOS/issues/223

I renamed the  ACL functions to reduce verbosity, ```npm test``` had all 224 tests passing. 
```npm run lint``` complained about using keyword ```emit``` for events but I did not find protocol for this since most files have different compiler versions, so I left alone.

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -106,7 +106,7 @@ contract ACL is IACL, AragonApp, ACLHelpers {
         public
     {
         bytes32 paramsHash = _params.length > 0 ? _saveParams(_params) : EMPTY_PARAM_HASH;
-        _set(_entity, _app, _role, paramsHash);
+        _setPermission(_entity, _app, _role, paramsHash);
     }
 
     /**
@@ -120,7 +120,7 @@ contract ACL is IACL, AragonApp, ACLHelpers {
         onlyPermissionManager(_app, _role)
         external
     {
-        _set(_entity, _app, _role, bytes32(0));
+        _setPermission(_entity, _app, _role, bytes32(0));
     }
 
     /**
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: contracts/acl/ACL.sol [hunk 1]
```diff
@@ -101,12 +101,12 @@ contract ACL is IACL, AragonApp, ACLHelpers {
...
     * @param _params Permission parameters
     */
-    function grantPermissionP(address _entity, address _app, bytes32 _role, uint256[] _params)
+    function grantP(address _entity, address _app, bytes32 _role, uint256[] _params)
         onlyPermissionManager(_app, _role)
         public
     {
         bytes32 paramsHash = _params.length > 0 ? _saveParams(_params) : EMPTY_PARAM_HASH;
-        _setPermission(_entity, _app, _role, paramsHash);
+        _set(_entity, _app, _role, paramsHash);
```
izqui: I wouldn't rename this, as `set` is too generic

izqui (changes requested):
Amazing, thanks so much for the contribution!

Just a tiny little thing

contracts/acl/ACL.sol
NFhbar: Changed back to ```_setPermission```.

sohkai (approved):
Let's wait a little bit before merging this, so we don't have to move the aragonOS 4.0 head too much.

sohkai:
Given the time since this was made, and that this would mean another breaking change (to an API that is used in quite a lot of places now), we won't be modifying these interfaces.

Thanks for doing this though @NFhbar, and sorry this didn't make it to `aragonOS@4`!

PR #310, opened 2018-05-14
Expose APMNamehash

sohkai:
Exposes `APMNamehash.sol` to let packages like the DAO kits to get an APM namehash without reimplementing it.

### DIFF
--- a/test/mocks/APMNamehash.sol
+++ b/test/mocks/APMNamehash.sol
@@ -1,10 +1,9 @@
 pragma solidity ^0.4.18;
 
+import "../../contracts/apm/APMConstants.sol";
 
-contract APMNamehash {
-    bytes32 constant public ETH_NODE = keccak256(bytes32(0), keccak256("eth"));
-    bytes32 constant public APM_NODE = keccak256(ETH_NODE, keccak256("aragonpm"));
 
+contract APMNamehash is APMConstants {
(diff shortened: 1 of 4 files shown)


sohkai:
@bingen See https://github.com/aragon/dao-kits/pull/1#discussion_r187854635

bingen:
I see, thanks. Then I would expose `APMNamehash` and keep `APM_NODE` inside it. As I see in that PR, `APM_NODE` is only used to compute `apmNamehash` just below (although not using the function) and I haven't seen any other use of `APM_NODE` anywhere else. `APMNamehash` contract is really small and to me it seems easier to read with the constant there than with an import and having to open another file.

test/mocks/APMNamehash.sol
```diff
@@ -1,10 +1,9 @@
 pragma solidity ^0.4.18;
 
+import "../../contracts/apm/APMConstants.sol";
 
-contract APMNamehash {
-    bytes32 constant public ETH_NODE = keccak256(bytes32(0), keccak256("eth"));
-    bytes32 constant public APM_NODE = keccak256(ETH_NODE, keccak256("aragonpm"));
```
bingen: I would keep this one inside `APMNamehash` and avoid the need for `APMConstants`.

PR #313, opened 2018-05-14
Fix incorrect role hashes

### LINKED ISSUE #307: ENSSubdomainRegistrar and APMRegistry roles should follow naming convention
They are still `bytes(1)` and so on. They should be converted to the new convention in order as our tooling assumes they do.

sohkai:
Fixes #307.


bingen:
Shouldn't we add tests using `test/mocks/KeccakConstants.sol`?

sohkai:
Good point! Missed that file yesterday and was wondering if we were testing these somewhere.

PR #317, opened 2018-05-16
Use npx everywhere

izqui:


package.json
```diff
@@ -3,16 +3,16 @@
   "version": "3.1.5",
   "description": "Core contracts for Aragon",
   "scripts": {
-    "compile": "truffle compile",
+    "compile": "npx truffle compile",
```
bpierre: I don’t think `npx` is needed in npm scripts? `node_modules/.bin` is already in the `$PATH`.

izqui: That's true, does it prioritize it meaning are binaries there checked before globally installed binaries?

bpierre: Yes it’s added at the beginning of the path, so it takes priority 👍

PR #323, opened 2018-05-23
Kernel: fix no returns from overloads of newAppInstance()

sohkai:
😰 I think I messed this up a while back.


bingen:
Kind of late, sorry, but some of comments:
- First of all, I'm afraid it was me who messed it up. :-o
- Could we then remove the name of the returned variable `appProxy`, as we are not using it?
- How come were tests using that wrapper working? See here:
https://github.com/aragon/aragonOS/blob/dev/test/mocks/KernelMock.sol#L22
https://github.com/aragon/aragonOS/blob/dev/test/proxy_recover_funds.js#L53
https://github.com/aragon/aragonOS/blob/dev/test/mocks/APMRegistryFactoryMock.sol#L38

sohkai:
Internally debated about removing the returned `appProxy` name, but ended up deciding to keep it to make the interfaces for both the exact same.

Those tests were working because JS tests don't get the return types, and the one solidity test uses the full parameterized function rather than the broken convenience overloads.

I didn't add any tests because of the JS tests not including return types, but we could include some solidity ones for the cases, like this, where the return values are really important.

PR #324, opened 2018-05-25
Remove delegate executors

sohkai:


test/evm_script.js
```diff
@@ -57,9 +52,7 @@ contract('EVM Script', accounts => {
```
bingen: Shouldn't the "3" in the description for this test be changed to "1"?

test/evm_script.js
```diff
@@ -171,187 +163,5 @@ contract('EVM Script', accounts => {
...
-
-            it('fails when executor selfdestructs', async () => {
-                return assertRevert(async () => {
-                    // passing some input makes executor to selfdestruct
-                    await executor.executeWithIO('0x00000004', '0x0001', [])
-                })
-            })
-        })
-
-        context('script overflow', async () => {
```
bingen: Why is this one being removed?

sohkai: Nice catch, this should stay :).

PR #344, opened 2018-06-20
Add SafeMath lib for uint8

bingen:


contracts/lib/zeppelin/math/SafeMath8.sol
```diff
@@ -0,0 +1,32 @@
...
+  function mul(uint8 a, uint8 b) internal pure returns (uint8) {
+    uint8 c = a * b;
+    require(a == 0 || c / a == b);
+    return c;
+  }
+
+  function div(uint8 a, uint8 b) internal pure returns (uint8) {
+    // assert(b > 0); // Solidity automatically throws when dividing by 0
+    uint8 c = a / b;
+    // assert(a == b * c + a % b); // There is no case in which this doesn't hold
```
izqui: We can use this opportunity to clean up these old comments

sohkai:
@bingen I just reverted the patch version upgrade and will merge this now. Another PR will be made to move these math libs to their latest gas optimized versions.

PR #346, opened 2018-07-12
EVMScriptRunner: remove protectState modifier

sohkai:
From https://github.com/ConsenSys/aragonOS-audit-repo/issues/28:

> While this works for the two attributes just mentioned it leaves all other storage variables exposed when the execution of `DeployDelegateScript` or `DelegateScript` is permitted.

> The protectState modifier gives a false sense of security and it should be removed from the contract.

Given that, in general, no `DelegateScript` executor would be considered safe if it contains an `SSTORE` opcode, this modifier is unnecessary.

### DIFF
--- a/contracts/evmscript/EVMScriptRunner.sol
+++ b/contracts/evmscript/EVMScriptRunner.sol
@@ -14,7 +14,7 @@ import "../apps/AppStorage.sol";
 contract EVMScriptRunner is AppStorage, EVMScriptRegistryConstants {
     using ScriptHelpers for bytes;
 
-    function runScript(bytes _script, bytes _input, address[] _blacklist) protectState internal returns (bytes output) {
+    function runScript(bytes _script, bytes _input, address[] _blacklist) internal returns (bytes output) {
         // TODO: Too much data flying around, maybe extracting spec id here is cheaper
         address executorAddr = getExecutor(_script);
         require(executorAddr != address(0));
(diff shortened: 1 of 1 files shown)


izqui:
@bingen I think `protectState` is a fine name (or `protectStateIntegrity`) and then properly comment what state can't be protected and why.

sohkai:
These are good points, although if

> someone with the right permissions can still add a malicious executor to their DAO

happens, I think we can assume they're pretty much screwed anyway. But, some protection is better than no protection, and perhaps `protectAppIntegrity` would best signal this intent.

thec00n:
I am just reading up on the conversation and would like to clarify why I have raised the issue as it is not a direct security issue. 

I generally think that building security controls that do not solve a specific security issue in its entirety is usually a sign that there is an underlying design problem and that the best solution to the problem has not been found yet. In the case of `protectState` and `kernelIntegrity` modifiers they only protect specific storage variables. This might be sufficient for now but should there be any changes in the future and more variables are added then the security controls are not sufficient and they give a false sense of security.

sohkai:
Closing this as it's something we'll defer until later.

In its current state, `protectState` can be misleading but does not present any inherent security risks.

In the future, we'll investigate how to make `protectState` more generic and user-driven, so that apps can define what they'd like to protect: https://github.com/aragon/aragonOS/issues/398.

Issue #325, opened 2018-06-04
Remove ACL bypass when Kernel is not set from AragonApp


izqui:
https://github.com/aragon/aragonOS/blob/dev/contracts/apps/AragonApp.sol#L32

This issue has been brought up by both audits as a major issue. I think given our knowledge of the system it doesn't seem like a big problem, but IMO there is no need to deploy production code that makes out unit testing of apps easier.

We should remove the `address(kernel) == address(0) || ...` shortcut from AragonApp and create an `AppProxyMock` that automatically sets the kernel to a kernel that will allow allow actions.

Pseudocode:
```solidity
contract KernelAllowAllMock is Kernel {
  function hasPermission(address _who, address _where, bytes32 _what, bytes _how) public view returns (bool) {
     return true;
  }
}

contract AppProxyAllowAllMock is AppProxyBase {
  address base;

  function AppProxyMock(address _base)
           AppProxyBase(new KernelAllowAllMock(), bytes32(0), new bytes(0)) {
    base = _base;
  }
  
  function implementation() public view returns (address) {
    return base;
  }
}
```

Then when testing apps, they would need to be tested through the mocked proxy

izqui:
https://github.com/ConsenSys/aragonOS-audit-repo/issues/22

PR #347, opened 2018-07-12
Kernel: remove kernelIntegrity modifier

sohkai:
`kernelIntegrity` has been made obsolete with the changes to check that addresses given to `setApp()` are always contracts (f7680657).

The only appropriate check `kernelIntegrity` could do is make sure that the base kernel is still a kernel contract, but this is impossible without something like a public `isKernel()` (which may not be a bad idea, as it also helps frontends).

### DIFF
--- a/contracts/kernel/Kernel.sol
+++ b/contracts/kernel/Kernel.sol
@@ -103,7 +103,7 @@ contract Kernel is IKernel, KernelStorage, Initializable, IsContract, AppProxyFa
     * @param _app Address of the app
     * @return ID of app
     */
-    function setApp(bytes32 _namespace, bytes32 _name, address _app) auth(APP_MANAGER_ROLE, arr(_namespace, _name)) kernelIntegrity public returns (bytes32 id) {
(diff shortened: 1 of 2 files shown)


bingen:
Oh, I see now that my second comment is done in PR #352 , so forget it.
https://github.com/aragon/aragonOS/pull/352/files#diff-807af7e85f274f9138c01132b86256b6R14

sohkai:
> Shouldn't we add the isContract check here too?

Good catch, I'll add the `isContract()` check there and in `AppProxyBase` as part of https://github.com/aragon/aragonOS/issues/325.

> This test will fail now. Not sure if the test makes sense, I'm afraid this question leads us again to the debate of whether aragonOS is intended to be used without proxies or not, haha.

This is an interesting find... I'd argue again that all the contracts should behave the same with and without a proxy in front of them. This particular test case highlights the odd behaviour of `kernelIntegrity`: if we're acting as the base contract, there should be no check this.

sohkai:
@bingen	I've parameterized the `kernel_apps` test to also test the Kernel contract by itself (as well as when behind a proxy): 3fa6e17.

It didn't make sense to do this for the other kernel tests, given one was for upgrading (proxy required) and the other is really just an ACL test we've neglected to rename as such.

PR #349, opened 2018-07-12
Allow tracking of number of authorized users for specific role

william2958:
Added functionality to check how many addresses are given permission for a specific role in the acl

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -17,6 +17,8 @@ contract ACL is IACL, AragonApp, ACLHelpers {
 
     // whether a certain entity has a permission
     mapping (bytes32 => bytes32) permissions; // 0 for no permission, or parameters id
+    // Number of people authorized with a permission
+    mapping (bytes32 => uint256) numberOfAuthorized;
     mapping (bytes32 => Param[]) public permissionParams;
 
     // who is the manager of a permission
@@ -222,6 +224,10 @@ contract ACL is IACL, AragonApp, ACLHelpers {
         return hasPermission(_who, _where, _what, empty);
     }
 
+    function numAuthorized(address _who, address _where, bytes32 _what) public view returns (uint) {
+        return numberOfAuthorized[permissionHash(_who, _where, _what)];
+    }
+
     function evalParams(
         bytes32 _paramsHash,
         address _who,
@@ -253,6 +259,11 @@ contract ACL is IACL, AragonApp, ACLHelpers {
     */
     function _setPermission(address _entity, address _app, bytes32 _role, bytes32 _paramsHash) internal {
         permissions[permissionHash(_entity, _app, _role)] = _paramsHash;
+        if (_paramsHash != bytes32(0)) {
+            numberOfAuthorized[permissionHash(_entity, _app, _role)] += 1;
+        } else {
+            numberOfAuthorized[permissionHash(_entity, _app, _role)] -= 1;
+        }
 
         SetPermission(_entity, _app, _role, _paramsHash != bytes32(0));
     }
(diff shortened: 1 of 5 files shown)


sohkai:
Thanks for the contribution @william2958!

Was there a particular use case you had in mind for accessing the number of permission holders on-chain? Otherwise, the information is quite easy to build off-chain with events (and therefore doesn't need to live in the contract itself).

william2958:
It is for in the situation where you use roles as a management tool for building an organization using Aragon, so it would be useful to be able to easily access how many people you have in a management role, or developer role for example. I think this would be a useful feature to integrate into the platform!

sohkai:
@william2958 Unfortunately I'm still not convinced this is entirely necessary on-chain. Would any contract or app behave differently based on the number of people that are holding a particular role?

florianlenz:
@sohkai Hi. The reason for having it on chain is, that it allows us to support voting based on a percentage of members instead of doing token based voting. It's something we definitely need and I would like to spend time on contributing to aragon rather than maintaining a fork.

bingen:
Hi @florianlenz Could you elaborate a little bit more how your Voting system would work? Are you implementing a new Voting app? Have you had a look at [aragon-apps/Voting](https://github.com/aragon/aragon-apps/tree/master/apps/voting)? Thanks.

florianlenz:
@bingen yes I did, tho not in depth (@william2958 did tho). As far as I can see it's token based voting.  

Imaging you have a proposal in an organisation. In your org, everyone who is a member (address with role xyz) can vote on a proposal. If a proposal get more than 50% of the votes it will be accepted (if not rejected). For that kind of voting you need to know how many addresses have role xyz.

bingen:
Hm, I see. How is the role management going to work in your organization? I mean, how are members going to be accepted into or removed from the organization? I see several solutions you could use:
- You could assimilate your role to a token, and only allowing 1 token per address, like with our Multisig (you can have a look at [that Kit](https://github.com/aragon/dao-kits/tree/master/kits/beta)). I guess the main problem would to remove that token from the address by the organization. For that you could extend a standard token with an special privilege, use `destroyToken` MiniMe function or use in combination with our [Staking app](https://github.com/aragon/aragon-apps/tree/staking_3/future-apps/staking) (soon coming) and force to stake that token in order to vote/prove memembership. There you have a `GOD_ROLE` for moving tokens.
- If your are developing some custom app for managing the membership, you could add the total count there.
- I guess doing it off-chain using `SetPermission` event doesn't work for you as you need it in another app, right?
If none of them work, I'll let @sohkai and @izqui propose another solution or accept this PR. We don't want to bloat `ACL` code with very special needs for a particular app, but at the same time we want of course the base system to be as generic and useful as possible. It's hard to find the right balance. But I think we should try to explore other possibilities first.

sohkai:
@bingen @florianlenz Our Token Manager app also has a ["burn" role](https://github.com/aragon/aragon-apps/blob/master/apps/token-manager/contracts/TokenManager.sol#L30) to forcibly remove tokens from a holder that could be used in the case of removing someone from the authorized list.

Again, hearing the use case, I'd hesitate to add this since the concept of "membership" is so well encapsulated with a token (would love to have this design documented better), and the current forwarding architecture is flexible enough to design these structures without additional apps and contracts. Hopefully, this saves you the work of implementing an entire voting app :).

florianlenz:
Thanks, I will check all options and see if they fit our use case.

sohkai:
@florianlenz @william2958 Just wanted to check in, did you take a look at using tokens for your use case?

florianlenz:
I checked it and we can close this PR since there are other ways to accomplish what we need. Thank you. @william2958 feel free to close the PR.

sohkai:
Thanks @florianlenz and @william2958!

PR #350, opened 2018-07-13
EVMScriptRunner: add ScriptResult log

sohkai:
From https://github.com/ConsenSys/aragon-apps-audit-repo/issues/32:

> The _executeVote function executes an EVMScript if a vote has sufficient support. The result of the execution is not logged nor is checked if it throws an exception.

### DIFF
--- a/contracts/evmscript/EVMScriptRunner.sol
+++ b/contracts/evmscript/EVMScriptRunner.sol
@@ -14,7 +14,7 @@ import "../apps/AppStorage.sol";
 contract EVMScriptRunner is AppStorage, EVMScriptRegistryConstants {
     using ScriptHelpers for bytes;
 
-    event ScriptResult(bytes script, bytes input, bytes returnData);
+    event ScriptResult(address indexed executor, bytes script, bytes input, bytes returnData);
 
     function runScript(bytes _script, bytes _input, address[] _blacklist) internal protectState returns (bytes output) {
         // TODO: Too much data flying around, maybe extracting spec id here is cheaper
@@ -28,7 +28,7 @@ contract EVMScriptRunner is AppStorage, EVMScriptRegistryConstants {
 
         output = returnedDataDecoded();
 
-        ScriptResult(_script, _input, output);
+        ScriptResult(executorAddr, _script, _input, output);
     }
 
     function getExecutor(bytes _script) public view returns (IEVMScriptExecutor) {
--- a/contracts/evmscript/EVMScriptRunner.sol
+++ b/contracts/evmscript/EVMScriptRunner.sol
@@ -14,7 +14,9 @@ import "../apps/AppStorage.sol";
 contract EVMScriptRunner is AppStorage, EVMScriptRegistryConstants {
     using ScriptHelpers for bytes;
 
-    function runScript(bytes _script, bytes _input, address[] _blacklist) protectState internal returns (bytes output) {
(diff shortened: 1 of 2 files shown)


test/evm_script.js
```diff
@@ -93,9 +93,16 @@ contract('EVM Script', accounts => {
...
 
-                await executor.execute(script)
+                const receipt = await executor.execute(script)
 
                 assert.equal(await executionTarget.counter(), 1, 'should have executed action')
+
+                // Check logs
+                // The Executor always uses 0x for the input and callscripts always have 0x returns
+                const scriptResult = receipt.logs.filter(l => l.event == 'ScriptResult')[0]
+                assert.equal(scriptResult.args.script, script)
```
bingen: I guess we could add some descriptions to these assertions, like for most of the rest of them, to help in case it fails.

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -24,7 +26,9 @@ contract EVMScriptRunner is AppStorage, EVMScriptRegistryConstants {
 
         require(executorAddr.delegatecall(sig, calldataArgs));
 
-        return returnedDataDecoded();
+        output = returnedDataDecoded();
+
+        ScriptResult(_script, _input, output);
```
izqui: Logging this can potentially be super expensive, we can log a hash of the script and that would be enough, as the full script has to be stored somewhere onchain anyway.

Also the `CallsScript` executor is logging all individual calls already.

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -14,7 +14,9 @@ import "../apps/AppStorage.sol";
 contract EVMScriptRunner is AppStorage, EVMScriptRegistryConstants {
     using ScriptHelpers for bytes;
 
-    function runScript(bytes _script, bytes _input, address[] _blacklist) protectState internal returns (bytes output) {
+    event ScriptResult(bytes script, bytes input, bytes returnData);
```
izqui: Logging the executor address (and indexing it) could be useful in this log.

contracts/evmscript/EVMScriptRunner.sol
sohkai: As far as I can tell with remix and the docs, logging more bytes doesn't actually cost more gas ATM. It's only indexed `bytes` and `string` parameters that get automatically keccaked (to provide the indexing).

From http://solidity.readthedocs.io/en/v0.4.24/contracts.html#events:

> If arrays (including string and bytes) are used as indexed arguments, the Keccak-256 hash of it is stored as topic instead.

----------------

Debating whether or not we should keccak / index these parameters anyway, for future-proofing.

contracts/evmscript/EVMScriptRunner.sol
sohkai: Added in 4c5000e.

contracts/evmscript/EVMScriptRunner.sol
izqui: Each byte of data logged costs 8 gas as defined in the yellowpaper:
2018-07-16 at 10 49 16 am

However I gave this a second thought and I think it is not a bad idea to log scripts as some forwarders never put scripts on storage.

sohkai: Just retested this and confirmed that it does cost gas.

PR #351, opened 2018-07-13
ACL: use more explicit nonces to calculate the role and params hash

sohkai:
Using `1` and `2` for the nonces seems kind of silly...

**This will be a backwards breaking change.**


izqui:
As you point out, upgrading an ACL app instance to this new version of the ACL will have catastrophic consequences (as all current roles and permissions will appear as never having been set). We need to make this very explicit, and discourage upgrading the ACL in current deployments

izqui:
As you point out, upgrading an ACL app instance to this new version of the ACL will have catastrophic consequences (as all current roles and permissions will appear as never having been set). We need to make this very explicit, and discourage upgrading the ACL in current deployments

PR #352, opened 2018-07-13
Cosmetic cleanup

sohkai:
Upgrades the linter, turns off unnecessary `pragma 0.5.0` rules, and fixes the linting.

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -3,11 +3,7 @@ pragma solidity 0.4.18;
 import "../apps/AragonApp.sol";
 import "./ACLSyntaxSugar.sol";
 import "./IACL.sol";
-
-
-interface ACLOracle {
-    function canPerform(address who, address where, bytes32 what, uint256[] how) public view returns (bool);
-}
+import "./IACLOracle.sol";
 
 
 /* solium-disable function-order */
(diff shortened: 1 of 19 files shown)


contracts/acl/ACL.sol
```diff
@@ -1,3 +1,6 @@
+/* solium-disable function-order */
```
izqui: Unsuccessfully tried to make this rule only apply to the `initialize(...)` function :( 

We can place this comment right before `contract ACL ...` instead of the beginning of the file, which looks weirder. Ideally we could create a custom solium rule that considers the initialize function as a constructor.

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -27,10 +31,6 @@ contract EVMScriptRunner is AppStorage, EVMScriptRegistryConstants {
         return returnedDataDecoded();
     }
 
-    function getExecutor(bytes _script) public view returns (IEVMScriptExecutor) {
-        return IEVMScriptExecutor(getExecutorRegistry().getScriptExecutor(_script));
-    }
-
     // TODO: Internal
```
izqui: We can remove this comment too

contracts/acl/ACL.sol
sohkai: Yes... we could write our own mini solium plugin for this, but ideally if you apply `solium-disable-next-line function-order`, it'd take the next line's function out of consideration for that rule.

sohkai: We'll see what happens: https://github.com/duaraghav8/Solium/issues/235

PR #353, opened 2018-07-13
Update SafeMath libraries to their current gas-optimized versions

sohkai:
Successor to #344.

The `SafeMath` libraries have undergone a few gas optimizations over the last year.


contracts/lib/zeppelin/math/SafeMath.sol
```diff
@@ -1,32 +1,55 @@
...
+        // uint256 c = a / b;
+        // assert(a == b * c + a % b); // There is no case in which this doesn't hold
+        return a / b;
+    }
+
+    /**
+    * @dev Subtracts two numbers, throws on overflow (i.e. if subtrahend is greater than minuend).
+    */
+    function sub(uint256 a, uint256 b) internal pure returns (uint256) {
+        assert(b <= a);
```
izqui: Should be a `require` as `b` being greater than `a` wouldn't be an internal inconsistency. By keeping this as an `assert` we would be making the assumption that consumers of the library always check that `b <= a` before performing the subtraction.

contracts/lib/zeppelin/math/SafeMath.sol
```diff
@@ -1,32 +1,55 @@
...
+        return a - b;
+    }
+
+    /**
+    * @dev Adds two numbers, throws on overflow.
+    */
+    function add(uint256 a, uint256 b) internal pure returns (uint256 c) {
+        c = a + b;
+        assert(c >= a);
```
izqui: Should be a `require` too.

sohkai:
> I think all `assert` should be made `require`

There's definitely some debate about this (now that it has made a difference), see as examples:

- [chriseth](https://medium.com/@chriseth/thanks-for-this-detailed-writeup-83addc4b7f87)
- Zeppelin rationale [1](https://github.com/OpenZeppelin/openzeppelin-solidity/issues/611#issuecomment-392594219) and [2](https://github.com/OpenZeppelin/openzeppelin-solidity/issues/565)

I imagine this is a great source of confusion in the community about how `SafeMath` is meant to be used. I certainly haven't been using it assuming those asserts should never be hit by adding more preconditions.

izqui:
Yeah, this is why I brought it up. Our codebase (and many more including OpenZeppelin's: [one of many examples](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/token/ERC20/StandardToken.sol#L93)) relies on SafeMath as a sanity check for integer inputs. 

If the SMT checker was passed, it would freak out anyway with the amount of potential assertions reached due to using SafeMath without checking input values first.

I would keep using reverts unless we add all the required input checks first.

sohkai:
After some offline discussion, we've opted to use `require()` instead of `assert()` in these libraries. For now we'll keep the names the same, since the old libraries that were available were also using `require()`.

sohkai:
Updated once more to the [latest version in open-zeppelin](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/d51e38758e1d985661534534d5c61e27bece5042/contracts/math/SafeMath.sol).

contracts/lib/zeppelin/math/SafeMath.sol
```diff
@@ -1,55 +1,69 @@
...
         // See: https://github.com/OpenZeppelin/openzeppelin-solidity/pull/522
-        if (a == 0) {
+        if (_a == 0) {
             return 0;
         }
 
-        c = a * b;
-        require(c / a == b);
+        uint256 c = _a * _b;
+        require(c / _a == _b);
```
bingen: Are we checking anything else than overflow here? If it's just overflow, we could do:
```
        uint256 c = _a * _b;
        if (_a > 1) { 
            if(_b > 1) {
                require(c > _a);
                require(c > _b);
            }
        }
        return c;
```
Which is another small gas gain.

contracts/lib/zeppelin/math/SafeMath.sol
```diff
@@ -1,32 +1,69 @@
...
+    function mul(uint256 _a, uint256 _b) internal pure returns (uint256) {
+        // Gas optimization: this is cheaper than requiring 'a' not being zero, but the
+        // benefit is lost if 'b' is also tested.
+        // See: https://github.com/OpenZeppelin/openzeppelin-solidity/pull/522
+        if (_a == 0) {
+            return 0;
+        }
+
+        uint256 c = _a * _b;
+        require(c / _a == _b);
```
bingen: Are we checking anything else than overflow here? If it's just overflow, we could do:
```
        uint256 c = _a * _b;
        if (_a > 1) { 
            if(_b > 1) {
                require(c > _a);
                require(c > _b);
            }
        }
        return c;
```
Which is another small gas gain.

sohkai: Not sure if this is actually a gas gain, but there might also be some numbers `a` and `b` where `(a * b) % uint256_max > max(a, b)`.

Probably best to suggest it to open-zeppelin if it turns out it is a gas gain :).

bingen: > Not sure if this is actually a gas gain, 

Yes, there is, I did some tests with Remix. Small though.

> but there might also be some numbers a and b where (a * b) % uint256_max > max(a, b).

I thought a little bit about it and I don't think so. I'll try to sketch a mathematical proof to be sure. Have you found a counter-example?

> Probably best to suggest it to open-zeppelin if it turns out it is a gas gain :).

You are absolutely right, I'll bring it up there. (Unless you found that counter-example, of course)

sohkai: Haven't, but a small script will probably find one or exhaust all cases pretty quickly, at least for `uint256_max`. Given arbitrary `a`, `b`, and constant `C` though, intuitively there should be counter examples (`uint256_max` might not as a special case).

An example would be `a = 3`, `b = 3`, `C = 5`; `(3 * 3) % 5` is `4`, which satisfies the above.

bingen: Oh, yes, I was so wrong! (sorry for distracting you). A similar counter-example would work for uint256: (2^255+1)*3. So forget it (and sorry again, my intuition was completely wrong and I should have checked it correctly before).

PR #354, opened 2018-07-14
Add test task to Travis CI

sohkai:
Updates `ganache-cli.sh` to `aragon-apps`' version.


sohkai:
Merging as CI is failing due to a failing test that has since been fixed in #352.

PR #356, opened 2018-07-15
[Prototype] Check for initialization during access control check

sohkai:
**The current tests haven't been adjusted (and new tests need to be added) so CI will be breaking**

Depends on #355.

Adds checks for initialization when checking if someone can perform an action (either `hasPermission()` or `canPerform()` in the `Kernel` and `AragonApp`, respectively). This solves 90% of the instances where people will need to add `isInitialized` in the face of the new requirement that all public-facing state-changing functions must check for initialization.


sohkai:
Will write tests with #355.

PR #357, opened 2018-07-15
Add more static contract types to state and function parameters

sohkai:
From https://github.com/ConsenSys/aragonOS-audit-repo/issues/14:

> Avoid using the address type for contract references. Instead use the actual contract type or an interface to statically restrict the different possible configurations that your contracts can be composed into.

-------------

Some parameters or state are left to be `address` due to their dynamic nature.

### DIFF
--- a/contracts/evmscript/EVMScriptRunner.sol
+++ b/contracts/evmscript/EVMScriptRunner.sol
@@ -16,13 +16,13 @@ contract EVMScriptRunner is AppStorage, EVMScriptRegistryConstants {
 
     function runScript(bytes _script, bytes _input, address[] _blacklist) protectState internal returns (bytes output) {
         // TODO: Too much data flying around, maybe extracting spec id here is cheaper
-        address executorAddr = getExecutor(_script);
-        require(executorAddr != address(0));
+        IEVMScriptExecutor executorAddr = getExecutor(_script);
+        require(address(executorAddr) != address(0));
 
         bytes memory calldataArgs = _script.encode(_input, _blacklist);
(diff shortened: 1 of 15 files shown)


contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -16,13 +16,13 @@ contract EVMScriptRunner is AppStorage, EVMScriptRegistryConstants {
 
     function runScript(bytes _script, bytes _input, address[] _blacklist) protectState internal returns (bytes output) {
         // TODO: Too much data flying around, maybe extracting spec id here is cheaper
-        address executorAddr = getExecutor(_script);
-        require(executorAddr != address(0));
+        IEVMScriptExecutor executorAddr = getExecutor(_script);
```
bingen: I guess we could remove the `Addr` suffix here.

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -16,13 +16,13 @@ contract EVMScriptRunner is AppStorage, EVMScriptRegistryConstants {
...
+        require(address(executorAddr) != address(0));
```
bingen: Do we need to explicitly cast to `address` here?

sohkai: For now, no, but it actually looks like in [`pragma 0.5.0`, the implicit `address` to `Contract` inheritance will change (and so we will need explicit casting)](https://solidity.readthedocs.io/en/v0.4.24/types.html#address).

This probably also means the type checker won't let us get away with the implicit conversion in function arguments, but let's save that for when we do move up.

PR #362, opened 2018-07-16
ACL: log whether a permission was set with parameters

### LINKED ISSUE #212: ACL: SetPermission event should log whether permission is granted with params or not

sohkai:
Fixes #212.


sohkai:
So just wanted to highlight that this change will be a headache for the frontend migration.

The new event signature means that aragon.js won't know about old DAOs' ACL's `setPermission` event.

sohkai:
I've since changed this to instead emit another event, so that it's backwards compatible.

PR #363, opened 2018-07-16
KernelConstants: remove unnecessary constants

sohkai:
Fixes the CI.

These constants aren't being used anymore as they were only used previously to build other constants.


sohkai:
> `APM_NODE` is defined in `APMNamehash` but still like `keccak256(ETH_TLD_NODE, keccak256("aragonpm"))`

As far as I can tell, this isn't actually being used by any of the contracts, and there's also some keccak constants left in `ENSConstants.sol`. The actual gas savings from converting these into bytes is also usually quite small (~few thousand and only during deployment) so I'd say we can make them hardcoded after splitting APM from aOS.

Issue #365, opened 2018-07-23
Truffle Migration Failure


mnaamani:
When running truffle migration I see the following error after the last migration step:
```
(node:63563) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): SyntaxError: Unexpected end of JSON input
(node:63563) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
node[63563]: ../src/timer_wrap.cc:107:static void node::(anonymous namespace)::TimerWrap::Stop(const FunctionCallbackInfo<v8::Value> &): Assertion `HandleWrap::IsAlive(wrap)' failed.
 1: node::Abort() [/usr/local/bin/node]
 2: node::(anonymous namespace)::DomainEnter(node::Environment*, v8::Local<v8::Object>) [/usr/local/bin/node]
 3: node::(anonymous namespace)::TimerWrap::~TimerWrap() [/usr/local/bin/node]
 4: v8::internal::FunctionCallbackArguments::Call(void (*)(v8::FunctionCallbackInfo<v8::Value> const&)) [/usr/local/bin/node]
 5: v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) [/usr/local/bin/node]
 6: v8::internal::Builtin_Impl_HandleApiCall(v8::internal::BuiltinArguments, v8::internal::Isolate*) [/usr/local/bin/node]
 7: 0x1f532b98463d
```

I believe this is a result of truffle not running the migrations sequentially, or issues with how async/await pattern is used in the migrations scripts.

Error seen when running following versions of node and truffle:

Node v8.9.4
Truffle v4.1.13

A good pattern that has worked well for me to use in the migrations scripts:

```
module.exports = function (deployer, network, accounts) {
  return deployer.then(async () => {

  })
}
```

sohkai:
Closed by #366.

PR #366, opened 2018-07-23
Fix truffle migration async scripts

mnaamani:
Solves #365 using the pattern seen here: https://github.com/trufflesuite/truffle/issues/501


sohkai:
Thanks @mnaamani, it looks like this was working for us out of luck, or something with how node timeouts.

The new migration logs look much better (was always puzzled before why they seemed to be out of order).

PR #371, opened 2018-07-24
Coverage: fix proxy_recover_funds tests failing due to vault being behind a proxy

sohkai:
As @izqui pointed out, the reason the ETH recover fails during coverage is because the `VaultMock` instance is behind a proxy (which is instrumented and therefore uses up all the available gas before completing the transfer).

One can make an argument that the vault being behind a proxy is more similar to real world scenarios, but I'd argue that an integration test would better serve that purpose.


sohkai:
@izqui 

> Is there any reason we are keeping aragonOS/test/helpers/skipCoverage.js in the repo? We should remove it if it isn't going to be used elsewhere.

Thought I'd keep it there to make the eventual transition of those test utils into a library easier, but yes, we can remove them and look up git history later :).

izqui:
I hear @bingen's complain and I think you are right that we shouldn't stop testing this completely. Maybe we can add some extra tests that don't run while checking coverage, that just tests the happy path (recovering eth and tokens) for depositing in a proxied Vault.

sohkai:
@bingen @izqui Alright guys, I've added some parametrization to the test so that it's tested both with a proxy-less and proxied vault.

The main reason I don't consider a realistic vault set up being too important here is that we don't use any aspects of the vault outside of it being a contract. It could be any mock contract and it'd work well in this test.

PR #373, opened 2018-07-27
Clean package

sohkai:
Small cleanup, removing the lock files and unnecessary dependencies.


yxliang01:
To be honest, lock files for packages should be committed to ensure the consistency of the dependencies across machines. Is this really wanted? @sohkai

sohkai:
It was because we had issues with installing off of a package-lock.json.

Those issues have been fixed in the last year though, and we will be re-introducing lock files.

yxliang01:
:+1: Created PR https://github.com/aragon/aragonOS/pull/560 @sohkai

PR #374, opened 2018-07-27
Truffle config keys management enhance

bingen:
- Allow for multiple private keys on Rinkeby
- Allow to define custom mnemonic
- Move key files to `~/.aragon/` folder

Waiting for this: https://github.com/rhlsthrm/truffle-hdwallet-provider-privkey/issues/9


truffle-config.js
```diff
@@ -10,8 +15,8 @@ if (process.env.LIVE_NETWORKS) {
   kovanProvider = new HDWalletProvider(mnemonic, 'https://kovan.infura.io')
 
   try {
-    const { rpc, key } = require(require('homedir')()+'/.rinkebykey.json')
-    rinkebyProvider = new PrivateKeyProvider(key, rpc)
+    const { rpc, keys } = require(require('homedir')()+'/.aragon/rinkebykey.json')
+    rinkebyProvider = new HDWalletProviderPrivkey(keys, rpc)
```
sohkai: Not familiar with how truffle handles HD wallets... what does truffle do with the HD wallet that's different from a private key?

bingen: I'm not familiar either, and not sure if I understood correctly the question, but we were using `HDWalletProvider` for mnemonics and `PrivateKerProvider` for private keys. The problem with this last one is that it only allows for 1 private key. With this new package we can have several keys.

sohkai: But what would truffle do with several keys? Does it fire off transactions in a round-robin fashion using them or just the root key (which would be the same as a single private key)?

bingen: Then you have, as in our tests, an array `accounts` with several non-empty values. It uses index `0` by default, but you can use `, { from: accounts[i] }` as you want to sign transactions with other accounts.

sohkai: Ahh, OK, that makes sense :+1:

sohkai:
@bingen Can you remove the `package-lock.json` file? I've since removed this from this repo as well.

PR #375, opened 2018-07-30
Uint and Time helpers

### LINKED ISSUE #361: Provide uint64 time helpers
We normally assume `uint64` is enough for time (it should be...), but to be explicitly safe, we can also provide a safe conversion utility for `uint256 -> uint64`.

We should also add base contracts to supply both time and block number so the interface is consistent (and maybe even provide a mocked version for easier testing) in apps.

bingen:
Closes #361 
TODO:
- [x]  Add tests for new functions


contracts/common/Initializable.sol
```diff
@@ -40,4 +40,31 @@ contract Initializable is AppStorage {
     function getBlockNumber() internal view returns (uint256) {
         return block.number;
     }
+
+    /**
+    * @dev Returns the current block number, converted to uint64.
+    *      Using a function rather than `block.number` allows us to easily mock the block number in
+    *      tests.
+    */
+    function getBlockNumber64() internal view returns (uint64) {
```
sohkai: I think it'd be nice if we moved these time helpers into their own contract file, e.g. `TimeHelpers`.

contracts/lib/misc/Uint64Helpers.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity ^0.4.18;
+
+
+library Uint64Helpers {
```
sohkai: I'd call this `Uint256Helpers`, since it's working on `uint256`s.

test/lib_uint64_helpers.js
```diff
@@ -0,0 +1,21 @@
...
+
+  it('converts from uint256 to uint64', async () => {
+    const a = 1234
+    assert.equal((await uint64Mock.convert.call(a)).toString(), a, "Values should match")
+  })
+
+  it('fails converting from uint256 to uint64 if too big', async () => {
+    const a = new web3.BigNumber(2).pow(64)
+    return assertRevert(async () => {
+      assert.equal((await uint64Mock.convert(a)).toString(), a, "Values should match")
```
sohkai: Small nitpick: this can just be `await uint64Mock.convert(a)` rather than an `assert()` test.

bingen: Haha, yes, problems of copy-pasting...

sohkai:
Lacking coverage will be fixed with https://github.com/aragon/aragonOS/pull/372.

contracts/lib/misc/TimeHelpers.sol
```diff
@@ -0,0 +1,44 @@
+/*
+ * SPDX-License-Identitifer:    MIT
+ */
+
+pragma solidity ^0.4.18;
+
+
+contract TimeHelpers {
```
sohkai: Last thing 😅, I'd move this to `contracts/common` since it's not imported / copied from somewhere else (it's similar to `IsContract`, which also lives there).

sohkai: Last thing 😅, I'd move this to `contracts/common` since it's not imported / copied from somewhere else (it's similar to `IsContract`, which also lives there).

contracts/common/TimeHelpers.sol
```diff
@@ -0,0 +1,44 @@
...
+        return block.number;
+    }
+
+    /**
+    * @dev Returns the current block number, converted to uint64.
+    *      Using a function rather than `block.number` allows us to easily mock the block number in
+    *      tests.
+    */
+    function getBlockNumber64() internal view returns (uint64) {
+        return uint64(block.number);
```
sohkai: On these conversions, can we use the new `Uint256Helpers`?

bingen: Yes, we can, and I thought about it, but I considered unnecessary to import another library as I thought if block number or timestamp were ever bigger than `2^64` (`~1.8e19`, for time almost `6e11` years) this would be a minor problem. ;-)

sohkai: Despite being improbable, I think it's best to use the safe conversion. We use SafeMath when incrementing a lot of these time-based variables in the apps so we should apply the same rigour here.

bingen: Ok, I'll change it, although I kind of disagree. I don't remember by heart all that occurrences, but I think they mostly involve "external" data (parameters, computed, etc), while here we are just converting "system" values (`block.number` and `now`) so I think it's pointless.

contracts/lib/misc/Uint256Helpers.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity ^0.4.18;
```
izqui: Shouldn't this file be in the `common` directory as well?

test/time_helpers.js
```diff
@@ -0,0 +1,15 @@
+contract('TimeHelpers test', accounts => {
+  let timeHelpersMock
+
+  before(async () => {
+    timeHelpersMock = await artifacts.require('TimeHelpersMock').new()
+  })
+
+  it('checks block number', async () => {
+    assert.equal((await timeHelpersMock.getBlockNumberExt.call()).toString(), (await timeHelpersMock.getBlockNumber64Ext.call()).toString(), "block number should match")
```
izqui: We should also test that the returned value is the actual timestamp or block number.

PR #376, opened 2018-07-30
Move AragonApp to unstructured storage

### LINKED ISSUE #342: Migrate to unstructured storage for some App and Kernel storage variables
See https://blog.zeppelinos.org/upgradeability-using-unstructured-storage/

Good candidates:

- `initializationBlock` (in `AppStorage`)
- `pinnedCode` (in `AppStorage`)
- `recoveryVaultId` (in `KernelStorage`)

Potential candidates:

- `kernel` (in `AppStorage`)
- `appId` (in `AppStorage`)

This would also let us remove the `offset` in `AppStorage`.

bingen:
Fixes #342 

TODO:

- [x] ~~Do the same for `Kernel`~~
- [x] Adapt dao kits and apps if needed


contracts/apps/AppStorage.sol
```diff
@@ -4,14 +4,14 @@
...
-    IKernel public kernel;
-    bytes32 public appId;
-    address internal pinnedCode; // used by Proxy Pinned
-    uint256 internal initializationBlock; // used by Initializable
-    uint256[95] private storageOffset; // forces App storage to start at after 100 slots
-    uint256 private offset;
+contract AppStorage is UnstructuredStorage {
+    // keccak256("IKernel.kernel")
+    bytes32 public constant kernelPosition = 0xe30296d9191ef86f01c8532636453e486db506a7be081a32b88c263fbc3a5e15;
+    // keccak256("bytes32.appIdPosition")
```
izqui: Should be `bytes32.appId` for consistency with other names.

contracts/apps/AppStorage.sol
```diff
@@ -4,14 +4,14 @@
...
 
-contract AppStorage {
-    IKernel public kernel;
-    bytes32 public appId;
-    address internal pinnedCode; // used by Proxy Pinned
-    uint256 internal initializationBlock; // used by Initializable
-    uint256[95] private storageOffset; // forces App storage to start at after 100 slots
-    uint256 private offset;
+contract AppStorage is UnstructuredStorage {
+    // keccak256("IKernel.kernel")
```
izqui: I would remove the type from the key names as types can change in the future and they are all storing a 32-byte word anyway.

It would make sense to have the type be explicit if we were storing more complex data types as arrays, mappings or structs, but for simple values I would remove it.

izqui: I would also add an `aragonOS.appStorage` prefix to all keys to remove the possibility of collisions.

contracts/common/UnstructuredStorage.sol
```diff
@@ -0,0 +1,35 @@
+/*
+ * SPDX-License-Identitifer:    MIT
+ */
+
+pragma solidity ^0.4.18;
+
+
+contract UnstructuredStorage {
+    // keccak256("uint256.initializationBlock"), used by Initializable
+    bytes32 public constant initializationBlockPosition = 0xd9120073d934f0c287a3a735b193740486dbba88578fd8f3cbb2b9a9b654e7ce;
```
izqui: Why not move to the `Initializable` contract directly? Same comment on the naming of the key as for the other storage values in `AppStorage`.

contracts/common/Initializable.sol
```diff
@@ -7,29 +7,29 @@ pragma solidity ^0.4.18;
...
+contract Initializable is UnstructuredStorage {
     modifier onlyInit {
-        require(initializationBlock == 0);
+        require(getStorageUint256(initializationBlockPosition) == 0);
         _;
     }
 
     modifier isInitialized {
-        require(initializationBlock > 0);
+        require(getStorageUint256(initializationBlockPosition) > 0);
```
izqui: Should use `getInitializationBlock()`

contracts/apps/AppStorage.sol
```diff
@@ -4,14 +4,14 @@
...
+    bytes32 public constant kernelPosition = 0xe30296d9191ef86f01c8532636453e486db506a7be081a32b88c263fbc3a5e15;
+    // keccak256("bytes32.appIdPosition")
+    bytes32 public constant appIdPosition = 0x71ca50ba614d1d1a9a2433a5d0187a13e1a8c20bbe0013968dd089c71fd760eb;
+    // keccak256("address.pinnedCode"), used by Proxy Pinned
+    bytes32 public constant pinnedCodePosition = 0x0e2081dcdbb92be1037e55b39e913e70697c33a8a8b9bb01110e6f9e576089b8;
```
izqui: I think the storage positions are an implementation detail that doesn't need to be exposed in the public API of the contract.

For all values I would add a getter and setter so the fact that AppStorage is using UnstructuredStorage is abstracted away from people interacting with the contract or deriving it:

```
function kernel() public view returns (IKernel) {
  return IKernel(getStorageAddress(kernelPosition));
}

function setKernel(IKernel newKernel) internal {
  setStorageAddress(kernelPosition, address(_kernel));
}
```

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -9,7 +9,7 @@ import "./IEVMScriptExecutor.sol";
 import "./IEVMScriptRegistry.sol";
 
 import "../apps/AppStorage.sol";
-
+import "../kernel/IKernel.sol";
```
izqui: No need to import

test/unstructured_storage.js
```diff
@@ -0,0 +1,61 @@
+const getContract = name => artifacts.require(name)
+
+contract('Constants', accounts => {
```
izqui: 'Constants' -> 'Unstructured storage'

test/unstructured_storage.js
```diff
@@ -0,0 +1,61 @@
...
+  let app, kernel
+
+  beforeEach(async () => {
+    app = await getContract('AppStubStorage').new()
+    kernel = await getContract('Kernel').new()
+  })
+
+  it('test', async () => {
```
izqui: Would split into one test case per value being checked

test/unstructured_storage.js
```diff
@@ -0,0 +1,61 @@
...
+    // set values
+    await app.initialize()
+    await app.setKernelExt(kernel.address)
+    await app.setAppIdExt(appId)
+    await app.setPinnedCodeExt(pinnedCode)
+
+    //checks
+    assert.equal(
+      parseInt(await web3.eth.getStorageAt(app.address, (await app.getInitializationBlockPosition())), 16),
+      (await app.getInitializationBlock.call()).toString(),
```
izqui: You can check what the block number was after `app.initialize()` and assert equality with the value stored at the storage position.

test/unstructured_storage.js
```diff
@@ -0,0 +1,61 @@
...
+
+    // set values
+    await app.initialize()
+    await app.setKernelExt(kernel.address)
+    await app.setAppIdExt(appId)
+    await app.setPinnedCodeExt(pinnedCode)
+
+    //checks
+    assert.equal(
+      parseInt(await web3.eth.getStorageAt(app.address, (await app.getInitializationBlockPosition())), 16),
```
izqui: I would get the positions from `test/mocks/KeccakConstants.sol` rather than from 'AragonApp', even though that equality already tested in `test/keccak_constants.js`. Would allow to remove all the position public getters from `AppStubStorage`.

bingen: Well, but if we remove the public getters from `AppStubStorage` then tests in `keccak_constants.js` won't work. Here I see we are testing 2 things:
- That hardcoded position is what it is intended to be (not a big deal, but better to have it, to avoid collisions)
- That things are properly stored at those positions.
Testing the second one first one would be implicitly tested if we use positions from `KeccaKConstants` but I think it's better to have both for 2 reasons: 1) in case of failing, it's eaiser to know what the problem is, 2) we could be pointing at another position that has the same value (yes, unlikely, but still) and having both tests we would have more chances to catch it.

sohkai:
This change should be transparent to the DAO Kits and apps.

migrations/2_apm.js
```diff
@@ -36,7 +36,7 @@ module.exports = function (deployer, network, accounts, arts = null) {
     console.log('Deployed APM at:', apmAddr)
 
     const apm = getContract('APMRegistry').at(apmAddr)
-    console.log('Kernel:', await apm.kernel())
+    console.log('Kernel:', await apm.kernel.call())
```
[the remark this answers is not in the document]
bingen: Yes, they are for `testrpc` but I had some problems without using `geth` with our dao-kits, so now I'm putting them whenever I remember, so in case we finally move our tests to geth they will work.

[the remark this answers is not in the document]
sohkai: Hmm, weird, this should be handled by truffle (truffle-contract changes its interface based on whether the function is marked `constant` or not), rather than a testrpc / geth issue.

sohkai:
For deployment cost:

```
AppProxyPinned.json           102000 more gas      +510               +346
AppProxyUpgradeable.json      88600 more gas       +443               +240
EVMScriptRegistryFactory.json 307800 more gas      +1539              +363
```

Ouch :(

sohkai:
Current deployment costs:

```
                         CODE DEPOSIT COST    DEPLOYED BYTES     INITIALIZATION BYTES              │~
AppProxyPinned.json      57600 more gas       +288               +375                              │~
AppProxyUpgradeable.json 40800 more gas       +204               +258                              │~
Initializable.json       30000 less gas       -150               -1
```

PR #378, opened 2018-07-31
[Prototype] Default bases to petrified

sohkai:
Experiments with #355, adding a few ideas:

1. `AragonApp` (c5806bf): default behaviour is to petrify immediately on deployment
1. `UnsafeAragonApp` (6ac3cb4): freestyle mode, for app developers. Removes all safety defaults (only `initializationBlock` for now) and lets the developer play with fire. Useful for mocking in tests too.
1. `Autopetrified` (ee28768): provide helper to petrify immediately on deployment
1. `EVMScriptExecutor` (17b12c1): petrify the executors, so that they can't be used outside of a `delegatecall` originating from an initialized contract (ideally an `EVMScriptRunner`).

Comments and tests to be adjusted.

### DIFF
--- a/contracts/evmscript/executors/CallsScript.sol
+++ b/contracts/evmscript/executors/CallsScript.sol
@@ -3,10 +3,10 @@ pragma solidity 0.4.18;
 // Inspired by https://github.com/reverendus/tx-manager
 
 import "../ScriptHelpers.sol";
-import "../IEVMScriptExecutor.sol";
+import "./BaseEVMScriptExecutor.sol";
 
 
-contract CallsScript is IEVMScriptExecutor {
+contract CallsScript is BaseEVMScriptExecutor {
     using ScriptHelpers for bytes;
 
     uint256 constant internal SCRIPT_START_LOCATION = 4;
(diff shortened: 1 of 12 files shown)


contracts/evmscript/executors/CallsScript.sol
```diff
@@ -3,10 +3,10 @@ pragma solidity 0.4.18;
 // Inspired by https://github.com/reverendus/tx-manager
 
 import "../ScriptHelpers.sol";
-import "../IEVMScriptExecutor.sol";
+import "./BaseEVMScriptExecutor.sol";
 
 
-contract CallsScript is IEVMScriptExecutor {
+contract CallsScript is BaseEVMScriptExecutor {
     using ScriptHelpers for bytes;
```
bingen: Why are we removing the interface here? I think we should still enforce that `CallScript` complies with `IEVMScriptExecutor` interface.

izqui: Right, `BaseEVMScriptExecutor` should be `IEVMScriptExecutor`

PR #380, opened 2018-08-15
Make UnstructuredStorage a library

sohkai:
Also moves the `pinnedCode` implementation details to `AppProxyPinned` as it's only relevant there, and unstructured storage is meant to hide these types of abstractions :).

### DIFF
--- a/contracts/apps/AppProxyPinned.sol
+++ b/contracts/apps/AppProxyPinned.sol
@@ -1,9 +1,15 @@
 pragma solidity 0.4.18;
 
+import "../common/UnstructuredStorage.sol";
 import "./AppProxyBase.sol";
 
 
 contract AppProxyPinned is AppProxyBase {
+    using UnstructuredStorage for bytes32;
+
+    // keccak256("aragonOS.appStorage.pinnedCode"), used by Proxy Pinned
+    bytes32 internal constant PINNED_CODE_POSITION = 0xdee64df20d65e53d7f51cb6ab6d921a0a6a638a91e942e1d8d02df28e31c038e;
+
     /**
     * @dev Initialize AppProxyPinned (makes it an un-upgradeable Aragon app)
     * @param _kernel Reference to organization kernel for the app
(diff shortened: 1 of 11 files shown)


sohkai:
So actually this is buried in the solidity docs, but `internal` functions in a library are copied just the same as base class functions. In this case, `UnstructuredStorage` being a library (with only internal functions) has two stylistic advantages:

- No need to add it to the inheritance graph (:heart:)
- Allows you to access them as if they were part of the type, e.g. we can mark `bytes32` as a storage position.

bingen:
What about the bytecode size? Did you check it? Is it happening the same as in #382 ? In that case, would we wait to 0.4.24 too? (https://github.com/aragon/aragonOS/pull/382#issuecomment-413700424)

sohkai:
Admittedly this was before #382 so I hadn't thought of checking the bytecode size... can do some tests soon.

bingen:
I'm fine with the changes, but I guess not that excited, haha.
I agree about how better is to put `pinnedCode` stuff in `AppProxyPinned`, but we can do that without converting to a library.
About simplifying inheritance graph, I see it more like "hiding" complexity than actually simplifying anything. When I look at that graph, I like to realize that `AppStorage` is `UnstructuredStorage`. Besides, we are only removing 2/3 dependencies in a huge graph. 
Anyway this is highly subjective, so, again, I'm fine with the changes. But I definitely wouldn't do it if it comes with an objective loss like increasing bytecode size.

sohkai:
For the library change (f322c87):

```
AppProxyPinned.json           44400 less gas       -222               +29
AppProxyUpgradeable.json      47800 less gas       -239               +18
EVMScriptRegistryFactory.json 82800 less gas       -414               -237
Initializable.json            57400 less gas       -287               -1
UnstructuredStorage.json      69000 less gas       -345               -3
```

Awkwardly, this seems to be a gas cost decrease for some reason that #382 was not.

sohkai:
Given that the bytecode increment seems to be halved by making `UnstructuredStorage` a library, I'd say there's objective improvement with this change :).

I also don't see this as "hiding" any complexity, since we're still declaring that `bytes32` can be unstructured storage positions (`using UnstructoredStorage for bytes32`). The reason I prefer libraries for these kinds of contracts is because inheriting in solidity can be misleading due to the C3 linearization, whereas libraries are plain and obvious. In that aspect, I think complexity is _greatly_ reduced when the inheritance list is smaller.

bingen:
Wow! `AragonApp` from `61800` to `11800`!! This is definitely a huge objective improvement! End of discussion ;-)
(Although I still kind of disagree about the complexity part)

PR #382, opened 2018-08-16
IsContract: convert to library instead

sohkai:
Mainly simplifies the inheritance graphs, and also might be more aesthetically pleasing.

Does make the contract _slightly_ larger (Kernel.sol is ~150 bytes more and AppProxyUpgradeable is ~40 bytes more), meaning more gas to deploy them.

### DIFF
--- a/contracts/apps/AppProxyBase.sol
+++ b/contracts/apps/AppProxyBase.sol
@@ -2,10 +2,13 @@ pragma solidity 0.4.18;
 
 import "./AppStorage.sol";
 import "../common/DepositableDelegateProxy.sol";
+import "../common/IsContract.sol";
 import "../kernel/KernelStorage.sol";
 
 
 contract AppProxyBase is AppStorage, DepositableDelegateProxy, KernelConstants {
+    using IsContract for address;
+
     /**
     * @dev Initialize AppProxy
     * @param _kernel Reference to organization kernel for the app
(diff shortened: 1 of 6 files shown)


sohkai:
Agreed, although the gas changes don't make any sense since it should be the same to the compiler.

I'll wait til after we upgrade to 0.4.24 and see if there's changes in the size of the bytecode, and if it's still this big, close this.

sohkai:
Was curious about the impact of optimizing:

Difference at this branch's HEAD with and without 200 optimizer runs:

```
APMRegistryFactory.json        863800 less gas      -4319              -330
AppProxyPinned.json            134200 less gas      -671               -326
AppProxyUpgradeable.json       128800 less gas      -644               -210
AppStorage.json                9400 less gas        -47                -1
Controlled.json                20800 less gas       -104               -35
DAOFactory.json                571600 less gas      -2858              -174
ENS.json                       191400 less gas      -957               -23
EVMScriptRegistryConstants.js… 9800 less gas        -49                -1
EVMScriptRegistryFactory.json  739800 less gas      -3699              -1955
KernelProxy.json               118800 less gas      -594               -27
Migrations.json                48800 less gas       -244               -35
MiniMeToken.json               920800 less gas      -4604              -161
PublicResolver.json            428800 less gas      -2144              -39
```

Difference between this branch's target and this branch's HEAD, both with 200 optimizer runs:

```
AppProxyPinned.json           1800 more gas        +9                 +9
AppProxyUpgradeable.json      11200 less gas       -56                +9
EVMScriptRegistryFactory.json 5800 less gas        -29                +9
```

So it's obvious optimizing actually does something, but the second result is honestly just baffling. Somehow the optimizer doesn't like the pinned version.

sohkai:
Closing this, since the cosmetic gain doesn't justify the gas costs right now.

However, if we do enable optimization, it looks like this should be a net gain, as we use `AppProxyUpgradeable` a lot more than the pinned version.

PR #387, opened 2018-08-20
Add bytecode manager and current bytecode to diff against

sohkai:
Adds a [small utility](https://github.com/sohkai/truffle-bytecode-manager) to extract and diff bytecode against.

~~I've opted to include the bytecode in the repo so we can track it, and update it on every merge by using `npm run bytecode:extract` (maybe we can get travis to do this eventually). That way, any change we do can be diffed against the current dev HEAD.~~


sohkai:
As @izqui pointed out offline, including the bytecode like this imposes requirements for humans to remember to update the bytecode on each PR.

We'll investigate how to do this in a more automated fashion, e.g. a script run on travis.

PR #391, opened 2018-08-21
EVMScriptExecutor: add type for off-chain reading

sohkai:
Was getting worried that we always assume in `aragon.js` that the first executor in an organization's `EVMScriptRegistry` is the callscript executor, rather than doing an on-chain confirmation.

**TODO**:

- [x] Add tests


sohkai:
Bytecode size:

```
CallsScript.json              20600 more gas       +103               0
```

PR #392, opened 2018-08-22
EVMScriptRegistry: add ability to re-enable disabled executors

sohkai:
Following #390, we realized there was no way to re-enable executors after they've been disabled.

Bytecode changes:

```
DAOFactory.json               10400 less gas       -52                0
EVMScriptRegistry.json        100800 more gas      +504               0
EVMScriptRegistryFactory.json 17000 less gas       -85                +504
KeccakConstants.json          23800 more gas       +119               0
```


contracts/evmscript/EVMScriptRegistry.sol
```diff
@@ -44,6 +44,12 @@ contract EVMScriptRegistry is IEVMScriptRegistry, EVMScriptRegistryConstants, Ar
         DisableExecutor(_executorId, executorEntry.executor);
     }
 
+    function enableScriptExecutor(uint256 _executorId) external auth(REGISTRY_MANAGER_ROLE) {
```
izqui: Wondering if we should add a parameter to the `REGISTRY_MANAGER_ROLE` as it can now do 3 different actions with different security implications. Alternatively we can split off the enabling/disabling functionality into a separate role. Thoughts?

contracts/evmscript/EVMScriptRegistry.sol
```diff
@@ -44,6 +44,12 @@ contract EVMScriptRegistry is IEVMScriptRegistry, EVMScriptRegistryConstants, Ar
...
+        ExecutorEntry storage executorEntry = executors[_executorId];
+        executorEntry.enabled = true;
```
izqui: I think we should check that it is actually disabled and revert otherwise.

contracts/evmscript/EVMScriptRegistry.sol
sohkai: Agreed on the enabling / disabling aspect having different security assumptions.

sohkai:
Created a `REGISTRY_ADD_EXECUTOR_ROLE` for adding executors. Initially I was considering splitting all three of them out, as there may be cases where disabling executors should be a faster decision (e.g. security issues) than adding or re-enabling, but then I realized parameterization on the permission should handle this well.

One can simply apply looser rules for disabling executors above a certain ID, assuming those are non-core executors.

contracts/factory/EVMScriptRegistryFactory.sol
```diff
@@ -25,12 +25,13 @@ contract EVMScriptRegistryFactory is AppProxyFactory, EVMScriptRegistryConstants
...
-        acl.createPermission(this, reg, reg.REGISTRY_MANAGER_ROLE(), this);
+        acl.createPermission(this, reg, reg.REGISTRY_ADD_EXECUTOR_ROLE(), this);
 
         reg.addScriptExecutor(baseCallScript);     // spec 1 = CallsScript
 
-        acl.revokePermission(this, reg, reg.REGISTRY_MANAGER_ROLE());
-        acl.setPermissionManager(_root, reg, reg.REGISTRY_MANAGER_ROLE());
+        // Clean up the permissions
+        acl.revokePermission(this, reg, reg.REGISTRY_ADD_EXECUTOR_ROLE());
+        acl.removePermissionManager(reg, reg.REGISTRY_ADD_EXECUTOR_ROLE());
```
sohkai: @izqui Note that I've changed the factory to remove the permission's manager to reset it entirely, rather than leaving it dangling with root as the permission manager.

izqui: Thanks for pointing it out, I think it is a good idea. Created an issue as it requires changes in to kits: https://github.com/aragon/dao-kits/issues/23

contracts/evmscript/EVMScriptRegistry.sol
```diff
@@ -31,18 +33,32 @@ contract EVMScriptRegistry is IEVMScriptRegistry, EVMScriptRegistryConstants, Ar
         executors.push(ExecutorEntry(IEVMScriptExecutor(0), false));
     }
 
-    function addScriptExecutor(IEVMScriptExecutor _executor) external auth(REGISTRY_MANAGER_ROLE) returns (uint256 id) {
-        uint256 executorId = executors.push(ExecutorEntry(_executor, true));
+    function addScriptExecutor(IEVMScriptExecutor _executor) external auth(REGISTRY_ADD_EXECUTOR_ROLE) returns (uint256 id) {
+        uint256 executorId = executors.push(ExecutorEntry(_executor, true)) - 1;
```
sohkai: Note that this off-by-one was missed by everyone ;).

izqui: Good catch!

PR #394, opened 2018-08-22
Proxies: add checks for base being a contract

sohkai:
Adds more contract checks to the proxies.

Bytecode change:

```
AppProxyPinned.json Same                 0                  -18
KernelProxy.json    Same                 0                  +126
```

### DIFF
--- a/contracts/apps/AppProxyPinned.sol
+++ b/contracts/apps/AppProxyPinned.sol
@@ -1,10 +1,11 @@
 pragma solidity 0.4.18;
 
 import "../common/UnstructuredStorage.sol";
+import "../common/IsContract.sol";
 import "./AppProxyBase.sol";
 
 
-contract AppProxyPinned is AppProxyBase {
+contract AppProxyPinned is IsContract, AppProxyBase {
     using UnstructuredStorage for bytes32;
 
     // keccak256("aragonOS.appStorage.pinnedCode"), used by Proxy Pinned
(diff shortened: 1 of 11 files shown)


contracts/apps/AppProxyPinned.sol
```diff
@@ -21,7 +22,7 @@ contract AppProxyPinned is AppProxyBase {
         public // solium-disable-line visibility-first
     {
         setPinnedCode(getAppBase(_appId));
-        require(pinnedCode() != address(0));
+        require(isContract(pinnedCode()));
```
sohkai: Note that strictly, this isn't necessary as our `Kernel` doesn't let an app be set unless it's a contract (https://github.com/aragon/aragonOS/blob/dev/contracts/kernel/Kernel.sol#L177), so this is really more of an assertion (should we make it one?).

This also complicates the tests quite a bit, as you'll see in 961a7d7a81fcff0951225f95e72e8df697b5213d

izqui: I don't think an assertion makes sense here, as a contract that follows the `IKernel` interface may not check whether a base is a contract. A require is fine IMO.

test/mocks/KernelSetAppMock.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity 0.4.18;
+
+import "../../contracts/kernel/Kernel.sol";
+
+contract KernelSetAppMock is Kernel {
+    function KernelSetAppMock() Kernel(false) public {
+    }
+
+    function setApp(bytes32 _namespace, bytes32 _name, address _app) public returns (bytes32 id) {
```
izqui: We should add a comment explaining that we are bypassing the `isContract` check

sohkai:
Yeah... there is a way right now, by overloading the `isContract()` base... but that won't be an option if we turn it into a library (which we probably will after turning on the optimizer).

PR #395, opened 2018-08-24
Tests: wrap Solidity tests in JS wrapper

sohkai:
Our very own version of [truffle's solidity test runner](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-core/lib/testing/soliditytest.js): [`runSolidityTest`](https://github.com/aragon/aragonOS/blob/f03891553244a6fe49fb59e7129c4184481798d9/test/helpers/runSolidityTest.js).

The main motivation was to avoid inconsistent behaviour with how `solidity-coverage` instruments contracts and then manipulates their ABIs to be the same as before instrumentation (so that `.call()`s are unnecessary).

Solidity tests break the ABI switching `solidity-coverage` does because they _always_ get compiled when tests start, resulting in some files having their instrumented ABIs.

-----------------

All sorts of hacks are required to inject `Assert.sol` into the environment (see [migration](https://github.com/aragon/aragonOS/blob/f03891553244a6fe49fb59e7129c4184481798d9/migrations/4_test.js) and [pulling in the whole contract](https://github.com/aragon/aragonOS/blob/f03891553244a6fe49fb59e7129c4184481798d9/contracts/test/helpers/Assert.sol)).

If anything, this is a good reminder to avoid writing solidity tests in the future.

**Todo**:

- [ ] Fix some compile warnings in the test contracts


sohkai:
From #383:

> Functions in interfaces should be declared external: I wouldn't do anything about it, as some of the functions that we expose in interfaces need to be public.

Fixed in bb32ba1.

PR #397, opened 2018-08-25
Use auth modifier for createPermissions

izqui:
I guess we forgot to do it when we moved the ACL out of the Kernel

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -74,9 +74,10 @@ contract ACL is IACL, AragonApp, ACLHelpers {
     * @param _role Identifier for the group of actions in app given access to perform
     * @param _manager Address of the entity that will be able to grant and revoke the permission further.
     */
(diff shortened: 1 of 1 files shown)


bingen:
Hm, interesting. So it would be an ACL controlled by another ACL (the default one)?. What could be the use case for this? What a mess! :sweat_smile:
Anyway, this way looks more consistent. Besides, it would prevent an ACL to create a permission without being initialized and having a Kernel, right?

sohkai:
> So it would be an ACL controlled by another ACL (the default one)?

Yeah... that would be interesting. It'd be complex to use, but someone _could_ install two ACLs and have different apps use the second ACL directly

> Besides, it would prevent an ACL to create a permission without being initialized and having a Kernel, right?

This is a good catch; yes, before there was no initialization checks. There still aren't on any of the other external functions, but they're implicitly protected by `createPermission()`. I'll add more docs.

PR #401, opened 2018-08-27
Add ability to enable / disable depositable functionality

sohkai:
Before, only the App and Kernel proxies were able to receive ETH through generic `send()` or `transfer()`s. There was no way to control whether this was desired functionality, or a way to turn it off after a certain condition (e.g. ending a token sale).

This PR adds depositable functionality to the App and Kernel base contracts, as well as a way for Apps to turn off the ability to receive ETH.

- 816e166 and 01821c6: cleans up a bit of test code
- fec0a9a adds `DepositableStorage`
- 3d48720 uses `DepositableStorage` to control depositing functionality in the App and Kernel
- 027c522 and 1baa570 fix some "surprising" test changes due to the changes.
  - Notably, the recent changes have forced `APMRegistryFactory.newAPM()` to go over 6.9m gas again
  - `Kernel`s now have a fallback, so unregistered functions won't automatically revert


contracts/kernel/Kernel.sol
```diff
@@ -28,6 +29,10 @@ contract Kernel is IKernel, KernelStorage, Petrifiable, IsContract, VaultRecover
         }
     }
 
+    function () external payable {
+        require(isDepositable());
```
sohkai: To work around the changes in https://github.com/aragon/aragonOS/pull/401/files#diff-61de883632b625f81d4e28af414a9c90R64, we could also require that the value must be greater than 0 here and in `AragonApp`.

contracts/kernel/Kernel.sol
```diff
@@ -44,6 +49,7 @@ contract Kernel is IKernel, KernelStorage, Petrifiable, IsContract, VaultRecover
 
         acl.initialize(_permissionsCreator);
 
+        setDepositable(true);
```
izqui: I think depositing to the kernel should be disabled by default, given that there is no reason to send funds to the kernel for now. I think it is good that the `KernelProxy` keeps the functionality (in case the Kernel having ETH makes sense at some point), but it should be disabled IMO.

contracts/apps/AragonApp.sol
```diff
@@ -27,6 +28,10 @@ contract AragonApp is AppStorage, Autopetrified, VaultRecoverable, EVMScriptRunn
         _;
     }
 
+    function () external payable {
+        require(isDepositable());
```
izqui: What's the reason behind adding a fallback by default to all apps?  IMO apps should define their own fallback function if they want to receive ETH. It may seem weird that depending on the gas value, a deposit can happen or not, but I'd say deposits and recovers should be used as the last resort and when a call is already forwarded to app, its own logic should be responsible for deciding on what to do. 

In case we decide to keep this, we should definitely check for data being empty and value being greater than 0.

sohkai: Good point. Now that `DepositableDelegateProxy`s also need their depositing functionality explicitly enabled, I agree that apps should define their own fallback if they want to receive ETH.

contracts/kernel/Kernel.sol
sohkai: My main reason for enabling it was because Kernels are assigned ENS names through our initial templates, and so I could see users sending ETH directly to that name.

However, thinking about it, it's less confusing if the kernel reverts on receiving ETH directly than requiring a user to manually use the escape hatch functionality to transfer the funds to the vault. In the future, a different ENS name could be used for a finance or funding app, e.g. `finance.<dao name>.aragonid.eth`.

sohkai:
Updated to disable depositable functionality by default.

test/proxy_recover_funds.js
```diff
@@ -129,47 +130,26 @@ contract('Proxy funds', accounts => {
               const vaultProxyAddress = getEvent(receipt, 'NewAppProxy', 'proxy')
               vault = VaultMock.at(vaultProxyAddress)
             }
+            await vault.initialize()
+
             await kernel.setApp(APP_ADDR_NAMESPACE, vaultId, vault.address)
             await kernel.setRecoveryVaultAppId(vaultId)
           })
 
-          onlyBaseKernel(() => {
```
bingen: `onlyBaseKernel` and `onlyKernelProxy` are not used anymore, right? Then we can remove above (lines 81, 82)

contracts/test/mocks/KernelDepositableMock.sol
```diff
@@ -0,0 +1,17 @@
+pragma solidity 0.4.24;
```
bingen: I'm not sure that we really need this test, as we are testing something that our codebase doesn't offer, right? Besides I don't think I like the idea of a depositable kernel, but maybe it's because I'm missing some possible use cases.

test/proxy_recover_funds.js
```diff
@@ -129,47 +130,26 @@ contract('Proxy funds', accounts => {
               const vaultProxyAddress = getEvent(receipt, 'NewAppProxy', 'proxy')
               vault = VaultMock.at(vaultProxyAddress)
             }
+            await vault.initialize()
```
bingen: (I'm putting this comment here because the proper line has not been changed, which probably means that this comment actually belongs to a previous PR, haha)
If I'm not wrong, with these 2 nested `for`'s for Kernel and Vault, we are testing 4 possibilities. I don't see the point of testing proxied Kernel with a proxy-less Vault, or proxy-less Kernel with a proxied Vault. It's giving me hard time to think about these tests, haha ;-)

contracts/test/mocks/KernelDepositableMock.sol
sohkai: I'd like to keep it, as it's functionality that _could_ be enabled, even if it's not directly offered.

test/proxy_recover_funds.js
sohkai: Yes, it's a matrix of all the combinations.

I'd like to keep it, again to ensure that behaviour across all the combinations is the same. Even if it's rare, it simplifies the mental model–regardless of using a base or proxy, they should act the same. This PR moves even closer to that ideal, now that we can remove the `onlyIf` targeted tests.

.solcover.js
```diff
@@ -1,4 +1,10 @@
-const skipFiles = ['lib', 'test', 'acl/ACLSyntaxSugar.sol']
+const skipFiles = [
+  'lib',
+  'test',
+  'acl/ACLSyntaxSugar.sol',
+  'common/DepositableStorage.sol', // Used in tests that send ETH
+  'common/UnstructuredStorage.sol' // Used in tests that send ETH
```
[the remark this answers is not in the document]
sohkai: Checking to see if we can remove those from the ignored list... Originally the problem was that instrumentation in those files meant using unstructured storage in a fallback was more expensive than 2.3k gas (the instrumentation emits two events before and after each line of code in those files).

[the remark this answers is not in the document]
sohkai: Tried removing either one... and unfortunately coverage fails :(. I think they're not too important to get coverage stat on.

PR #413, opened 2018-09-03
Remove migrations and revamp scripts

sohkai:
We don't use these migrations to deploy anything, and they take up time when running tests.

Also revamped the deployment scripts to be more modular.


migrations/3_factory.js
```diff
@@ -1,24 +0,0 @@
-const getContract = name => artifacts.require(name)
```
izqui: This migration in particular we are using from some scripts that deploy kits, for example the [beta kits](https://github.com/aragon/aragon-apps/blob/master/templates/beta/migrations/2_deploy.js#L6)

The reason for doing this inside a migration in the past is that we needed library linking IIRC which we don't need anymore. Doing it in a migration is probably not the best way to do this, but this logic should be moved to a script

PR #414, opened 2018-09-03
Clean up truffle-config.js and lazy load providers

izqui:
Consolidates the configuration for different networks. It will now look for a key file for the network under `~/.aragon/[network]key.json` and use the keys and RPC specified in the file. If no keys are found, then it will default to using the user defined mnemonic or the default mnemonic. As for the RPC, it will use the one in the key file, and if that is not present, it will use the infura endpoint for the network.

Because we are now lazy loading providers, we can remove the need for having the `LIVE_NETWORKS` environment variable set to true when interacting with live networks, as if one of the networks that aren't being used cannot be connected to it is not a problem.

### DIFF
--- a/truffle-config.js
+++ b/truffle-config.js
@@ -61,7 +61,7 @@ module.exports = {
       port: 8545,
       gas: 6.9e6
     },
-    development: {
+    devnet: {
       network_id: 15,
       host: 'localhost',
       port: 8535,
--- a/truffle-config.js
+++ b/truffle-config.js
@@ -1,26 +1,47 @@
+const homedir = require('homedir')
+const path = require('path')
+
 const HDWalletProvider = require('truffle-hdwallet-provider')
 const HDWalletProviderPrivkey = require('truffle-hdwallet-provider-privkey')
 
-let mnemonic
-try {
-  mnemonic = require(require('homedir')()+'/.aragon/mnemonic.json').mnemonic
-} catch (e) {
-  mnemonic = 'stumble story behind hurt patient ball whisper art swift tongue ice alien';
-}
+const DEFAULT_MNEMONIC = 'stumble story behind hurt patient ball whisper art swift tongue ice alien'
 
-let ropstenProvider, kovanProvider, rinkebyProvider = {}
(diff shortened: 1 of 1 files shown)


truffle-config.js
```diff
@@ -1,27 +1,48 @@
...
+  return require(configFilePath('mnemonic.json').mnemonic)
   } catch (e) {
-    rinkebyProvider = new HDWalletProvider(mnemonic, 'https://rinkeby.infura.io')
+    return DEFAULT_MNEMONIC
   }
 }
 
+const settingsForNetwork = (network) => {
+  try {
+    return require(configFilePath(`${network}key.json`))
```
sohkai: Maybe using `_key.json` would be better?

truffle-config.js
```diff
@@ -1,27 +1,48 @@
...
+}
+
+// Lazily loaded provider
+const providerForNetwork = (network) => (
+  () => {
+    let { rpc, keys } = settingsForNetwork(network)
+
+    rpc = rpc || defaultRPC(network)
+
+    if (!keys || keys.lenght == 0) {
```
sohkai: Typo: `.length`

truffle-config.js
```diff
@@ -1,27 +1,48 @@
...
-  ropstenProvider = new HDWalletProvider(mnemonic, 'https://ropsten.infura.io/')
-  kovanProvider = new HDWalletProvider(mnemonic, 'https://kovan.infura.io')
+const configFilePath = (filename) =>
+  path.join(homedir(), `.aragon/${filename}`)
 
+const mnemonic = () => {
   try {
-    const { rpc, keys } = require(require('homedir')()+'/.aragon/rinkebykey.json')
-    rinkebyProvider = new HDWalletProviderPrivkey(keys, rpc)
+  return require(configFilePath('mnemonic.json').mnemonic)
```
bingen: I don't care that much, but shouldn't this be indented 2 to the right? ;-)

izqui:
I did end up renaming to `[network]_key.json`, @sohkai @bingen make sure to move your local files!

PR #416, opened 2018-09-03
Modular deployment scripts

sohkai:
Adds optional parameters to the deployment scripts to allow them to be used by other scripts being executed by `truffle exec`.

`deploy-beta-apm.js` automatically deploys a custom ENS if none is provided.

Also adds a script for deploying `DAOFactory`.

### DIFF
--- a/scripts/deploy-beta-apm.js
+++ b/scripts/deploy-beta-apm.js
@@ -1,63 +1,104 @@
 const namehash = require('eth-ens-namehash').hash
 const keccak256 = require('js-sha3').keccak_256
 
-const APMRegistryFactory = artifacts.require('APMRegistryFactory')
-const DAOFactory = artifacts.require('DAOFactory')
-const ENS = artifacts.require('ENS')
+const deployENS = require('./deploy-beta-ens')
 
-const owner = process.env.OWNER || '0x4cb3fd420555a09ba98845f0b816e45cfb230983'
-const ens = process.env.ENS || '0xfbae32d1cde62858bc45f51efc8cc4fa1415447e'
+const globalArtifacts = this.artifacts // Not injected unless called directly via truffle
 
-const tld = namehash('eth')
-const label = '0x'+keccak256('aragonpm')
-
-const getContract = name => artifacts.require(name)
+const defaultOwner = process.env.OWNER || '0x4cb3fd420555a09ba98845f0b816e45cfb230983'
+const defaultENS = process.env.ENS
 
 const baseInitArguments = {
   Kernel: [ true ] // petrify
 }
 
-const deployBases = async baseNames => {
-  const baseContracts = await Promise.all(baseNames.map(c => getContract(c).new(...(baseInitArguments[c] || []))))
-  return baseContracts.map(c => c.address)
+const deployBases = async baseContracts => {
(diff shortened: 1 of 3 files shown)


scripts/deploy-daofactory.js
```diff
@@ -0,0 +1,32 @@
+const globalArtifacts = this.artifacts // Not injected unless called directly via truffle
```
sohkai: Using `this.artifacts` is a bit of a hack; not sure how this is injected by truffle but `global.artifacts` doesn't work :(.

scripts/deploy-daofactory.js
```diff
@@ -0,0 +1,32 @@
...
+  log('Deploying DAOFactory with bases...')
+  const kernelBase = await Kernel.new(true) // immediately petrify
+  const aclBase = await ACL.new()
+  const evmScriptRegistryFactory = await EVMScriptRegistryFactory.new()
+  const daoFactory = await DAOFactory.new(kernelBase.address, aclBase.address, evmScriptRegistryFactory.address)
+  log('DAOFactory deployed:', daoFactory.address)
+
+  if (typeof truffleExecCallback === 'function') {
+    // Called directly via `truffle exec`
+    truffleExecCallback()
```
sohkai: Also a bit of a hack, to tell if we're executing at the root level of `truffle-exec` or only being used in another script.

scripts/deploy-beta-apm.js
```diff
@@ -1,63 +1,104 @@
...
+  log('Deploying APM...')
+  log('Owner:', owner)
 
-  console.log('deploying DAO bases')
-  const daoBases = await deployBases(['Kernel', 'ACL'])
-  console.log('deployed DAO bases', daoBases)
+  if (!ens) {
+    log('=========')
+    log('Missing ENS! Deploying a custom ENS...')
+    ens = (await deployENS(null, { artifacts, owner, verbose: false })).ens
```
bingen: Shouldn't the `verbose` param passed in here instead of `false`?

scripts/deploy-beta-apm.js
```diff
@@ -1,63 +1,104 @@
...
+  } else {
+    ens = ENS.at(ens)
```
bingen: Maybe we could rename `ens` param to `ensAddress`

scripts/deploy-beta-apm.js
```diff
@@ -1,63 +1,104 @@
...
-const getContract = name => artifacts.require(name)
+const defaultOwner = process.env.OWNER || '0x4cb3fd420555a09ba98845f0b816e45cfb230983'
+const defaultENS = process.env.ENS
 
 const baseInitArguments = {
   Kernel: [ true ] // petrify
 }
 
-const deployBases = async baseNames => {
-  const baseContracts = await Promise.all(baseNames.map(c => getContract(c).new(...(baseInitArguments[c] || []))))
```
bingen: Hm, you don't like `getContract`, right? ;-)

scripts/deploy-beta-apm.js
sohkai: I wanted to silence all the logging in `deployENS()`, so that this script can control what gets logged.

PR #417, opened 2018-09-04
lib: optimize multiplication for SafeMath64 and SafeMath8

bingen:
Plus add tests for them.

Small optimizations avoiding one extra division. Tested on Remix.

### DIFF
--- a/contracts/lib/math/SafeMath64.sol
+++ b/contracts/lib/math/SafeMath64.sol
@@ -15,7 +15,7 @@ library SafeMath64 {
     */
     function mul(uint64 _a, uint64 _b) internal pure returns (uint64) {
         uint256 c = uint256(_a) * uint256(_b);
-        require(c < 0x010000000000000000);
+        require(c < 0x010000000000000000); // 2**64 (less gas this way)
 
         return uint64(c);
     }
--- a/contracts/lib/math/SafeMath64.sol
+++ b/contracts/lib/math/SafeMath64.sol
@@ -14,17 +14,10 @@ library SafeMath64 {
     * @dev Multiplies two numbers, reverts on overflow.
     */
     function mul(uint64 _a, uint64 _b) internal pure returns (uint64) {
-        // Gas optimization: this is cheaper than requiring 'a' not being zero, but the
-        // benefit is lost if 'b' is also tested.
-        // See: https://github.com/OpenZeppelin/openzeppelin-solidity/pull/522
-        if (_a == 0) {
-            return 0;
-        }
+        uint256 c = uint256(_a) * uint256(_b);
+        require(c < 0x010000000000000000);
 
-        uint64 c = _a * _b;
-        require(c / _a == _b);
-
-        return c;
+        return uint64(c);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


sohkai:
Let's push this upstream to zeppelin too!

contracts/lib/math/SafeMath64.sol
```diff
@@ -14,17 +14,10 @@ library SafeMath64 {
...
     */
     function mul(uint64 _a, uint64 _b) internal pure returns (uint64) {
-        // Gas optimization: this is cheaper than requiring 'a' not being zero, but the
-        // benefit is lost if 'b' is also tested.
-        // See: https://github.com/OpenZeppelin/openzeppelin-solidity/pull/522
-        if (_a == 0) {
-            return 0;
-        }
+        uint256 c = uint256(_a) * uint256(_b);
+        require(c < 0x010000000000000000);
```
sohkai: Maybe we could use a `MAX_UINT64` constant here instead?

test/lib_safemath64.js
```diff
@@ -0,0 +1,54 @@
...
+  })
+
+  it('fails if addition overflows', async () => {
+    const a = new web3.BigNumber(2).pow(63)
+    const b = new web3.BigNumber(2).pow(63)
+    return assertRevert(async () => {
+      await safeMath64Mock.addExt(a, b)
+    })
+  })
+})
```
sohkai: I guess we might as well add a division case?

sohkai:
We should probably also add a note for the optimization (and how it differs from the zeppelin lib) in each file.

bingen:
Yes, my attempt with 256 version was a complete failure! Haha. I guess with a more complete instruction set  (things like `ffs`) we could do better. I have some other ideas, but someday. It just bothers me a little bit making a division for each multiplication ;-). Although the fact that multiplication and division have the same gas cost (which seems unfair to me) doesn't help either.

test/lib_safemath64.js
bingen: I didn't add it because I actually I wanted to discuss about it with you: shouldn't we remove division instead? I don't think we are using it, and Solidity is supposed to throw when dividing by zero.

contracts/lib/math/SafeMath64.sol
bingen: I tried several options on Remix (like that one, `1 << 64`, `0xFFFFFFFFFFFFFFFF`, `2**64`) and this one seemed to the be the one using less gas.

sohkai: Oh weird, the constant should just get replaced by the compiler.

In that case, a comment would help :).

test/lib_safemath64.js
sohkai: Solidity automatically `asserts`, but we should `revert` since all the other operations are reverting.

contracts/lib/math/SafeMath64.sol
bingen: Yes, that what I thought, but it seems not. I'll add a comment.

test/lib_safemath64.js
bingen: I did a quick test on Remix and gas execution cost went from `359` to `539` from just dividing to using that SafeMathlib function. I don't think it's worth it. Dividing by zero should be the exception so we would be losing gas in most of the cases.

sohkai: Yeah there's definitely some overhead, but I'd strongly suggest keeping it to simplify the mental model when using SafeMath. You know you're protected by a revert when you use it, in the worst case scenario, rather than an assert that will eat up all the gas.

PR #422, opened 2018-09-06
Deployment: use DAOFactory deployment in APM deployment

sohkai:
Makes the DAOFactory deployment script accept a parameter to not deploy an `EVMScriptRegistryFactory`, and changes the APM deployment script to directly use the DAOFactory deployment script.

### DIFF
--- a/scripts/deploy-beta-apm.js
+++ b/scripts/deploy-beta-apm.js
@@ -6,6 +6,8 @@ const deployDaoFactory = require('./deploy-daofactory')
 
 const globalArtifacts = this.artifacts // Not injected unless called directly via truffle
 
+const ZERO_ADDR = '0x0000000000000000000000000000000000000000'
+
 const defaultOwner = process.env.OWNER || '0x4cb3fd420555a09ba98845f0b816e45cfb230983'
 const defaultDaoFactoryAddress = process.env.DAO_FACTORY
 const defaultENSAddress = process.env.ENS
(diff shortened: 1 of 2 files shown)


scripts/deploy-beta-apm.js
```diff
@@ -68,14 +65,15 @@ module.exports = async (
...
-  log('Deploying DAO bases...')
-  const daoBases = await deployBases([Kernel, ACL])
-  log('Deployed DAO bases', daoBases)
-
-  log('Deploying DAOFactory without EVMScripts...')
-  const evmScriptRegistry = '0x00' // Basic APM needs no forwarding
-  const daoFactory = await DAOFactory.new(...daoBases, evmScriptRegistry)
-  log('Deployed DAOFactory:', daoFactory.address)
+  let daoFactory
+  if (daoFactoryAddress) {
```
sohkai: I guess we could also check if the `daoFactory` has EVMScripts enabled, and go back to creating one if it doesn't.

scripts/deploy-beta-apm.js
```diff
@@ -68,14 +65,15 @@ module.exports = async (
...
+    log('Using provided DAOFactory:', daoFactoryAddress)
+    daoFactory = DAOFactory.at(daoFactoryAddress)
+  } else {
+    log('Deploying DAOFactory without EVMScripts...')
```
sohkai: Should be updated to **with~~out~~**

scripts/deploy-beta-apm.js
izqui: I think not checking this is actually a feature, as someone may decide to deploy an APM instance without scripts. However, I think we should log that.

PR #423, opened 2018-09-07
Typo fix

pcowgill:


sohkai:
@pcowgill We need to you to sign the CLA first before we can merge, but yes, you'll notice there's a lot of places where we misuse `specially` instead of `especially` ;)

pcowgill:
@sohkai Signed!

PR #424, opened 2018-09-12
Fix visibility order for constants

sohkai:
Generally, visibility modifers are preferred before any other modifier (strong social consensus, e.g. OpenZeppelin).

Our linter doesn't have a rule to check this yet (see https://github.com/duaraghav8/Solium/issues/239).

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -31,12 +31,12 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
         // op and id take less than 1 byte each so it can be kept in 1 sstore
     }
 
-    uint8 constant BLOCK_NUMBER_PARAM_ID = 200;
-    uint8 constant TIMESTAMP_PARAM_ID    = 201;
+    uint8 constant internal BLOCK_NUMBER_PARAM_ID = 200;
+    uint8 constant internal TIMESTAMP_PARAM_ID    = 201;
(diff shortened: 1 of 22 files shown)


contracts/acl/ACL.sol
```diff
@@ -31,20 +31,21 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
-    uint8 constant TIMESTAMP_PARAM_ID    = 201;
+    uint8 constant internal BLOCK_NUMBER_PARAM_ID = 200;
+    uint8 constant internal TIMESTAMP_PARAM_ID    = 201;
     // 202 is unused
-    uint8 constant ORACLE_PARAM_ID       = 203;
-    uint8 constant LOGIC_OP_PARAM_ID     = 204;
-    uint8 constant PARAM_VALUE_PARAM_ID  = 205;
+    uint8 constant internal ORACLE_PARAM_ID       = 203;
+    uint8 constant internal LOGIC_OP_PARAM_ID     = 204;
+    uint8 constant internal PARAM_VALUE_PARAM_ID  = 205;
```
sohkai: We could expose these param IDs publicly, but I don't think anyone would ever need to use it if we exposed a contract that defined these IDs as internal constants.

bingen: Shouldn't all of these be `internal constant`?

PR #425, opened 2018-09-13
Add test for recevering ETH from Kernel

bingen:


sohkai:
@bingen This is likely failing due to the gas limit on sending ETH being blown by coverage; worst case we can disable this test on coverage.

test/proxy_recover_funds.js
```diff
@@ -223,4 +224,35 @@ contract('Proxy funds', accounts => {
       }
     })
   }
+  // Kernel is not depositable by default, but in order to test transferToVault,
```
sohkai: Let's add a space above this :)

test/proxy_recover_funds.js
```diff
@@ -223,4 +224,35 @@ contract('Proxy funds', accounts => {
...
+  // we create a mockup to make it depositable
+  context('Depositable Kernel', async () => {
```
sohkai: I think this should be "KernelProxy"

PR #429, opened 2018-09-24
Add `petrifyPermission` function to ACL

bingen:


contracts/acl/ACL.sol
```diff
@@ -44,6 +44,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     bytes32 public constant EMPTY_PARAM_HASH = 0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563;
     bytes32 public constant NO_PERMISSION = bytes32(0);
     address public constant ANY_ENTITY = address(-1);
+    address public constant PETRIFY_ENTITY = address(1);
```
izqui: We should explain why it needs to be `address(1)`

contracts/acl/ACL.sol
```diff
@@ -156,6 +157,18 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
         _setPermissionManager(address(0), _app, _role);
     }
 
+    /**
+    * @notice Lock `_role` in `_app`, so no midification can be made to it (grant, revoke, permission manager)
+    * @param _app Address of the app in which the permission is being unmanaged
+    * @param _role Identifier for the group of actions being unmanaged
+    */
+    function petrifyPermission(address _app, bytes32 _role)
```
izqui: Thinking about the name, I think `petrify` could be confused with the aragonOS concept of petrifying apps. I think `lock` could be a better name, it is what you actually used to explain it in the `@notice`.

contracts/acl/ACL.sol
```diff
@@ -156,6 +157,18 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
         _setPermissionManager(address(0), _app, _role);
     }
 
+    /**
+    * @notice Lock `_role` in `_app`, so no midification can be made to it (grant, revoke, permission manager)
```
izqui: `after locking, no mOdifications can be made to the permission`

contracts/acl/ACL.sol
```diff
@@ -156,6 +157,18 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
+    * @param _app Address of the app in which the permission is being unmanaged
+    * @param _role Identifier for the group of actions being unmanaged
+    */
+    function petrifyPermission(address _app, bytes32 _role)
+        external
+        onlyPermissionManager(_app, _role)
```
izqui: Thinking about whether this action should be done by the manager, or also allow to do it on permissions that haven't been created yet. It feels a bit off that you would need to create a permission and then immediately lock it, as after creating the permission you would have to revoke it from the entity that got the permission so it is completely clean and no one has the permission.

bingen: You can create the permission using `LOCK_ENTITY` (i.e., `address(1)`) as manager and granted but it's certainly not so clean, so yes, changing it.

bingen: Btw, would it make sense to add helpers/wrappers like `createLockedPermission` or `createEmptyLockedPermission`? (they would pass `LOCK_ENTITY` as granted or manager)

contracts/acl/ACL.sol
```diff
@@ -156,6 +157,30 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
+    /**
+    * @notice Lock non-existing yet `_role` in `_app`, so no modification can be made to it (grant, revoke, permission manager)
+    * @param _app Address of the app in which the permission is being locked
+    * @param _role Identifier for the group of actions being locked
+    */
+    function createLockedPermission(address _app, bytes32 _role)
+        external
+        auth(CREATE_PERMISSIONS_ROLE)
+    {
+        _setPermissionManager(LOCK_ENTITY, _app, _role);
```
sohkai: This should also check `require(getPermissionManager(_app, _role) == address(0));`

test/kernel_acl.js
```diff
@@ -319,6 +321,70 @@ contract('Kernel ACL', accounts => {
...
+
+            context('> locking permission', () => {
+                const MOCK_ROLE = keccak256("MOCK_ROLE")
+                let LOCK_ENTITY
+
+                before(async () => {
+                    LOCK_ENTITY = await acl.LOCK_ENTITY()
+                })
+
+                it('petrifies existing permission', async () => {
```
sohkai: A reminder to change some of the prior nomenclature in the tests :).

contracts/acl/ACL.sol
bingen: Good catch, thanks! I accidentally lost it during the split in last commit: https://github.com/aragon/aragonOS/pull/429/commits/9a89419c3cd053b444762de64ee1310ab044d536#diff-325598653bbd8046a79c33ddabaa86e1L170
Thinking about it, as the same condition is being used in `_createPermission`, and for consistency with some of the recent changes for aragon OS 4 / audit freeze, should I move that into a modifier and use it in external functions instead of that internal one?

test/kernel_acl.js
```diff
@@ -375,13 +375,23 @@ contract('Kernel ACL', accounts => {
...
+                    })
+                })
+
+                it('fails trying to create a burned permission which already has a manager', async () => {
+                    // create permission
+                    await acl.createPermission(granted, kernelAddr, MOCK_ROLE, granted, { from: permissionsRoot })
+
+                    return assertRevert(async () => {
+                        // try to create it burnt
+                        await acl.createBurnedPermission(kernelAddr, MOCK_ROLE, { from: permissionsRoot })
```
sohkai: I'd add another test either here, or separately, where we then revoke the only grantee of the permission and test that `acl.createBurnedPermission()` still fails.

contracts/acl/ACL.sol
```diff
@@ -19,6 +19,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     mapping (bytes32 => Param[]) internal permissionParams; // params hash => params
 
     // Who is the manager of a permission
+
```
sohkai: Nitpick: this extra line seems unnecessary :).

contracts/acl/ACL.sol
sohkai: Really like https://github.com/aragon/aragonOS/pull/429/commits/28df834b7b932499dc6992626474405af49870ec!

test/kernel_acl.js
```diff
@@ -319,6 +321,88 @@ contract('Kernel ACL', accounts => {
...
+                })
+
+                it('burns existing permission', async () => {
+                    // create permission
+                    await acl.createPermission(granted, kernelAddr, MOCK_ROLE, granted, { from: permissionsRoot })
+
+                    // burn it
+                    const receipt = await acl.burnPermissionManager(kernelAddr, MOCK_ROLE, { from: granted })
+                    assertEvent(receipt, 'ChangePermissionManager')
+
```
izqui: Missing an assertion that the permission manager is actually `BURN_ENTITY`.

sohkai: @bingen Thoughts on using `acl.getPermissionManager()` to check in both the `'burns existing permission'` and `'burns non-existing permission'` instead of the event?

PR #430, opened 2018-09-27
Change Repo version storage from an array to a mapping

izqui:
Intentionally didn't modify the tests to make sure the internal change doesn't affect how the repo works.

### DIFF
--- a/contracts/apm/Repo.sol
+++ b/contracts/apm/Repo.sol
@@ -12,7 +12,8 @@ contract Repo is AragonApp {
         bytes contentURI;
     }
 
-    Version[] public versions;
+    uint256 internal versionsNextIndex;
+    mapping (uint256 => Version) internal versions;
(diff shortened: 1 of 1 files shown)


contracts/apm/Repo.sol
```diff
@@ -27,6 +28,7 @@ contract Repo is AragonApp {
     */
     function initialize() public onlyInit {
         initialized();
+        versionsNextIndex = 1;
```
sohkai: Wonder if we could just use `versionsLength`, since we don't need to reserve the 0 slot?

izqui: It is needed for [these reverse mappings](https://github.com/aragon/aragonOS/pull/430/files#diff-460eb60e27bc67ff984d38f5b91246fdR66) to work, so if you ask for a version with an inexistent semantic version or inexistent contract address, it reverts.

sohkai: :+1:

PR #432, opened 2018-09-28
Optimize proxies and AragonApp by removing unnecessary constants from bytecode

izqui:
Leaving constants in the bytecode that are not used is quite expensive, in an attempt to optimize bytecode size for proxies, removing constants has had good results. Using @sohkai's bytecode comparison tool: 

After applying 0bf39da and edd8120
```
APMRegistry.json               119600 less gas      -598               0
AppProxyFactory.json           119600 less gas      -598               0
AppProxyPinned.json            54400 less gas       -272               -54
AppProxyPinnedStorageMock.json 54400 less gas       -272               0
AppProxyUpgradeable.json       54400 less gas       -272               0
DAOFactory.json                40800 less gas       -204               0
EVMScriptRegistryFactory.json  119600 less gas      -598               0
Kernel.json                    119600 less gas      -598               0
KernelDepositableMock.json     119600 less gas      -598               0
KernelPinnedStorageMock.json   119600 less gas      -598               0
KernelProxy.json               40800 less gas       -204               0
KernelSetAppMock.json          119600 less gas      -598               0
KernelStorage.json             69800 less gas       -349               0
UpgradedKernel.json            119600 less gas      -598               0
```

Applying 9e763ca as well results in:
```
ACL.json                       54400 less gas       -272               0
APMRegistry.json               177200 less gas      -886               0
AppProxyFactory.json           119600 less gas      -598               0
AppProxyPinned.json            54400 less gas       -272               -54
AppProxyPinnedStorageMock.json 54400 less gas       -272               0
AppProxyUpgradeable.json       54400 less gas       -272               0
AragonApp.json                 54400 less gas       -272               0
DAOFactory.json                40800 less gas       -204               0
ENSSubdomainRegistrar.json     54400 less gas       -272               0
EVMScriptRegistry.json         54400 less gas       -272               0
EVMScriptRegistryFactory.json  119600 less gas      -598               -272
EVMScriptRunner.json           54400 less gas       -272               0
Kernel.json                    119600 less gas      -598               0
KernelPinnedStorageMock.json   119600 less gas      -598               0
KernelProxy.json               40800 less gas       -204               0
KernelStorage.json             69800 less gas       -349               0
Repo.json                      54400 less gas       -272               0
```

### DIFF
--- a/contracts/apps/AppStorage.sol
+++ b/contracts/apps/AppStorage.sol
@@ -15,8 +15,6 @@ contract AppStorage {
     bytes32 internal constant KERNEL_POSITION = 0x4172f0f7d2289153072b0a6ca36959e0cbe2efc3afe50fc81636caa96338137b;
     // keccak256("aragonOS.appStorage.appId")
     bytes32 internal constant APP_ID_POSITION = 0xd625496217aa6a3453eecb9c3489dc5a53e6c67b444329ea2b2cbc9ff547639b;
-    // keccak256("aragonOS.appStorage.pinnedCode"), used by Proxy Pinned
-    bytes32 internal constant PINNED_CODE_POSITION = 0xdee64df20d65e53d7f51cb6ab6d921a0a6a638a91e942e1d8d02df28e31c038e;
 
     function kernel() public view returns (IKernel) {
         return IKernel(KERNEL_POSITION.getStorageAddress());
--- a/contracts/apps/AppProxyBase.sol
+++ b/contracts/apps/AppProxyBase.sol
@@ -6,7 +6,7 @@ import "../kernel/KernelConstants.sol";
 import "../kernel/IKernel.sol";
 
 
-contract AppProxyBase is AppStorage, DepositableDelegateProxy, KernelConstants {
+contract AppProxyBase is AppStorage, DepositableDelegateProxy, KernelConstantsAppBases {
     /**
     * @dev Initialize AppProxy
     * @param _kernel Reference to organization kernel for the app
--- a/contracts/evmscript/EVMScriptRunner.sol
+++ b/contracts/evmscript/EVMScriptRunner.sol
@@ -12,7 +12,7 @@ import "../kernel/KernelConstants.sol";
 import "../common/Initializable.sol";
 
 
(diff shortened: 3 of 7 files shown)


sohkai:
There's a few linting errors that should be fixed, but I'm also wondering if we should just make all these constants `internal` and only expose some of them publicly in `Kernel`. The namespaces are important to expose, but I'm not sure if the app IDs are really important.

contracts/apps/AppStorage.sol
```diff
@@ -15,8 +15,6 @@ contract AppStorage {
     bytes32 internal constant KERNEL_POSITION = 0x4172f0f7d2289153072b0a6ca36959e0cbe2efc3afe50fc81636caa96338137b;
     // keccak256("aragonOS.appStorage.appId")
     bytes32 internal constant APP_ID_POSITION = 0xd625496217aa6a3453eecb9c3489dc5a53e6c67b444329ea2b2cbc9ff547639b;
-    // keccak256("aragonOS.appStorage.pinnedCode"), used by Proxy Pinned
-    bytes32 internal constant PINNED_CODE_POSITION = 0xdee64df20d65e53d7f51cb6ab6d921a0a6a638a91e942e1d8d02df28e31c038e;
```
bingen: oh, good catch!

sohkai:
> The namespaces are important to expose, but I'm not sure if the app IDs are really important.

Something I've been thinking about is exposing these app IDs as part of the apps' external interfaces. They are actually necessary, since you need it to upgrade an app and it [could save us some hassle in the DAO Kits](https://github.com/aragon/dao-kits/blob/master/kits/beta/contracts/BetaTemplateBase.sol#L50).

For most of the other apps, we could just expose an `bytes32 public constant APP_ID = <apm hash>`, but for the `ACL` and `Kernel`, where we need the constants without a contract instance, we could declare public functions that expose an internal constant.

izqui:
> is exposing these app IDs as part of the apps' external interfaces

I'm not so sure about this, as AppProxy already has a public `appId` and they could potentially be different (unless we enforced that they need to match). Also by hardcoding `APP_ID` in the contract code, it would require extra tooling for deploying the same app with different appId's (for example, staging vs production) and you couldn't 'migrate' an app between APM registries or repos without recompiling the code.

sohkai:
Superseded by #443.

PR #433, opened 2018-09-28
Remove redundant kernel check in AragonApp.getRecoveryVault()

izqui:
When using the native call syntax `target.method(...)` solc introduces
a check for the code size of the target (and reverts if it is 0), therefore
we don't need to manually ensure that the target address is not 0 before making
a call, as the size of the address 0 will be 0, and therefore already revert

### DIFF
--- a/contracts/kernel/Kernel.sol
+++ b/contracts/kernel/Kernel.sol
@@ -193,8 +193,7 @@ contract Kernel is IKernel, KernelStorage, Petrifiable, IsContract, VaultRecover
     */
     function hasPermission(address _who, address _where, bytes32 _what, bytes _how) public view returns (bool) {
         IACL defaultAcl = acl();
-        return address(defaultAcl) != address(0) && // Poor man's initialization check (saves gas)
-            defaultAcl.hasPermission(_who, _where, _what, _how);
+        return defaultAcl.hasPermission(_who, _where, _what, _how); // if acl is not set, it will revert
     }
 
     function _setApp(bytes32 _namespace, bytes32 _appId, address _app) internal {
--- a/contracts/kernel/Kernel.sol
+++ b/contracts/kernel/Kernel.sol
@@ -193,7 +193,8 @@ contract Kernel is IKernel, KernelStorage, Petrifiable, IsContract, VaultRecover
     */
     function hasPermission(address _who, address _where, bytes32 _what, bytes _how) public view returns (bool) {
         IACL defaultAcl = acl();
-        return defaultAcl.hasPermission(_who, _where, _what, _how); // if acl is not set, it will revert
+        return address(defaultAcl) != address(0) && // Poor man's initialization check (saves gas)
+            defaultAcl.hasPermission(_who, _where, _what, _how);
     }
 
     function _setApp(bytes32 _namespace, bytes32 _appId, address _app) internal {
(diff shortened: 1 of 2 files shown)


izqui:
There are also minor gas savings here due to removing some logic:

```
ACL.json                       8400 less gas        -42                0
APMRegistry.json               9000 less gas        -45                0
AppStub.json                   8400 less gas        -42                0
AppStub2.json                  8400 less gas        -42                0
AppStubConditionalRecovery.js… 8400 less gas        -42                0
AppStubDepositable.json        8400 less gas        -42                -17
AragonApp.json                 8400 less gas        -42                0
ENSSubdomainRegistrar.json     8400 less gas        -42                0
EVMScriptRegistry.json         8400 less gas        -42                0
EVMScriptRegistryFactory.json  Same                 0                  -42
Kernel.json                    6400 less gas        -32                0
MockScriptExecutorApp.json     8400 less gas        -42                0
Repo.json                      8400 less gas        -42                0
TestACLInterpreter.json        8400 less gas        -42                0
UnsafeAppStub.json             8400 less gas        -42                0
UnsafeAppStubDepositable.json  8400 less gas        -42                0
UnsafeAragonApp.json           8400 less gas        -42                0
UnsafeAragonAppMock.json       8400 less gas        -42                0
UnsafeRepo.json                8400 less gas        -42                0
VaultMock.json                 8400 less gas        -42                0
```

contracts/kernel/Kernel.sol
```diff
@@ -193,8 +193,7 @@ contract Kernel is IKernel, KernelStorage, Petrifiable, IsContract, VaultRecover
     */
     function hasPermission(address _who, address _where, bytes32 _what, bytes _how) public view returns (bool) {
         IACL defaultAcl = acl();
-        return address(defaultAcl) != address(0) && // Poor man's initialization check (saves gas)
-            defaultAcl.hasPermission(_who, _where, _what, _how);
+        return defaultAcl.hasPermission(_who, _where, _what, _how); // if acl is not set, it will revert
```
sohkai: ⚠️ In this case I don't think we want to revert, but to return `false`, so we still need this case.

izqui: Great catch :)

izqui: Actually, almost everywhere else we are reverting in getters when data doesn't exist or it is not initialized. Was there any other reason `kernel#hasPermission` returned false rather than failing?

bingen: Not sure about this, but the only way `defaultACL` is `0` is if `Kernel` has not been initialized, right? Would it make sense to call `hasPermission` on a non-initialized kernel?
About the getters, that's true at least for out-of-bounds arrays and mappings mimicking arrays. We did that because it's the default compiler behaviour.

sohkai: I think it makes sense for an uninitialized `Kernel` to respond with `false` when asked for permission, since an uninitialized kernel should not allow anything and this is a _read_ operation. An uninitialized `ACL` acts similarly; you can't _write_ any permissions, but all _read_ attempts will return false.

Another reason I'd advocate for it to keep returning only a boolean, rather than throwing, is because `hasPermission()` is part of the `IKernel` interface. We could, of course, add to the interface that one may revert based on certain conditions, but this gets more and more tricky for consumers to handle.

> Actually, almost everywhere else we are reverting in getters when data doesn't exist or it is not initialized.

As @bingen said, this should be only in cases where we mimic array behaviour.

izqui:
Fair points, reverted the change in `hasPermission`

PR #436, opened 2018-10-04
KernelConstants: use nonstrict pragma

0mkara:
Use non-strict version to be able to import aragon contracts on Remix & Etheratom.

`KernelConstants` is eventually used by `AragonApp`.


0mkara:
@sohkai I was importing `AragonApp.sol`. Thing is currently both RemixIDE & EtheratomIDE both selects one version of solidity before they compile. So if a solidity program has different versions used by its imported contracts compilation fails.

sohkai:
@0mkara I've just added our MIT licensing to the unpinned files :).

PR #439, opened 2018-10-14
Improve deployment to ease verification and documentation

izqui:
Builds on #422

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -5,7 +5,7 @@ This repo contains Aragon's reference implementation for [aragonOS](https://hack
 #### 🚨 Security review status: bug bounty
 aragonOS 4 has undergone two independent professional security reviews, and the issues raised have been resolved. However there is a [bug bounty program](https://wiki.aragon.org/dev/bug_bounty/) for rewarding hackers who find security vulnerabilities. There is a bounty pool of $250,000 USD, you can find more information [here](https://wiki.aragon.org/dev/bug_bounty/).
 
-#### 👋 We are tagging tasks that are [good first issues](https://github.com/aragon/aragonOS/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) so you can get started contributing to Aragon Core.
+#### 👋 Get started contributing with a [good first issue](https://github.com/aragon/aragonOS/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22).
 Don't be shy to contribute even the smallest tweak. Everyone will be especially nice and helpful to beginners to help you get started!
 
 ## Documentation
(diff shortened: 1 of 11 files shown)


readme.md
```diff
@@ -1,19 +1,16 @@
...
 
 #### 👋 We are tagging tasks that are [good first issues](https://github.com/aragon/aragonOS/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) so you can get started contributing to Aragon Core.
 Don't be shy to contribute even the smallest tweak. Everyone will be especially nice and helpful to beginners to help you get started!
 
 ## Documentation
 
-Visit the [Aragon wiki](https://wiki.aragon.one/dev/aragonOS_index/) for in depth documentation on the [architecture](https://wiki.aragon.one/documentation/aragonOS/) and different parts of the system.
+Visit the [Aragon Developer Portal](https://hack.aragon.org/docs/aragonos-intro.html) for in depth documentation on the [architecture](https://hack.aragon.org/docs/aragonos-ref.html) and different parts of the system.
 
 ## Installing aragonOS
```
sohkai: While we're at it, I think we should change the heading of this too, to be something like "Get started contributing" under the "Contributing" section at the bottom.

Installing `aragonOS` is literally `npm install`.

scripts/deploy-apm.js
```diff
@@ -99,7 +106,10 @@ module.exports = async (
 
   log('=========')
   const apmAddr = receipt.logs.filter(l => l.event == 'DeployAPM')[0].args.apm
-  log('Deployed APM:', apmAddr)
+  log('# APM:')
+  log('Address:', apmAddr)
+  log('Transaction hash:', receipt.tx)
+  log('=========')
   log(apmAddr)
```
sohkai: Probably don't need this trailing bit anymore.

scripts/helpers/create-logger.js
```diff
@@ -0,0 +1,39 @@
+const flatten = require('truffle-flattener')
```
sohkai: Seems like this file would be better named as "deploy-logger"

scripts/helpers/create-logger.js
```diff
@@ -0,0 +1,39 @@
...
+const mkdirp = require('mkdirp')
+const fs = require('fs')
+const path = require('path')
+
+const truffleConfig = require('../../truffle-config')
```
sohkai: Using `path.resolve()` would be better

scripts/deploy-apm.js
```diff
@@ -76,12 +77,18 @@ module.exports = async (
...
+    apmRegistryBase.address,
+    apmRepoBase.address,
+    ensSubdomainRegistrarBase.address,
+    ensAddress,
+    '0x00'
+  )
+  await logDeploy(apmFactory, { verbose })
 
   log(`Assigning ENS name (${labelName}.${tldName}) to factory...`)
   try {
```
sohkai: Tried `truffle exec --network rpc scripts/deploy-apm.js` by itself and hit the ENS error here.

Maybe we should also set `aragonpm.eth` if we deploy our own instance of ENS in this script?

scripts/deploy-daofactory.js
```diff
@@ -10,31 +12,30 @@ module.exports = async (
...
+    await logDeploy(evmScriptRegistryFactory, { verbose })
   }
   const daoFactory = await DAOFactory.new(
     kernelBase.address,
     aclBase.address,
     evmScriptRegistryFactory ? evmScriptRegistryFactory.address : ZERO_ADDR
   )
 
-  log('DAOFactory deployed:', daoFactory.address)
+  logDeploy(daoFactory, { verbose })
```
sohkai: I think this should have an `await`?

scripts/deploy-apm.js
izqui: It blows up if you don't pass the first account as the `OWNER` env variable, fixing it.

izqui: Added this in 04813a1:

```
...
OWNER env variable not found, setting APM owner to the provider\'s first account
Owner: 0xb4124ceb3451635dacedd11767f004d8a28c6ee7
...
```

If you run `truffle exec --network rpc scripts/deploy-apm.js` now it should just work.

Issue #438, opened 2018-10-11
Revert with reason


izqui:
On mainnet having revert reasons will help us debug issues when transactions revert.

There is no clear standard for how this should be done, other projects seem to just be returning error strings. Adding long strings bloats the contract bytecode and the cost for deployment.

I think we should at least add revert reasons for everything that may cause a revert in `AppProxy` and `KernelProxy`, as that code will remain static. At the same time, the proxies should be as small as possible so they are cheap to deploy, so we should be careful with what errors we return.

`DelegateProxy` forwards the error data if a revert occurs in the `delegatecall`, so if an app reverts with reason the error message would propagate up the call stack. We should still test that this is indeed behaving as we expect (**Edit:** tested, it works).

sohkai:
@izqui I was thinking that, as a potential way to avoid spoofing of the error messages, the `DelegateProxy` could detect if an error happened and return a differently formatted `returndata`.

Instead of just:

```solidity
let ptr := mload(0x40)
returndatacopy(ptr, 0, size)
switch result case 0 { revert(ptr, size) }
```

We could encode something like `1<error data>` (where `1` is the "app" error namespace and `0` could be the proxy's own error namespace).

---------------

However, this might be too much, given we can always look at the debug trace and also spot if it was the app's base contract failing by the gas cost.

izqui:
For simplicity I would avoid manipulating the returned error strings as the returned data is ABI encoded, and for adding a new char we would need to decode and re-encode the data.

The spoofing we are trying to protect against is apps returning errors that are the same as the proxy errors, right? Since the proxy can only revert before performing the the delegatecall, the spoofing would be pretty easy to detect, as native proxy errors are only 'legit' if the proxy doesn't perform a `delegatecall` (which can be easily seen by analyzing a debug trace).

izqui:
After implementing them, I changed my mind and I think we should have revert reasons everywhere but on proxies (See https://github.com/aragon/aragonOS/pull/440#issuecomment-430323016)

PR #440, opened 2018-10-16
Add revert reasons to proxies

izqui:
Partially solves #438 

The naming of errors optimizes for proxy size, given that having strings in contract bytecode is quite expensive.

### DIFF
--- a/contracts/apps/AppProxyBase.sol
+++ b/contracts/apps/AppProxyBase.sol
@@ -7,6 +7,9 @@ import "../kernel/IKernel.sol";
 
 
 contract AppProxyBase is AppStorage, DepositableDelegateProxy, KernelConstants {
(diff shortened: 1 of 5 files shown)


izqui:
Solidity doesn't optimize revert reasons well, adding just one character of reason results in ~115 bytes added to the contract (23k gas). Because it pads string constants with 0s so they fit in a 32 byte word, having short strings as we do in this PR doesn't save any gas (contract bytecode gets charged at 200 gas per byte regardless of whether it is 0 or not).

After discussing it offline, I think we shouldn't add revert reasons to proxies and just have them everywhere else, so proxies remain cheap to deploy. Adding revert reasons has resulted in a 33% gas increase for deploying proxies (300k to 400k), which given how easy it would be to debug them (if everything running in the implementation code has reasons, if there is no error data returned then it must be an issue in the proxy which is easier to debug)

PR #441, opened 2018-10-16
Revert reasons

### LINKED ISSUE #438: Revert with reason

izqui:
Closes #438

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -48,14 +48,18 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
 
     uint256 internal constant ORACLE_CHECK_GAS = 30000;
 
+    string private constant AUTH_INIT_KERNEL_ERROR = "ACL_AUTH_INIT_KERNEL";
+    string private constant AUTH_NO_MANAGER_ERROR = "ACL_AUTH_NO_MANAGER";
+    string private constant EXISTENT_MANAGER_ERROR = "ACL_EXISTENT_MANAGER";
+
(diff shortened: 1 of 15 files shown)


izqui:
Final bytecode diff from adding revert reasons:

```
ACL.json                      425200 more gas      +2126              +359
APMRegistry.json              279800 more gas      +1399              +359
AragonApp.json                70800 more gas       +354               +379
Autopetrified.json            Same                 0                  +354
CallsScript.json              131600 more gas      +658               +354
ENSSubdomainRegistrar.json    336200 more gas      +1681              +359
EVMScriptRegistry.json        306800 more gas      +1534              +359
EVMScriptRegistryFactory.json Same                 0                  +2905
Kernel.json                   369600 more gas      +1848              +359
Repo.json                     247400 more gas      +1237              +359
SafeMath64Mock.json           161000 more gas      +805               0
SafeMath8Mock.json            161000 more gas      +805               0
TestACLInterpreter.json       425200 more gas      +2126              +359
TokenMock.json                71200 more gas       +356               0
```

PR #442, opened 2018-10-18
[Prototype] AragonApp: signed calls and multi call execution (aka meta-txs support)

izqui:
Super early prototype of how we could add a way to execute signed calls in a contract inheriting `AragonApp`. Signed calls allow the entity performing an action doing so without ever having to own any ETH to pay for transaction fees, by having a relayer submit their signed calls on-chain. There seems to be some momentum around the [EIP1077](https://eips.ethereum.org/EIPS/eip-1077) standard, which could be interesting to support, as a standard interface could help create a ecosystem of relayer services.

By putting this in the `AragonApp` contract, any Aragon app would automatically benefit from this functionality. There is native support for ACL checks, but apps willing to use this for custom logic (e.g. Voting) should use `AragonApp.sender()` instead of `msg.sender`.

It uses volatile storage, `AragonApp` will write to a storage slot who the actual sender of the call is before calling itself with that calldata. After the sub-call ends executing, that storage slot is emptied. This would be quite expensive to do right now, but the new [net gas metering](https://eips.ethereum.org/EIPS/eip-1283) in Constantinople this is quite cheap to do.

Got down this rabbit-hole after having a discussing with @bingen about whether it would make sense to add [`multicall`](https://github.com/cartoucheco/registrar/blob/master/contracts/OwnedRegistrar.sol#L73-L77) to the Staking app, and his proposal to add it directly to `AragonApp`.

The current implementation doesn't allow to perform multiple calls from sender as in `multicall`, but it could be implemented by accepting empty signatures as calls executed by their sender.

### DIFF
--- a/contracts/apps/AppStorage.sol
+++ b/contracts/apps/AppStorage.sol
@@ -18,6 +18,9 @@ contract AppStorage {
     // keccak256("aragonOS.appStorage.pinnedCode"), used by Proxy Pinned
     bytes32 internal constant PINNED_CODE_POSITION = 0xdee64df20d65e53d7f51cb6ab6d921a0a6a638a91e942e1d8d02df28e31c038e;
 
+    bytes32 internal constant VOLATILE_SENDER_POSITION = keccak256("aragonOS.appStorage.volatile.sender");
+    bytes32 internal constant USED_NONCE_POSITION_BASE = keccak256("aragonOS.appStorage.usedNonce");
+
     function kernel() public view returns (IKernel) {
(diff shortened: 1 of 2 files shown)


izqui:
This could use the same signature validator as Agent uses: https://github.com/aragon/aragon-apps/blob/master/apps/agent/contracts/SignatureValidator.sol

sohkai:
Closing; this will be approached separately with baseline meta transactions support in aragonOS (e.g. https://github.com/aragon/aragonOS/pull/526).

PR #443, opened 2018-10-22
Optimization: prune constants

sohkai:
Prunes most public constants, leaving only important constants that are likely to be useful on-chain:

- Roles
- Default app IDs (useful across versions of `aragonOS` in case they change)
- Kernel namespaces (useful across versions of the `Kernel` in case they change)
- ACL scripting constants (potentially useful across versions / verifying information from getters)

Commits:

- e204fa2: cosmetic reordering / comments
- 0a0983e: renames `APMRegistryConstants` to `APMInternalAppNames` as these APM "internal" apps are registered in the same way as other APM packages and the contract only holds the string names
- **cd19f32**: actually removes constants
- f9fe190: fixes `kernel.newAppInstance()` in tests due to the default version changing :(

Against #432, the bytecode comparison:

```
ACL.json                       33600 less gas       -168               0
APMNamehash.json               270200 less gas      -1351              -3
APMRegistry.json               189200 less gas      -946               0
APMRegistryFactory.json        75600 less gas       -378               0
APMRegistryFactoryMock.json    75600 less gas       -378               0
AppProxyFactory.json           73600 less gas       -368               0
AppProxyPinned.json            36800 less gas       -184               0
AppProxyPinnedStorageMock.json 36000 less gas       -180               -53
AppProxyUpgradeable.json       36800 less gas       -184               0
AragonApp.json                 34600 less gas       -173               0
DAOFactory.json                36000 less gas       -180               0
ENSConstants.json              186600 less gas      -933               -3
ENSFactory.json                152800 less gas      -764               0
ENSSubdomainRegistrar.json     170800 less gas      -854               0
EVMScriptRegistry.json         34600 less gas       -173               0
EVMScriptRegistryConstants.js… 26000 less gas       -130               -1
EVMScriptRegistryFactory.json  91600 less gas       -458               -173
EVMScriptRunner.json           27200 less gas       -136               0
EtherTokenConstant.json        24400 less gas       -122               -2
KeccakConstants.json           183000 more gas      +915               0
Kernel.json                    97200 less gas       -486               0
KernelConstants.json           55400 less gas       -277               -2
KernelOverloadMock.json        14800 more gas       +74                0
KernelPinnedStorageMock.json   97200 less gas       -486               0
KernelProxy.json               36000 less gas       -180               0
KernelStorage.json             15400 more gas       +77                0
Repo.json                      34600 less gas       -173               0
Uint256Helpers.json            18800 less gas       -94                -3
```

### DIFF
--- a/contracts/apm/APMRegistry.sol
+++ b/contracts/apm/APMRegistry.sol
@@ -8,16 +8,14 @@ import "../acl/ACL.sol";
 import "./Repo.sol";
 
 
-contract APMRegistryConstants {
-    // Cant have a regular APM appId because it is used to build APM
-    // TODO: recheck this
-    string public constant APM_APP_NAME = "apm-registry";
-    string public constant REPO_APP_NAME = "apm-repo";
-    string public constant ENS_SUB_APP_NAME = "apm-enssub";
+contract APMInternalAppNames {
+    string internal constant APM_APP_NAME = "apm-registry";
+    string internal constant REPO_APP_NAME = "apm-repo";
+    string internal constant ENS_SUB_APP_NAME = "apm-enssub";
 }
 
 
-contract APMRegistry is AragonApp, AppProxyFactory, APMRegistryConstants {
+contract APMRegistry is AragonApp, AppProxyFactory, APMInternalAppNames {
     /* Hardcoded constants to save gas
     bytes32 public constant CREATE_REPO_ROLE = keccak256("CREATE_REPO_ROLE");
     */
--- a/contracts/apm/APMRegistry.sol
+++ b/contracts/apm/APMRegistry.sol
@@ -18,14 +18,16 @@ contract APMRegistryConstants {
 
 
 contract APMRegistry is AragonApp, AppProxyFactory, APMRegistryConstants {
-    AbstractENS public ens;
-    ENSSubdomainRegistrar public registrar;
-
-    // bytes32 public constant CREATE_REPO_ROLE = keccak256("CREATE_REPO_ROLE");
+    /* Hardcoded constants to save gas
+    bytes32 public constant CREATE_REPO_ROLE = keccak256("CREATE_REPO_ROLE");
+    */
     bytes32 public constant CREATE_REPO_ROLE = 0x2a9494d64846c9fdbf0158785aa330d8bc9caf45af27fa0e8898eb4d55adcea6;
 
     string private constant ERROR_INIT_PERMISSIONS = "APMREG_INIT_PERMISSIONS";
     string private constant ERROR_EMPTY_NAME = "APMREG_EMPTY_NAME";
+
+    AbstractENS public ens;
+    ENSSubdomainRegistrar public registrar;
 
     event NewRepo(bytes32 id, string name, address repo);
 
--- a/contracts/common/EtherTokenConstant.sol
+++ b/contracts/common/EtherTokenConstant.sol
@@ -8,5 +8,5 @@ pragma solidity ^0.4.24;
 // aragonOS and aragon-apps rely on address(0) to denote native ETH, in
 // contracts where both tokens and ETH are accepted
 contract EtherTokenConstant {
-    address public constant ETH = address(0);
+    address internal constant ETH = address(0);
 }
--- a/contracts/ens/ENSConstants.sol
+++ b/contracts/ens/ENSConstants.sol
@@ -6,9 +6,9 @@ pragma solidity ^0.4.24;
 
 
 contract ENSConstants {
-    bytes32 public constant ENS_ROOT = bytes32(0);
-    bytes32 public constant ETH_TLD_LABEL = keccak256("eth");
-    bytes32 public constant ETH_TLD_NODE = keccak256(abi.encodePacked(ENS_ROOT, ETH_TLD_LABEL));
-    bytes32 public constant PUBLIC_RESOLVER_LABEL = keccak256("resolver");
-    bytes32 public constant PUBLIC_RESOLVER_NODE = keccak256(abi.encodePacked(ETH_TLD_NODE, PUBLIC_RESOLVER_LABEL));
+    bytes32 internal constant ENS_ROOT = bytes32(0);
+    bytes32 internal constant ETH_TLD_LABEL = keccak256("eth");
+    bytes32 internal constant ETH_TLD_NODE = keccak256(abi.encodePacked(ENS_ROOT, ETH_TLD_LABEL));
+    bytes32 internal constant PUBLIC_RESOLVER_LABEL = keccak256("resolver");
+    bytes32 internal constant PUBLIC_RESOLVER_NODE = keccak256(abi.encodePacked(ETH_TLD_NODE, PUBLIC_RESOLVER_LABEL));
 }
--- a/contracts/evmscript/IEVMScriptRegistry.sol
+++ b/contracts/evmscript/IEVMScriptRegistry.sol
@@ -9,9 +9,9 @@ import "./IEVMScriptExecutor.sol";
 
 contract EVMScriptRegistryConstants {
     /* Hardcoded constants to save gas
-    bytes32 public constant EVMSCRIPT_REGISTRY_APP_ID = apmNamehash("evmreg");
+    bytes32 internal constant EVMSCRIPT_REGISTRY_APP_ID = apmNamehash("evmreg");
     */
-    bytes32 public constant EVMSCRIPT_REGISTRY_APP_ID = 0xddbcfd564f642ab5627cf68b9b7d374fb4f8a36e941a75d89c87998cef03bd61;
+    bytes32 internal constant EVMSCRIPT_REGISTRY_APP_ID = 0xddbcfd564f642ab5627cf68b9b7d374fb4f8a36e941a75d89c87998cef03bd61;
 }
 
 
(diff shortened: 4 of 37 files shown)


contracts/apm/APMRegistry.sol
```diff
@@ -8,25 +8,25 @@ import "../acl/ACL.sol";
 import "./Repo.sol";
 
 
-contract APMRegistryConstants {
-    // Cant have a regular APM appId because it is used to build APM
```
sohkai: AFAIK these apps are all registered in [`APMRegistryFactory` as having normal `<name>.aragonpm.eth` names now](https://github.com/aragon/aragonOS/blob/dev/contracts/factory/APMRegistryFactory.sol#L102)

contracts/common/EtherTokenConstant.sol
```diff
@@ -8,5 +8,5 @@ pragma solidity ^0.4.24;
 // aragonOS and aragon-apps rely on address(0) to denote native ETH, in
 // contracts where both tokens and ETH are accepted
 contract EtherTokenConstant {
-    address public constant ETH = address(0);
+    address internal constant ETH = address(0);
```
sohkai: Note that this will require a small change in the tests for aragon-apps.

contracts/evmscript/IEVMScriptRegistry.sol
```diff
@@ -9,9 +9,9 @@ import "./IEVMScriptExecutor.sol";
 
 contract EVMScriptRegistryConstants {
     /* Hardcoded constants to save gas
-    bytes32 public constant EVMSCRIPT_REGISTRY_APP_ID = apmNamehash("evmreg");
+    bytes32 internal constant EVMSCRIPT_REGISTRY_APP_ID = apmNamehash("evmreg");
     */
-    bytes32 public constant EVMSCRIPT_REGISTRY_APP_ID = 0xddbcfd564f642ab5627cf68b9b7d374fb4f8a36e941a75d89c87998cef03bd61;
+    bytes32 internal constant EVMSCRIPT_REGISTRY_APP_ID = 0xddbcfd564f642ab5627cf68b9b7d374fb4f8a36e941a75d89c87998cef03bd61;
```
sohkai: Was debating exposing this as part of `EVMScriptRegistry`; I didn't want to leave it `public` here because `AragonApp` inherits from this contract.

contracts/kernel/Kernel.sol
```diff
@@ -12,16 +12,26 @@ import "../factory/AppProxyFactory.sol";
...
-    //bytes32 public constant APP_MANAGER_ROLE = keccak256("APP_MANAGER_ROLE");
-    //bytes32 public constant DEFAULT_VAULT_APP_ID = apmNamehash("vault");
+    /* Hardcoded constants to save gas
+    bytes32 public constant APP_MANAGER_ROLE = keccak256("APP_MANAGER_ROLE");
+    bytes32 public constant DEFAULT_ACL_APP_ID = apmNamehash("acl");
+    bytes32 internal constant DEFAULT_VAULT_APP_ID = apmNamehash("vault");
+    */
     bytes32 public constant APP_MANAGER_ROLE = 0xb6d92708f3d4817afc106147d969e229ced5c46e65e0a5002a0d391287762bd0;
-    bytes32 public constant DEFAULT_VAULT_APP_ID = 0x7e852e0fcfce6551c13800f1e7476f982525c2b5277ba14b24339c68416336d1;
+    bytes32 public constant DEFAULT_ACL_APP_ID = 0xe3262375f45a6e2026b7e7b18c2b807434f2508fe1a2a3dfb493c7df8f4aad6a;
```
sohkai: Moved the `ACL_APP_ID` to here, as it's really the _default_ ACL ID (and also isn't needed anywhere else).

contracts/kernel/Kernel.sol
```diff
@@ -12,16 +12,26 @@ import "../factory/AppProxyFactory.sol";
...
 
     string private constant ERROR_APP_NOT_CONTRACT = "KERNEL_APP_NOT_CONTRACT";
     string private constant ERROR_INVALID_APP_CHANGE = "KERNEL_INVALID_APP_CHANGE";
     string private constant ERROR_AUTH_FAILED = "KERNEL_AUTH_FAILED";
 
+    // External access to namespace constants to mimic default getters for constants
+    /* solium-disable function-order, mixedcase */
+    function CORE_NAMESPACE() external pure returns (bytes32) { return KERNEL_CORE_NAMESPACE; }
+    function APP_BASES_NAMESPACE() external pure returns (bytes32) { return KERNEL_APP_BASES_NAMESPACE; }
+    function APP_ADDR_NAMESPACE() external pure returns (bytes32) { return KERNEL_APP_ADDR_NAMESPACE; }
```
sohkai: A little hack to keep the namespaces exposed in the same was as before :).

contracts/ens/ENSConstants.sol
```diff
@@ -6,9 +6,9 @@ pragma solidity ^0.4.24;
...
-    bytes32 public constant ENS_ROOT = bytes32(0);
-    bytes32 public constant ETH_TLD_LABEL = keccak256("eth");
-    bytes32 public constant ETH_TLD_NODE = keccak256(abi.encodePacked(ENS_ROOT, ETH_TLD_LABEL));
-    bytes32 public constant PUBLIC_RESOLVER_LABEL = keccak256("resolver");
-    bytes32 public constant PUBLIC_RESOLVER_NODE = keccak256(abi.encodePacked(ETH_TLD_NODE, PUBLIC_RESOLVER_LABEL));
+    bytes32 internal constant ENS_ROOT = bytes32(0);
+    bytes32 internal constant ETH_TLD_LABEL = keccak256("eth");
+    bytes32 internal constant ETH_TLD_NODE = keccak256(abi.encodePacked(ENS_ROOT, ETH_TLD_LABEL));
+    bytes32 internal constant PUBLIC_RESOLVER_LABEL = keccak256("resolver");
+    bytes32 internal constant PUBLIC_RESOLVER_NODE = keccak256(abi.encodePacked(ETH_TLD_NODE, PUBLIC_RESOLVER_LABEL));
```
izqui: Probably a good moment to hardcode these as well.

contracts/evmscript/IEVMScriptRegistry.sol
izqui: Maybe we can make `getExecutorRegistry()` public in EVMScriptRunner?

contracts/kernel/KernelConstants.sol
```diff
@@ -5,19 +5,21 @@
...
+    bytes32 internal constant KERNEL_APP_BASES_NAMESPACE = keccak256("base");
+    bytes32 internal constant KERNEL_APP_ADDR_NAMESPACE = keccak256("app");
+    */
+    bytes32 internal constant KERNEL_CORE_NAMESPACE = 0xc681a85306374a5ab27f0bbc385296a54bcd314a1948b6cf61c4ea1bc44bb9f8;
+    bytes32 internal constant KERNEL_APP_BASES_NAMESPACE = 0xf1f3eb40f5bc1ad1344716ced8b8a0431d840b5783aea1fd01786bc26f35ac0f;
+    bytes32 internal constant KERNEL_APP_ADDR_NAMESPACE = 0xd6f028ca0e8edb4a8c9757ca4fdccab25fa1e0317da1188108f7d2dee14902fb;
+}
 
+
+contract KernelConstants {
```
bingen: Do we need to keep this just for 1 constant? That constant is only used in `KernelProxy`, right?

sohkai: I've now moved all the kernel's default app ids into `KernelDefaultAppIds`.

sohkai:
@izqui @bingen would appreciate a look at fd0c69e.

My thinking around the grouping of these constants is to make it easier for a contract to get access to them in case it needed to interact with a kernel on-chain. I've also removed the constants from `KernelStorage` to make their usage more explicit.

bingen:
Hm, I wonder why would a contract want to know what's the `KERNEL_APP_ID` taking into account that it can just get `KernelProxy.implementation()`. I just can't think up of a use case now, but of course that doesn't mean it doesn't exist ;-)
About the names, I think I liked better `KERNEL_APP_ID` than `KERNEL_CORE_APP_ID` and `DEFAULT_XXX_APP_ID` than `KERNEL_DEFAULT_XXX_APP_ID` (it looks unnecessarily long to me) but it's fine.

sohkai:
> Hm, I wonder why would a contract want to know what's the KERNEL_APP_ID taking into account that it can just get KernelProxy.implementation()

Mainly if it wanted to use `setApp()` to update the Kernel.

> About the names, I think I liked better `KERNEL_APP_ID` than `KERNEL_CORE_APP_ID` and `DEFAULT_XXX_APP_ID` than `KERNEL_DEFAULT_XXX_APP_ID` (it looks unnecessarily long to me) but it's fine.

Totally agree, but we have to use different names to keep the public interface as `KERNEL_APP_ID`, `DEFAULT_XXX_APP_ID` :(.

bingen:
> Mainly if it wanted to use setApp() to update the Kernel. I could also see us possibly using this as a version identifier in the future.

I wonder what I was thinking about, haha. :man_facepalming: 
Anyway, this made me think that it would be nice to have a helper for that in `KernelProxy`, so there's no need to know and use `KERNEL_CORE_NAMESPACE` and `KERNEL_CORE_APP_ID`.

sohkai:
Yeah... the problem would then be having to duplicate access control in the proxy itself

PR #448, opened 2018-10-25
AragonApp: fix possibly uninitialized bytes array when checking permissions

sohkai:
From https://github.com/ConsenSys/aragonOS-audit-repo/issues/46:

> The `how` object in the `AragonApp.canPerform` function depends on possibly undefined Solidity behavior. Due to the sensitive nature of the function, we recommend initializing the `how` variable when `_params.length == 0` to avoid the possibility of a vulnerability.

------------------

I tried removing all the logic related to casting to/from bytes and uint arrays to their own separate contract or libraries, but in both cases the bytecode diff was extreme (+50k to +120k in gas) so I've opted not to consolidate and just align all the comments.

Bytecode comparison:

```
ACL.json                       1600 less gas        -8                 0
APMRegistry.json               1800 less gas        -9                 0
AppStub.json                   1600 less gas        -8                 0
AppStub2.json                  1600 less gas        -8                 0
AppStubConditionalRecovery.js… 1600 less gas        -8                 0
AppStubDepositable.json        1600 less gas        -8                 0
AragonApp.json                 1600 less gas        -8                 0
ENSSubdomainRegistrar.json     1600 less gas        -8                 0
EVMScriptRegistry.json         1600 less gas        -8                 0
EVMScriptRegistryFactory.json  Same                 0                  -8
MockScriptExecutorApp.json     1600 less gas        -8                 0
Repo.json                      1600 less gas        -8                 0
TestACLInterpreter.json        1600 less gas        -8                 0
UnsafeAppStub.json             1600 less gas        -8                 0
UnsafeAppStubDepositable.json  1600 less gas        -8                 0
UnsafeAragonApp.json           1600 less gas        -8                 0
UnsafeAragonAppMock.json       1600 less gas        -8                 0
VaultMock.json                 1600 less gas        -8                 0
```

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -244,7 +244,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     function hasPermission(address _who, address _where, bytes32 _what, bytes memory _how) public view returns (bool) {
         // Force cast the bytes array into a uint256[], by overwriting its length
         // Note that the uint256[] doesn't need to be initialized as we immediately overwrite it
-        // with _how and a new length, and _howbecomes invalid from this point forward
+        // with _how and a new length, and _how becomes invalid from this point forward
         uint256[] memory how;
         uint256 intsLength = _how.length / 32;
         assembly {
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -242,13 +242,16 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     * @return boolean indicating whether the ACL allows the role or not
     */
     function hasPermission(address _who, address _where, bytes32 _what, bytes memory _how) public view returns (bool) {
+        // Force cast the bytes array into a uint256[], by overwriting its length
+        // Note that the uint256[] doesn't need to be initialized as we immediately overwrite it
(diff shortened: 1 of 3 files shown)


contracts/acl/ACL.sol
```diff
@@ -242,13 +242,16 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     * @return boolean indicating whether the ACL allows the role or not
     */
     function hasPermission(address _who, address _where, bytes32 _what, bytes memory _how) public view returns (bool) {
+        // Force cast the bytes array into a uint256[], by overwriting its length
+        // Note that the uint256[] doesn't need to be initialized as we immediately overwrite it
+        // with _how and a new length, and _howbecomes invalid from this point forward
```
izqui:
```suggestion
        // with _how and a new length, and _how becomes invalid from this point forward
```

sohkai:
I'm not sure that'd be an optimization; the casting is pretty efficient (it literally just uses the input array's memory slot and overwrites its length).

izqui:
We could potentially save the 1 `MSTORE` that sets the length to 0 when it already is, but depending on how solidity allocates memory when doing `new bytes(0)` it could actually be more expensive. So I'd say let's keep it as it is right now.

Issue #368, opened 2018-07-24
Move APM into its own repo


sohkai:
Given `APM`'s current implementation, as its own DAO built on top of aOS, it doesn't make a lot of sense for it to be bundled together with aOS. Auditors have also commented on it being a source of confusion as to how the system works and is deployed. For all intents and purposes, `APM` is already logically decoupled from aOS.

This also lowers the general complexity of aOS, in its contracts, tests, and deployment strategy.

A few nice things to do when migrating:

- [x] [Remove the dynamically calculated `keccak256()` constants and replace them with their hashes](https://github.com/aragon/aragonOS/pull/363#issuecomment-407415022)
- [x] ~Investigate the APM-native apps' app IDs (e.g. `apm-repo` and `apm-enssub`)~ (moved to apm repo: https://github.com/aragon/apm/issues/2)
- [x] ~Investigate whether `APMRegistryFactoryMock` is necessary, and if so, fix the app IDs in the test mock (moved to apm repo: https://github.com/aragon/apm/issues/3)

PR #450, opened 2018-10-26
Remove warnings

bingen:


sohkai:
Going to close this for now, since most of the warnings are from ENS and will be removed when #368 is done.

PR #453, opened 2018-10-29
Inject web3 to deploy-apm script

bingen:


scripts/deploy-apm.js
```diff
@@ -7,6 +7,7 @@ const logDeploy = require('./helpers/deploy-logger')
 const getAccounts = require('./helpers/get-accounts')
 
 const globalArtifacts = this.artifacts // Not injected unless called directly via truffle
+const globalWeb3 = this.artifacts // Not injected unless called directly via truffle
```
izqui:
```suggestion
const globalWeb3 = this.web3 // Not injected unless called directly via truffle
```

scripts/deploy-test-ens.js
```diff
@@ -2,10 +2,19 @@ const logDeploy = require('./helpers/deploy-logger')
 const getAccounts = require('./helpers/get-accounts')
 
 const globalArtifacts = this.artifacts // Not injected unless called directly via truffle
+const globalWeb3 = this.artifacts // Not injected unless called directly via truffle
```
izqui:
```suggestion
const globalWeb3 = this.web3 // Not injected unless called directly via truffle
```

scripts/deploy-apm.js
bingen: :man_facepalming:

PR #456, opened 2018-11-09
encodeParam and encodeParams functions

### LINKED ISSUE #427: Add a function to convert an ACL Param to uint256
Currently, ACL parameters are created with the `Param` struct:
```solidity
struct Param {
    uint8 id;
    uint8 op;
    uint240 value;
}
```
but the ACL `grantPermissionP` function takes a `uint256[]` for its parameters:
```solidity
function grantPermissionP(address _entity, address _app, bytes32 _role, uint256[] _params)
```

It could be interesting to have a helper function to convert a `Param` struct to `uint256`. 
Something like this:

```solidity
function encodeParam(uint8 id, uint8 op, uint240 value) internal pure returns (uint256) {
    return uint256(id) << 248 | uint256(op) << 240 | value;
}
```

leftab:
Fixes #427 

Adds the following functions:
- `encodeParam(Param param) internal pure returns (uint256)`
- `encodeParams(Param[] params) internal pure returns (uint256[])`

I didn't add the `encodeParam(uint8 id, uint8 op, uint240 value)` function since I think `encodeParam(Param param)` covers pretty much the same use cases and is easier to use.

TODO:

- [x] Sanity check gas usage (these changes shouldn't cause any gas increases)


contracts/acl/ACLSyntaxSugar.sol
```diff
@@ -86,6 +86,31 @@ contract ACLSyntaxSugar {
 
 
 contract ACLHelpers {
+
+    enum Op { NONE, EQ, NEQ, GT, LT, GTE, LTE, RET, NOT, AND, OR, XOR, IF_ELSE } // op types
```
sohkai: What do you think about putting this in another contract, e.g. `ACLParams`?

contracts/acl/ACLSyntaxSugar.sol
```diff
@@ -86,6 +86,31 @@ contract ACLSyntaxSugar {
...
+        for (uint i = 0; i < params.length; i++) {
+            encodedParams[i] = encodeParam(params[i]);
+        }
+
+        return encodedParams;
+    }
+
+    function encodeParam(Param param) internal pure returns (uint256) {
+        return uint256(param.id) << 248 | uint256(param.op) << 240 | param.value;
+    }
```
sohkai: We could also expose the helpers in [`ACLHelpers`](https://github.com/aragon/aragonOS/blob/dev/contracts/test/helpers/ACLHelper.sol#L7).

contracts/test/TestACLHelpers.sol
```diff
@@ -0,0 +1,56 @@
...
+        Assert.equal(uint256(params[0].op), uint256(op0), "Encoded op is not equal");
+        Assert.equal(uint256(params[0].value), uint256(value0), "Encoded value is not equal");
+
+        (uint32 id1, uint32 op1, uint32 value1) = decodeParamsList(encodedParam[1]);
+
+        Assert.equal(uint256(params[1].id), uint256(id1), "Encoded id is not equal");
+        Assert.equal(uint256(params[1].op), uint256(op1), "Encoded op is not equal");
+        Assert.equal(uint256(params[1].value), uint256(value1), "Encoded value is not equal");        
+    }    
+
```
sohkai: Let's test `encodeOperator` and `encodeIfElse` as well :)

sohkai (commented):
Looking really good! Just thinking we could expose this as its own contract separate from `ACLSyntaxSugar`.

contracts/acl/ACLSyntaxSugar.sol
leftab: Good idea! I agree they fit more inside `ACLParams`

contracts/acl/ACLSyntaxSugar.sol
leftab: I exposed all the functions except those having `struct` params since they would require the experimental ABI encoder.

contracts/test/TestACLHelpers.sol
leftab: Done :)

leftab:
Thanks for the review @sohkai ! I applied the changes but somehow the Travis tests seems to fail now. I'm a bit perplexed since everything looks fine when I run the tests locally.

contracts/test/TestACLHelpers.sol
leftab: I noticed that the `encodeOperator`  and `encodeIfElse` functions are in a separate file in the test folder. Do you think it would be a good idea if I move them to the `ACLHelpers` contract?

contracts/acl/ACLSyntaxSugar.sol
```diff
@@ -85,18 +87,34 @@ contract ACLSyntaxSugar {
 }
 
 
-contract ACLHelpers {
-    function decodeParamOp(uint256 _x) internal pure returns (uint8 b) {
+contract ACLHelpers is ACLParams {
+
```
bingen: Why are we adding 2 newlines here? Not a big deal, but we don't usually do that.

contracts/acl/ACLSyntaxSugar.sol
```diff
@@ -85,18 +87,34 @@ contract ACLSyntaxSugar {
...
+
+    function decodeParamOp(uint256 _x) public pure returns (uint8 b) {
```
bingen: Why are we converting these 3 functions from `internal` to `public`?

contracts/test/TestACLHelpers.sol
```diff
@@ -0,0 +1,60 @@
...
+        (uint32 id, uint32 op, uint32 value) = decodeParamsList(encodedParam);
+
+        Assert.equal(uint256(param.id), uint256(id), "Encoded id is not equal");
+        Assert.equal(uint256(param.op), uint256(op), "Encoded op is not equal");
+        Assert.equal(uint256(param.value), uint256(value), "Encoded value is not equal");
+    }
+
+    function testEncodeParams() public {
+        Param[] memory params = new Param[](6);
```
bingen: Why 6? Only 4 are being used, right? (not a big deal)

contracts/test/TestACLHelpers.sol
```diff
@@ -0,0 +1,60 @@
...
+
+        params[0] = Param(LOGIC_OP_PARAM_ID, uint8(Op.IF_ELSE), encodeIfElse(1, 2, 3));
+        params[1] = Param(LOGIC_OP_PARAM_ID, uint8(Op.AND), encodeOperator(2, 3));   
+        params[2] = Param(2, uint8(Op.EQ), 1);   
+        params[3] = Param(3, uint8(Op.NEQ), 2);                                   
+
+
+        uint256[] memory encodedParam = encodeParams(params);
+
+        (uint32 id0, uint32 op0, uint32 value0) = decodeParamsList(encodedParam[0]);
```
bingen: Maybe we could use a loop here.

contracts/test/TestACLHelpers.sol
bingen: Yes, I think it's a good idea, but let's wait for @sohkai opinion, maybe there is a reason I don't know for them to be there.

sohkai: Ah I didn't notice they were only in the tests.

I think moving them could be helpful; AFAIK they're not simply because we only used them in test functionality. We should either add comments to explain how they work or make `encodeOperator()` more explicit (it's a bit confusing at first why it builds off of `encodeIfElse()`).

leftab:
@sohkai @bingen Sorry guys for the delay, we've been quite busy working on our demo for Aracon and the end of the Drive app milestone. I should be able to post an update this week.

contracts/test/TestACLHelpers.sol
leftab: True! :)

contracts/test/TestACLHelpers.sol
leftab: Totally!

contracts/acl/ACLSyntaxSugar.sol
leftab: So it can be used by composition instead of only by inheritance. Do you think it would impact gaz usage and we should keep them `internal`?

leftab:
Currently looking into why one test is failing.

contracts/acl/ACLSyntaxSugar.sol
sohkai: Most likely wouldn't impact gas by very much (more or less only on deployment), but wondering how useful this would be. Do you think outside contracts might want access to this functionality, which is only accessible on the ACL?

We should also make sure `encodeParams()` and `encodeParam()` are the same visibility as their decode counterparts.

leftab: Sorry for the delay. hmm.. it’s been so long that I don’t even remember haha! I guess I wanted the `encodeParam` and `encodeParams` functions public, but it’s not supported yet anyway. I converted them back to `internal` and exposed the `encodeOperator` and `encodeIfElse` functions. The two functions are used in the ACL Interpreter tests and I made `encodeOperator` a little bit more explicit.

leftab: What do you think if I move the `ACLHelpers` contract into its own `ACLHelpers.sol` file? It would be a breaking change however.

sohkai: Let's mark it as something we should do in the future 👍 .

leftab:
As for the gas usage sanity check, since the new functions are not used in aragonOS and are `internal`, there should be no impact on gas usage.

sohkai:
@leftab Will check the compiled bytecode of the `ACL`. I don't see why it would change with this change, but need to be sure (don't want it to accidentally change from our initial deployment).

contracts/test/helpers/ACLHelper.sol
```diff
@@ -3,17 +3,6 @@ pragma solidity 0.4.24;
...
-        return encodeIfElse(param1, param2, 0);
-    }
-
-    function encodeIfElse(uint256 condition, uint256 successParam, uint256 failureParam) internal pure returns (uint240) {
-        return uint240(condition + (successParam << 32) + (failureParam << 64));
-    }
-}
-
-
 contract AcceptOracle is IACLOracle {
```
sohkai: Could we rename this file, since it only contains test `ACLOracle`s now?

contracts/acl/ACLParams.sol
```diff
@@ -0,0 +1,15 @@
+pragma solidity ^0.4.24;
```
sohkai: Could we add the MIT header here (e.g. see https://github.com/aragon/aragonOS/blob/dev/contracts/acl/IACL.sol#L1)?

contracts/test/helpers/ACLHelper.sol
leftab: Renamed it to ACLOracleHelper

contracts/acl/ACLParams.sol
leftab: Done! ;) Is there a rule of thumb on which file an MIT header should be added?

sohkai: I need to write more documentation on this, but on all the unpinned contracts (since they're the only ones meant to be used by apps, which frees them up to use any licensing model they want).

sohkai:
@leftab I've pushed a commit to split `ACLHelpers` from `ACLSyntaxSugar` and clean up some formatting.

Unfortunately though, it looks like the changes **_do_** cause a bytecode difference against the old contracts :(. Going between this PR and `dev`, there's a bit of difference at the very end of the deployed bytecode:

screen shot 2019-03-07 at 7 51 25 pm

Will have to investigate further to see what's exactly causing this, but it does mean it's a bigger change than expected.

sohkai:
@leftab Because this wasn't pulled into some of the minor aragonOS 4.x releases, due to the incompatible bytecode (and because we didn't want to risk re-deploying the ACL for 0.7), we'll roll this into aragonOS 5 :).

PR #459, opened 2018-11-15
revokeAll function

### LINKED ISSUE #333: Revoke all granted permissions for a role

juslar:
Fixes #333 

Implements `revokeAll(address _app, bytes32 _role)`

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -54,6 +54,9 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     // Who is the manager of a permission
     mapping (bytes32 => address) internal permissionManager;
 
+    // Eras for the different roles
+    mapping (bytes32 => uint256) internal roleEras;
+
     event SetPermission(address indexed entity, address indexed app, bytes32 indexed role, bool allowed);
     event SetPermissionParams(address indexed entity, address indexed app, bytes32 indexed role, bytes32 paramsHash);
     event ChangePermissionManager(address indexed app, bytes32 indexed role, address indexed manager);
@@ -144,6 +147,19 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     {
         _setPermission(_entity, _app, _role, NO_PERMISSION);
     }
+
+    /**
+    * @dev Revokes all permissions for a specified role if allowed. This requires `msg.sender` to be the the permission manager
+    * @notice Revoke from all the ability to perform actions requiring `_role` on `_app`
+    * @param _app Address of the app in which the role will be revoked
+    * @param _role Identifier for the group of actions in app being revoked
+    */
+    function revokeAll(address _app, bytes32 _role)
+        external
+        onlyPermissionManager(_app, _role)
+    {
+        roleEras[roleHash(_app, _role)] = roleEras[roleHash(_app, _role)] + 1;
+    }     
 
     /**
     * @notice Set `_newManager` as the manager of `_role` in `_app`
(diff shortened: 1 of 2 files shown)


juslar:
I thought about a potential breaking change with the implementation I submitted. If a DAO currently in production upgrades to this version of the ACL, all the permissions would be invalid. Not too sure how to address this. An easy fix would be to add something like:

```javascript
function permissionHash(address _who, address _where, bytes32 _what) internal view returns (bytes32) {
	uint256 roleEra = roleEras[roleHash(_where, _what)];
	
	if (roleEra == 0)
		return keccak256(abi.encodePacked("PERMISSION", _who, _where, _what));
	
	return keccak256(abi.encodePacked("PERMISSION", roleEra, _who, _where, _what));
}
```

But that would consume a bit more gas. What do you think?

sohkai:
@juslar Nice! I think the backwards-compatible version isn't a terrible cost to swallow, and we should make sure to comment it.

An interesting thing we might want to set up in the future is upgradability tests between compatible versions.

juslar:
Thanks! I added the backwards-compatible line.

Regarding the coverage decreasing, it's due to this line:

`require(newRoleEra >= roleEras[roleHash(_app, _role)], ERROR_ROLE_ERA_INCREMENT);`

Do you have an idea how I could cover this? Is it ok to leave it this way?

contracts/acl/ACL.sol
```diff
@@ -145,6 +149,21 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
+    * @dev Revokes all permissions for a specified role if allowed. This requires `msg.sender` to be the the permission manager
+    * @notice Revoke from all the ability to perform actions requiring `_role` on `_app`
+    * @param _app Address of the app in which the role will be revoked
+    * @param _role Identifier for the group of actions in app being revoked
+    */
+    function revokeAll(address _app, bytes32 _role)
+        external
+        onlyPermissionManager(_app, _role)
+    {
+        uint256 newRoleEra = roleEras[roleHash(_app, _role)] + 1;
```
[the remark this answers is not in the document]
bingen: Sure, although it seems really unlikely that this function gets called `2^256` times for the same app and role, I guess you mean instead of the next line, right?

contracts/acl/ACL.sol
```diff
@@ -145,6 +149,21 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
+        require(newRoleEra >= roleEras[roleHash(_app, _role)], ERROR_ROLE_ERA_INCREMENT);
+        roleEras[roleHash(_app, _role)] = newRoleEra;
```
bingen: I wonder if using a pointer to `roleEras[roleHash(_app, _role)]` could save some gas here.

juslar:
Thank you @sohkai. I made the changes :)

contracts/acl/ACL.sol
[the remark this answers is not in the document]
juslar: I added SafeMath and will remove the require on the next line :)

contracts/acl/ACL.sol
juslar: Do you mean using a storage pointer? I don't think I can on a uint256 value. Maybe I'm missing something though.

bingen: Oh, yes, you are right, sorry for the confusion. We could still save 2 calls to `roleHash` function, but as it's going to be one less if you remove the `require`, I don't know if that's a big gain.

sohkai: Probably still a decent net gain, since there also wouldn't be another `SLOAD`.

juslar: Good :) Pushed the optimization.

contracts/acl/ACL.sol
```diff
@@ -145,6 +150,20 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
+    * @notice Revoke from all the ability to perform actions requiring `_role` on `_app`
+    * @param _app Address of the app in which the role will be revoked
+    * @param _role Identifier for the group of actions in app being revoked
+    */
+    function revokeAll(address _app, bytes32 _role)
+        external
+        onlyPermissionManager(_app, _role)
+    {
+        bytes32 hash = roleHash(_app, _role);
+        roleEras[hash] = roleEras[hash].add(1);
```
sohkai: Just realized, we should definitely emit an event here as otherwise it'd be very difficult to know this happened in a frontend.

juslar: Sorry for the delay! I added a `RevokeAllPermissions` event :)

sohkai:
@juslar Because this wasn't pulled into some of the minor aragonOS 4.x releases (and because we didn't want to risk re-deploying the ACL for 0.7), we'll roll this into aragonOS 5 :).

PR #463, opened 2018-11-23
update license

Smokyish:


LICENSE
```diff
@@ -3,7 +3,7 @@
                     GNU GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
 
- Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Copyright (C) Aragon Association
```
sohkai: Should we use aragon.org as the URL?

package.json
```diff
@@ -26,7 +26,7 @@
     "/test",
     "/truffle-config.js"
   ],
-  "author": "Aragon Institution MTU <contact@aragon.one>",
+  "author": "Aragon Association <contact@aragon.one>",
```
sohkai: Should we change this to `contact@aragon.org` or another aragon.org email?

luisivan: Indeed

LICENSE
luisivan: +1

PR #466, opened 2018-11-27
KernelProxy: emit SetApp event on construction

### LINKED ISSUE #457: Ensure KernelProxy integrity
As mentioned in https://github.com/aragon/aragon/issues/462 (under Bonus > History integrity), there is currently no way to detect if a KernelProxy was ever pointed to a malicious Kernel `implementation` (other that checking debug traces for all transactions in Ethereum for calls to the KernelProxy 🙃) that may have installed a backdoor in the storage of some contract.

In order to detect this, we cannot rely on saving any storage in the KernelProxy contract as a malicious Kernel implementation could clean up this storage after it is done adding the backdoor and removes itself as the implementation.

The only possible way to mitigate this is by having the KernelProxy emit a log with its implementation address every time that it forwards a call to its implementation. These logs cannot be tampered with by any malicious Kernel implementation, even if the first implementation is malicious (provided that the KernelProxy code is checked and it is not malicious itself).

Alternatively, being able to track down the creation of a KernelProxy to a trusted DAOFactory address would provide the same guarantee, as if the kernel is ever upgraded to a malicious Kernel, it will emit an untamperable log.

sohkai:
Fixes #457 by allowing a frontend to follow all `SetApp()` events emitted from the start of a KernelProxy's lifespan.

### DIFF
--- a/test/helpers/decodeEvent.js
+++ b/test/helpers/decodeEvent.js
@@ -1,0 +1,13 @@
+const abi = require('web3-eth-abi')
+
+module.exports = {
+    decodeEventsOfType: (receipt, eventAbi) => {
+        const eventSignature = abi.encodeEventSignature(eventAbi)
+        const eventLogs = receipt.logs.filter(l => l.topics[0] === eventSignature)
+        return eventLogs.map(log => {
+            log.event = abi.name
+            log.args = abi.decodeLog(eventAbi.inputs, log.data, log.topics.slice(1))
+            return log
+        })
+    }
+}
--- a/test/helpers/decodeEvent.js
+++ b/test/helpers/decodeEvent.js
@@ -1,7 +1,8 @@
 const abi = require('web3-eth-abi')
 
 module.exports = {
-    decodeEventsOfType: (receipt, eventAbi) => {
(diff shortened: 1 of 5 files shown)


izqui:
We should merge rhis

test/helpers/decodeEvent.js
```diff
@@ -0,0 +1,13 @@
+const abi = require('web3-eth-abi')
+
+module.exports = {
+    decodeEventsOfType: (receipt, eventAbi) => {
```
bingen: Wow, this is really nice! Way better than what I was doing [here](https://github.com/aragon/dao-kits/blob/master/kits/multisig/test/multisig.js#L22). I wonder if we could put it under `@aragon/test-helpers` or some other place which makes it easy for reuse.
I also think it may be nice to add a wrapper with `abi.filter(abi => abi.name === eventName && abi.type === 'event')[0]`, so just passing the receipt, the global abi and the event name is enough.

sohkai: Yes... we should migrate to `@aragon/test-helpers` :D.

Will change the API to take in the abi too.

PR #469, opened 2018-12-17
feat: Add SafeERC20

sohkai:
Required for https://github.com/aragon/aragon-apps/issues/562.

Adds a `SafeERC20` library that should be used to enhance `ERC20` types. The `safeTransfer()`, `safeTransferFrom()`, and `safeApprove()` methods ensure standards-compliance (e.g. always returning a value) with different token interfaces in the wild.

Note that the change for `VaultRecoverable.sol` means every `AragonApp` is affected.

Bytecode comparison:

```
ACL.json                       49200 more gas       +246               0
APMRegistry.json               50600 more gas       +253               0
AppStub.json                   49000 more gas       +245               0
AppStub2.json                  49200 more gas       +246               0
AppStubConditionalRecovery.js… 49200 more gas       +246               0
AppStubDepositable.json        49200 more gas       +246               0
AragonApp.json                 49200 more gas       +246               0
ENSSubdomainRegistrar.json     47600 more gas       +238               0
EVMScriptRegistry.json         47600 more gas       +238               0
EVMScriptRegistryFactory.json  Same                 0                  +238
Kernel.json                    48600 more gas       +243               0
KernelConstantsMock.json       48600 more gas       +243               0
KernelDepositableMock.json     48600 more gas       +243               0
KernelPinnedStorageMock.json   48600 more gas       +243               0
KernelSetAppMock.json          48600 more gas       +243               0
MockScriptExecutorApp.json     47800 more gas       +239               0
Repo.json                      49000 more gas       +245               0
TestACLInterpreter.json        49200 more gas       +246               0
TokenMock.json                 202400 more gas      +1012              +13
UnsafeAppStub.json             49000 more gas       +245               0
UnsafeAppStubDepositable.json  49200 more gas       +246               0
UnsafeAragonApp.json           49200 more gas       +246               0
UnsafeAragonAppMock.json       49200 more gas       +246               0
UnsafeRepo.json                49000 more gas       +245               0
UpgradedKernel.json            48600 more gas       +243               0
VaultMock.json                 49200 more gas       +246               0

Warning: could not compare some contracts that might have been renamed or removed
  Missing from /Users/brett/Development/aragon/aragonOS/bytecode: GeneralERC20.json, SafeERC20.json, SafeERC20Mock.json, TokenReturnFalseMock.json, TokenReturnMissingMock.json
```

### DIFF
--- a/contracts/common/SafeERC20.sol
+++ b/contracts/common/SafeERC20.sol
@@ -92,6 +92,10 @@ library SafeERC20 {
         return checkSuccess();
     }
 
+    /**
+    * @dev Static call into ERC20.balanceOf().
+    * Reverts if the call fails for some reason (should never fail).
+    */
     function staticBalanceOf(ERC20 _token, address _owner) internal view returns (uint256) {
         bytes memory balanceOfCallData = abi.encodeWithSelector(
             ERC20(_token).balanceOf.selector,
@@ -104,6 +108,10 @@ library SafeERC20 {
         return tokenBalance;
     }
 
+    /**
+    * @dev Static call into ERC20.allowance().
+    * Reverts if the call fails for some reason (should never fail).
+    */
     function staticAllowance(ERC20 _token, address _owner, address _spender) internal view returns (uint256) {
         bytes memory allowanceCallData = abi.encodeWithSelector(
             ERC20(_token).allowance.selector,
--- a/contracts/common/SafeERC20.sol
+++ b/contracts/common/SafeERC20.sol
@@ -122,7 +122,7 @@ library SafeERC20 {
     */
     function staticBalanceOf(ERC20 _token, address _owner) internal view returns (uint256) {
         bytes memory balanceOfCallData = abi.encodeWithSelector(
-            ERC20(_token).balanceOf.selector,
+            _token.balanceOf.selector,
             _owner
         );
 
@@ -138,7 +138,7 @@ library SafeERC20 {
     */
     function staticAllowance(ERC20 _token, address _owner, address _spender) internal view returns (uint256) {
         bytes memory allowanceCallData = abi.encodeWithSelector(
-            ERC20(_token).allowance.selector,
+            _token.allowance.selector,
             _owner,
             _spender
         );
--- a/contracts/common/SafeERC20.sol
+++ b/contracts/common/SafeERC20.sol
@@ -15,6 +15,9 @@ interface GeneralERC20 {
 
 
 library SafeERC20 {
+    string private constant ERROR_TOKEN_BALANCE_REVERTED = "SAFE_ERC_20_BALANCE_REVERTED";
+    string private constant ERROR_TOKEN_ALLOWANCE_REVERTED = "SAFE_ERC_20_ALLOWANCE_REVERTED";
+
     function checkSuccess() private pure returns (bool ret) {
         assembly {
             // Check number of bytes returned from last function call
@@ -36,6 +39,29 @@ library SafeERC20 {
 
             // Not sure what was returned: don't mark as success
             default { }
+        }
+    }
+
+    function staticInvoke(address _addr, bytes memory _calldata)
+        private
+        view
+        returns (bool success, uint256 ret)
+    {
+        assembly {
+            success := staticcall(
+                gas,                  // forward all gas
+                _addr,                // address
+                add(_calldata, 0x20), // calldata start
+                mload(_calldata),     // calldata length
+                0,                    // write output over scratch
+                0x20                  // uint256 return
+            )
+
+            switch success
+            case 1 {
+                ret := mload(0)
+            }
+            default {}
         }
     }
 
@@ -64,5 +90,30 @@ library SafeERC20 {
     function safeApprove(ERC20 _token, address _spender, uint256 _amount) internal returns (bool) {
         GeneralERC20(_token).approve(_spender, _amount);
         return checkSuccess();
+    }
+
+    function staticBalanceOf(ERC20 _token, address _owner) internal view returns (uint256) {
+        bytes memory balanceOfCallData = abi.encodeWithSelector(
+            ERC20(_token).balanceOf.selector,
+            _owner
+        );
+
+        (bool success, uint256 tokenBalance) = staticInvoke(_token, balanceOfCallData);
+        require(success, ERROR_TOKEN_BALANCE_REVERTED);
+
+        return tokenBalance;
+    }
+
+    function staticAllowance(ERC20 _token, address _owner, address _spender) internal view returns (uint256) {
+        bytes memory allowanceCallData = abi.encodeWithSelector(
+            ERC20(_token).allowance.selector,
+            _owner,
+            _spender
+        );
+
+        (bool success, uint256 allowance) = staticInvoke(_token, allowanceCallData);
+        require(success, ERROR_TOKEN_ALLOWANCE_REVERTED);
+
+        return allowance;
     }
 }
--- a/contracts/common/SafeERC20.sol
+++ b/contracts/common/SafeERC20.sol
@@ -5,16 +5,12 @@ pragma solidity ^0.4.24;
 
 import "../lib/token/ERC20.sol";
 
-// Same as ERC20 except the return values of these interfaces have been removed to allow us to
-// manually check them
-interface GeneralERC20 {
-    function transfer(address to, uint256 value) external;
-    function transferFrom(address from, address to, uint256 value) external;
-    function approve(address spender, uint256 value) external;
-}
-
 
 library SafeERC20 {
+    // Before 0.5, solidity has a mismatch between `address.transfer()` and `token.transfer()`:
+    // https://github.com/ethereum/solidity/issues/3544
+    bytes4 private constant TRANSFER_SELECTOR = 0xa9059cbb;
+
     string private constant ERROR_TOKEN_BALANCE_REVERTED = "SAFE_ERC_20_BALANCE_REVERTED";
     string private constant ERROR_TOKEN_ALLOWANCE_REVERTED = "SAFE_ERC_20_ALLOWANCE_REVERTED";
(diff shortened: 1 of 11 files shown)


contracts/common/VaultRecoverable.sol
```diff
@@ -27,8 +31,9 @@ contract VaultRecoverable is IVaultRecoverable, EtherTokenConstant, IsContract {
         if (_token == ETH) {
             vault.transfer(address(this).balance);
         } else {
-            uint256 amount = ERC20(_token).balanceOf(this);
-            ERC20(_token).transfer(vault, amount);
+            ERC20 token = ERC20(_token);
+            uint256 amount = token.balanceOf(this);
+            require(ERC20(_token).safeTransfer(vault, amount), ERROR_TOKEN_TRANSFER_FAILED);
```
sohkai: Not sure if we should add this change at this moment—making sure the token transfer was successful is the _right_ thing to do, but it's really only a problem for non-transferable tokens.

contracts/common/SafeERC20.sol
```diff
@@ -0,0 +1,68 @@
+// Inspired by AdEx (https://github.com/AdExNetwork/adex-protocol-eth/blob/b9df617829661a7518ee10f4cb6c4108659dd6d5/contracts/libs/SafeERC20.sol)
+// and 0x (https://github.com/0xProject/0x-monorepo/blob/737d1dc54d72872e24abce5a1dbe1b66d35fa21a/contracts/protocol/contracts/protocol/AssetProxy/ERC20Proxy.sol#L143)
+
+pragma solidity ^0.4.24;
+
+import "../lib/token/ERC20.sol";
+
+// Same as ERC20 except the return values of these interfaces have been removed to allow us to
+// manually check them
+interface GeneralERC20 {
```
bingen: I would call it BrokenERC20 ;-)

contracts/common/VaultRecoverable.sol
bingen: Can't you use `token.safeTransfer` here instead of `ERC20(_token).safeTransfer`?

contracts/test/mocks/SafeERC20Mock.sol
```diff
@@ -0,0 +1,28 @@
...
+
+import "../../common/SafeERC20.sol";
+import "../../lib/token/ERC20.sol";
+
+
+contract SafeERC20Mock {
+    using SafeERC20 for ERC20;
+    event Result(bool result);
+
+	function transfer(ERC20 token, address to, uint256 amount) external returns (bool) {
```
bingen: You know I don't mind these things ;-), but just in case you didn't notice there seems to be a mix of 4 and 2 spaces indentation here.

contracts/common/VaultRecoverable.sol
bingen: I think it's fine this way.

contracts/common/SafeERC20.sol
izqui: I propose `NERC20` or `ANERC20` ('accidentally not ERC20')

contracts/common/SafeERC20.sol
```diff
@@ -0,0 +1,68 @@
...
+            case 0x20 {
+                // Copy 32 bytes into scratch space
+                returndatacopy(0x0, 0x0, 0x20)
+
+                // Only return success if returned data was true
+                ret := eq(mload(0x0), 1)
+            }
+
+            // Not sure what was returned: don't mark as success
+            default { }
```
izqui: Really nice assembly 👏

izqui:
Will review in detail closer to releasing this

contracts/common/SafeERC20.sol
```diff
@@ -0,0 +1,68 @@
...
+
+            // No bytes returned: assume success
+            case 0x0 {
+                ret := 1
+            }
+
+            // 32 bytes returned: check if non-zero
+            case 0x20 {
+                // Copy 32 bytes into scratch space
+                returndatacopy(0x0, 0x0, 0x20)
```
izqui: I like how by doing this optimization we save a few `mload` and `mstore`. Given that we read this memory just immediately after we should be okey using this memory address, but I always feel a little uneasy that something else may be counting on this memory space not changing.

contracts/test/mocks/TokenReturnMissingMock.sol
```diff
@@ -0,0 +1,105 @@
+// Non-standards compliant token that is missing return values for
+// `transfer()`, `transferFrom()`, and `approve().
+// Modified from https://github.com/OpenZeppelin/openzeppelin-solidity/blob/a9f910d34f0ab33a1ae5e714f69f9596a02b4d91/contracts/token/ERC20/StandardToken.sol
+
+pragma solidity 0.4.24;
+
+import "../../lib/math/SafeMath.sol";
+
+
+contract TokenReturnMissingMock {
```
izqui: Maybe inherit from the `GeneralERC20` interface?

contracts/common/SafeERC20.sol
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+
+    /**
+    * @dev Same as a standards-compliant ERC20.transfer().
+    *      Note that this makes an external call to the token.
+    */
+    function safeTransfer(ERC20 _token, address _to, uint256 _amount) internal returns (bool) {
+        GeneralERC20(_token).transfer(_to, _amount);
+        return checkSuccess();
```
izqui: I'm actually wondering whether all these 'safe' functions should provide a unified way to handle errors.

Right now depending on implementation details at the token level `safeTransfer` can either revert or return false to communicate that the transfer has failed. I think it would be a much better API if the call was really safe and it always returned false on transfer errors (by doing a raw rather than using the standard Solidity function calling syntax that just reverts on reverts).

contracts/common/SafeERC20.sol
sohkai: > Something else may be counting on this memory space not changing

I really hope not... [`0x00 - 0x3f` is scratch space for doing keccaks](https://solidity.readthedocs.io/en/v0.4.24/miscellaneous.html#layout-in-memory), so you should never rely on storing something there unless you know you're going to retrieve it immediately.

contracts/test/mocks/TokenReturnMissingMock.sol
sohkai: Don't think this is necessary, especially as the other token mocks wouldn't inherit from it.

contracts/common/SafeERC20.sol
sohkai: Actually it looks like the scratch space could be used by solidity even in-between inline assembly statements, so it's not safe to ever use (except for throwaway values). Will use the free memory pointer.

contracts/common/SafeERC20.sol
```diff
@@ -0,0 +1,151 @@
...
+            success := staticcall(
+                gas,                  // forward all gas
+                _addr,                // address
+                add(_calldata, 0x20), // calldata start
+                mload(_calldata),     // calldata length
+                ptr,                  // write output over free memory
+                0x20                  // uint256 return
+            )
+
+            if gt(success, 0) {
```
sohkai: It turns out this is a micro optimization over `switch`. In particular, the docs say `call` will only ever return `1` on success, but everyone seems to use `case 0` for revert (and everything else as success) and so this mimics that behaviour.

sohkai:
@bingen @izqui Would be great if you could review [`SafeERC20`](https://github.com/aragon/aragonOS/pull/469/files#diff-363c77684f4b2f5a3f3c94a40f0032e8) again, given the assembly changes for https://github.com/aragon/aragonOS/pull/469#discussion_r244948954.

test/recovery_to_vault.js
```diff
@@ -64,6 +102,25 @@ contract('Proxy funds', accounts => {
     })
   }
 
+  // Token test groups
+  const tokenTestGroups = [
+    {
+      title: 'standards compliant, reverting token',
+      tokenContract: TokenMock,
+      revertsOnFailure: true,
```
bingen: Where are these being used?

sohkai: Ah good catch, I must've copied those over from the `safe_erc20` tests and not used them (since everything should revert in `VaultDepositable`)!

contracts/common/SafeERC20.sol
```diff
@@ -0,0 +1,151 @@
...
+pragma solidity ^0.4.24;
+
+import "../lib/token/ERC20.sol";
+
+
+library SafeERC20 {
+    // Before 0.5, solidity has a mismatch between `address.transfer()` and `token.transfer()`:
+    // https://github.com/ethereum/solidity/issues/3544
+    bytes4 private constant TRANSFER_SELECTOR = 0xa9059cbb;
+
```
izqui:
```suggestion
bytes4 private constant TRANSFER_FROM_SELECTOR = ERC20(0).transferFrom.selector;
```

contracts/common/SafeERC20.sol
```diff
@@ -0,0 +1,151 @@
...
+    string private constant ERROR_TOKEN_BALANCE_REVERTED = "SAFE_ERC_20_BALANCE_REVERTED";
+    string private constant ERROR_TOKEN_ALLOWANCE_REVERTED = "SAFE_ERC_20_ALLOWANCE_REVERTED";
+
+    function invokeAndCheckSuccess(address _addr, bytes memory _calldata)
+        private
+        returns (bool ret)
```
izqui: I would not use an implicit name return value for this, and I would explicitly have `bool ret = false` in the function body and then have an explicit `return ret` as it makes the function easier to understand

contracts/common/SafeERC20.sol
sohkai: Since this doesn't impact the bytecode and we're not doing it for the other selectors (`transfer` is a special case), won't be changing this.

PR #470, opened 2018-12-17
feat(SafeERC20): add static call utilities

sohkai:
Builds on #469.

Adds some `staticcall` utilities to the `ERC20` interface, as before solidity 0.5 these `balanceOf()` and `allowance()` functions were not called with a `staticcall`.

Useful for cases like [here](https://github.com/aragon/aragonOS/pull/470/files#diff-32170f8bb97d146f323f07e8ee82f9acR35), [here](https://github.com/aragon/aragon-apps/blob/master/apps/finance/contracts/Finance.sol#L362), or [here](https://github.com/aragon/aragon-apps/blob/master/apps/vault/contracts/Vault.sol#L77) where users may be unsuspecting of a function actually calling out to another contract (that may be malicious).

Bytecode diff:

```
ACL.json                       48600 more gas       +243               0
APMRegistry.json               49800 more gas       +249               0
AppStub.json                   48600 more gas       +243               0
AppStub2.json                  48600 more gas       +243               0
AppStubConditionalRecovery.js… 48600 more gas       +243               0
AppStubDepositable.json        48600 more gas       +243               0
AragonApp.json                 48600 more gas       +243               0
ENSSubdomainRegistrar.json     48600 more gas       +243               0
EVMScriptRegistry.json         48600 more gas       +243               0
EVMScriptRegistryFactory.json  Same                 0                  +243
Kernel.json                    49800 more gas       +249               0
KernelConstantsMock.json       49800 more gas       +249               0
KernelDepositableMock.json     49800 more gas       +249               0
KernelPinnedStorageMock.json   49800 more gas       +249               0
KernelSetAppMock.json          49800 more gas       +249               0
MockScriptExecutorApp.json     48600 more gas       +243               0
Repo.json                      48600 more gas       +243               0
SafeERC20Mock.json             195800 more gas      +979               0
TestACLInterpreter.json        48600 more gas       +243               0
UnsafeAppStub.json             48600 more gas       +243               0
UnsafeAppStubDepositable.json  48600 more gas       +243               0
UnsafeAragonApp.json           48600 more gas       +243               0
UnsafeAragonAppMock.json       48600 more gas       +243               0
UnsafeRepo.json                48600 more gas       +243               0
UpgradedKernel.json            49800 more gas       +249               0
VaultMock.json                 48600 more gas       +243               0
```

### DIFF
--- a/contracts/common/SafeERC20.sol
+++ b/contracts/common/SafeERC20.sol
@@ -92,24 +92,28 @@ library SafeERC20 {
         return checkSuccess();
     }
 
-    function staticBalanceOf(ERC20 _token, address _owner) internal view returns (uint256 tokenBalance) {
+    function staticBalanceOf(ERC20 _token, address _owner) internal view returns (uint256) {
         bytes memory balanceOfCallData = abi.encodeWithSelector(
             ERC20(_token).balanceOf.selector,
             _owner
         );
-        bool success;
-        (success, tokenBalance) = staticInvoke(_token, balanceOfCallData);
+
+        (bool success, uint256 tokenBalance) = staticInvoke(_token, balanceOfCallData);
         require(success, ERROR_TOKEN_BALANCE_REVERTED);
+
+        return tokenBalance;
     }
 
-    function staticAllowance(ERC20 _token, address _owner, address _spender) internal view returns (uint256 allowance) {
+    function staticAllowance(ERC20 _token, address _owner, address _spender) internal view returns (uint256) {
         bytes memory allowanceCallData = abi.encodeWithSelector(
             ERC20(_token).allowance.selector,
             _owner,
             _spender
         );
-        bool success;
-        (success, allowance) = staticInvoke(_token, allowanceCallData);
+
+        (bool success, uint256 allowance) = staticInvoke(_token, allowanceCallData);
         require(success, ERROR_TOKEN_ALLOWANCE_REVERTED);
+
+        return allowance;
     }
 }
--- a/contracts/common/SafeERC20.sol
+++ b/contracts/common/SafeERC20.sol
@@ -54,7 +54,7 @@ library SafeERC20 {
                 add(_calldata, 0x20), // calldata start
                 mload(_calldata),     // calldata length
                 0,                    // write output over scratch
-                32                    // uint256 return
+                0x20                  // uint256 return
             )
 
             switch success
--- a/contracts/common/SafeERC20.sol
+++ b/contracts/common/SafeERC20.sol
@@ -15,6 +15,9 @@ interface GeneralERC20 {
 
 
 library SafeERC20 {
+    string private constant ERROR_TOKEN_BALANCE_REVERTED = "SAFE_ERC_20_BALANCE_REVERTED";
+    string private constant ERROR_TOKEN_ALLOWANCE_REVERTED = "SAFE_ERC_20_BALANCE_REVERTED";
+
     function checkSuccess() private pure returns (bool ret) {
         assembly {
             // Check number of bytes returned from last function call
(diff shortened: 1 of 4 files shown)


contracts/common/SafeERC20.sol
```diff
@@ -65,4 +91,25 @@ library SafeERC20 {
...
     }
+
+    function staticBalanceOf(ERC20 _token, address _owner) internal view returns (uint256 tokenBalance) {
+        bytes memory balanceOfCallData = abi.encodeWithSelector(
+            ERC20(_token).balanceOf.selector,
+            _owner
+        );
+        bool success;
+        (success, tokenBalance) = staticInvoke(_token, balanceOfCallData);
+        require(success, ERROR_TOKEN_BALANCE_REVERTED);
```
izqui: I would do an explicit return

contracts/common/SafeERC20.sol
```diff
@@ -65,4 +91,25 @@ library SafeERC20 {
...
+
+    function staticAllowance(ERC20 _token, address _owner, address _spender) internal view returns (uint256 allowance) {
+        bytes memory allowanceCallData = abi.encodeWithSelector(
+            ERC20(_token).allowance.selector,
+            _owner,
+            _spender
+        );
+        bool success;
+        (success, allowance) = staticInvoke(_token, allowanceCallData);
+        require(success, ERROR_TOKEN_ALLOWANCE_REVERTED);
```
izqui: I would do an explicit return

contracts/common/SafeERC20.sol
```diff
@@ -39,6 +42,29 @@ library SafeERC20 {
...
+        returns (bool success, uint256 ret)
+    {
+        assembly {
+            success := staticcall(
+                gas,                  // forward all gas
+                _addr,                // address
+                add(_calldata, 0x20), // calldata start
+                mload(_calldata),     // calldata length
+                0,                    // write output over scratch
+                32                    // uint256 return
```
izqui:
```suggestion
                0x20                    // uint256 return
```

for consistency

contracts/common/SafeERC20.sol
sohkai: Main reason I tend to favour the implicit return is finding out that it's a little bit cheaper than an explicit return.

Although in this case, it's probably worth being explicit since the value's set as part of a tuple.

sohkai:
> Are we supposed to revert these changes when we upgrade to 0.5?

I'd say the upgrade would make it redundant, amongst the other static calls we do :). Having this handled transparently by the compiler is one of the better reasons in upgrading to 0.5.0 but it's likely going to be a while before that happens (and would require a major version upgrade), and until then, I'd like to provide a safer way to access this information in aragonOS 4.

> Finally if we assume the ERC token is malicious, we would still have a problem with the rest of the calls (trasnfer, etc)

This is a good point, although perhaps an argument could be made that this could protect against weird implementation mistakes rather than just malicious contracts. I do agree with the sentiment that having two external calls back-to-back ("supposedly static" balance then transfer) is a bit sketchy, even if we can't think of an immediate exploit.

Will spin up the gas reporter and bytecode comparison to see how much of a gas difference this makes compared to a normal call (I suspect very little).

contracts/common/SafeERC20.sol
```diff
@@ -15,6 +15,9 @@ interface GeneralERC20 {
 
 
 library SafeERC20 {
+    string private constant ERROR_TOKEN_BALANCE_REVERTED = "SAFE_ERC_20_BALANCE_REVERTED";
+    string private constant ERROR_TOKEN_ALLOWANCE_REVERTED = "SAFE_ERC_20_BALANCE_REVERTED";
```
bingen:
```suggestion
    string private constant ERROR_TOKEN_ALLOWANCE_REVERTED = "SAFE_ERC_20_ALLOWANCE_REVERTED";
```

sohkai:
@bingen I've updated https://github.com/aragon/aragonOS/pull/470#issue-239276237 with the bytecode diff; it does increase the contract creation cost a bit for more code but doesn't impact the proxies so I think it's fine.

Running the gas profiler though, having this change actually _decreases_ the cost of the [`VaultRecoverable.transferToVault()`](https://github.com/aragon/aragonOS/blob/dev/contracts/common/VaultRecoverable.sol#L22) implementation by a nominal amount of gas on tokens (profiler shows ~400-500 saved).

PR #472, opened 2019-01-20
chore: pin ganache-cli to 6.2.3

sohkai:
Check with the CI that `ganache-cli@6.2.3` and `ganache-core@2.3.1` are not breaking our tests.


sohkai:
Will merge this for now since `ganache-cli@6.4.0` is still having testing problems (see https://github.com/trufflesuite/ganache-core/issues/283).

Issue #454, opened 2018-11-08
EVMScriptRegistry should also log the type of the executor being enabled / disabled


sohkai:
Adding this information makes it much easier to figure out which executors in an org are installed / enabled from a frontend.

sohkai:
Closing as this can be read off-chain with a standard interface: https://github.com/aragon/aragonOS/pull/490.

Issue #455, opened 2018-11-08
Refactor EVMScriptRegistry to be based on types, rather than IDs


sohkai:
Seems much more reasonable if the information in the registry was accessible by an executor's type rather than the ID it was installed with, since it'd deterministic across all organizations.

That way, instead of passing a `SPEC_ID` in scripts, we could pass a `SPEC_TYPE`. Doing so would make it easier for apps to declare their forwarding interface requires a particular executor type (albeit at a small cost disadvantage).

We could still accomplish apps requiring specific-forwards with `SPEC_ID`s, but it may be harder for a frontend to evaluate if an org has a particular executor type enabled without #454.

PR #473, opened 2019-01-20
emit executorType

rperez89:
PR for [log executor type](https://github.com/aragon/aragonOS/issues/454) 
@sohkai let me know if this was what we need


sohkai:
Hey @rperez89, looking at your implementation made me realize that we could get away with not doing #455 if we had these types exposed on the contract! I think we can go with https://github.com/aragon/aragonOS/pull/490 for now, if others build and install custom executors.

Thanks for doing this, and sorry we didn't merge it!

PR #474, opened 2019-02-08
tests: Coverage improvements

sohkai:
Improves the coverage of some missing branches.

### DIFF
--- a/contracts/test/mocks/APMRegistryFactoryMock.sol
+++ b/contracts/test/mocks/APMRegistryFactoryMock.sol
@@ -1,31 +1,49 @@
 pragma solidity 0.4.24;
 
+import "../../apm/APMRegistry.sol";
+import "../../apm/Repo.sol";
+import "../../ens/ENSSubdomainRegistrar.sol";
+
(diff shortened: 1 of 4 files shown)


test/evm_script.js
```diff
@@ -326,37 +336,23 @@ contract('EVM Script', accounts => {
                 })
             })
 
-            context('registry', () => {
-                it('can be disabled', async () => {
+            context('> Registry actions', () => {
```
sohkai: Note that these tests have been simplified as most of the functionality was [already tested earlier](https://github.com/aragon/aragonOS/blob/2042452a3b248a56e98339e3e6cbccd6e888b2fc/test/evm_script.js#L92).

contracts/test/mocks/APMRegistryFactoryMock.sol
```diff
@@ -1,31 +1,49 @@
 pragma solidity 0.4.24;
```
sohkai: This mock has been refactored to be simpler, and more targeted about what's missing.

PR #480, opened 2019-02-20
VaultRecoverable: emit event on successful recovery

sohkai:
Not sure if we made a conscious decision before to not emit an event, but it seems useful to know about for frontends and successful recoveries.

### DIFF
--- a/contracts/common/IVaultRecoverable.sol
+++ b/contracts/common/IVaultRecoverable.sol
@@ -6,7 +6,7 @@ pragma solidity ^0.4.24;
 
 
 interface IVaultRecoverable {
-    event RecoverToVault(address token, uint256 amount);
+    event RecoverToVault(address indexed token, uint256 amount);
 
     function transferToVault(address token) external;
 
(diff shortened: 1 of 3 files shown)


contracts/common/IVaultRecoverable.sol
```diff
@@ -6,6 +6,8 @@ pragma solidity ^0.4.24;
 
 
 interface IVaultRecoverable {
+    event RecoverToVault(address token, uint256 amount);
```
izqui: Let's index the token address and also log the vault address where the tokens or eth are sent

sohkai: Also added the recovery vault as a parameter (since this could change in the org later).

test/recovery_to_vault.js
```diff
@@ -134,7 +145,7 @@ contract('Recovery to vault', accounts => {
 
   // Test both the Kernel itself and the KernelProxy to make sure their behaviours are the same
   for (const kernelType of ['Kernel', 'KernelProxy']) {
-    context(`> ${kernelType}`, () => {
+    context.only(`> ${kernelType}`, () => {
```
bingen: Why are we adding `.only` here? Are we trying to skip something?

sohkai: Fixed in https://github.com/aragon/aragonOS/pull/483

PR #490, opened 2019-03-07
feat: add executorType() as a required interface for EVMScriptExecutors

sohkai:
Simply forces subclasses of `BaseEVMScriptExecutor` to include `executorType()` as an interface they expose.

This alleviates the need for https://github.com/aragon/aragonOS/issues/455 until we want to make that breaking change. If every executor implements the same `executorType()` interface, frontends can reduce the set of enabled executors and find out which executor spec ID a particular forwarding interaction may require.


sohkai:
@bingen Hmm, yes, I don't know how I missed that earlier 😅. Thanks for noticing!

PR #496, opened 2019-03-20
EVMScripts: check byte lengths, avoid returning unallocated memory, and forward error data

sohkai:
Add length checks to bytes passed to `EVMScriptRegistry` and `CallsScript` and makes sure `EVMScriptRunner` doesn't return an unallocated bytes array if no return data is received from `execScript()`.

Also transparently forwards any error data if a script executor or call script failed.

Bytecode diff (after removing the error strings):

```
CallsScript.json              30400 more gas       +152               0
EVMScriptExecutorMock.json    13200 more gas       +66                0
EVMScriptRegistry.json        3400 more gas        +17                0
EVMScriptRegistryFactory.json Same                 0                  +169
ExecutionTarget.json          84600 more gas       +423               0
```


contracts/evmscript/EVMScriptRegistry.sol
```diff
@@ -19,8 +19,11 @@ contract EVMScriptRegistry is IEVMScriptRegistry, EVMScriptRegistryConstants, Ar
     // WARN: Manager can censor all votes and the like happening in an org
```
sohkai: Note that upgrading this app for existing organizations will be a PITA; we've [pinned these apps to their current implementations](https://github.com/aragon/aragonOS/blob/dev/contracts/factory/EVMScriptRegistryFactory.sol#L23) and so these orgs will have to reset their default `EVMScriptRegistry` app in the kernel to get this change (along with uninstalling the old one completely, for the frontend to not look weird).

contracts/evmscript/executors/CallsScript.sol
```diff
@@ -25,11 +25,14 @@ contract CallsScript is BaseEVMScriptExecutor {
     * @param _script [ specId (uint32) ] many calls with this structure ->
```
sohkai: Note that upgrading this is also a PITA, for similar reasons to `EVMScriptRegistry`.

contracts/evmscript/executors/CallsScript.sol
```diff
@@ -55,6 +58,7 @@ contract CallsScript is BaseEVMScriptExecutor {
 
             require(success, ERROR_CALL_REVERTED);
         }
+        // No need to allocate empty bytes for the return as this can only be called via an delegatecall
```
sohkai: Given the modifier checks this has, it's actually impossible to call this without going through a delegatecall.

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -50,16 +50,19 @@ contract EVMScriptRunner is AppStorage, Initializable, EVMScriptRegistryConstant
...
+
+            ret := mload(0x40) // free mem ptr get
+            mstore(0x40, add(ret, add(size, 0x20))) // free mem ptr set
+
             switch size
             case 0 {}
             default {
-                ret := mload(0x40) // free mem ptr get
-                mstore(0x40, add(ret, add(size, 0x20))) // free mem ptr set
                 returndatacopy(ret, 0x20, sub(size, 0x20)) // copy return data
```
facuspagnuolo: If I don't get this wrong we are ignoring the first word of the return data, right? Is it sth intentional?

contracts/evmscript/executors/CallsScript.sol
facuspagnuolo: It may be obvious, but why the `isInitialized` would prevent a normal call?

test/evm_script.js
```diff
@@ -313,8 +311,56 @@ contract('EVM Script', accounts => {
...
+                it('fails if data length is too small to contain address', async () => {
+                    const action = { to: executionTarget.address, calldata: executionTarget.contract.execute.getData() }
+                    const script = encodeCallScriptAddressUnderflow([action])
+
+                    return assertRevert(async () => {
+                        await executorApp.execute(script)
+                    })
+                })
+
+                it('fails if data length is too small to contain length', async () => {
```
facuspagnuolo: maybe
```suggestion
                it('fails if data length is too small to contain calldata', async () => {
```

test/evm_script.js
```diff
@@ -313,8 +311,56 @@ contract('EVM Script', accounts => {
...
+                const encodeCallScriptAddressUnderflow = actions => {
+                    return actions.reduce((script, { to }) => {
+                        const addr = rawEncode(['address'], [to]).toString('hex')
+
+                        // Remove too much of the address (should just remove first 12 0s of padding as addr)
+                        return script + addr.slice(24 + 4)
+                    }, '0x00000001') // spec 1
+                }
+
+                const encodeCallScriptLengthUnderflow = actions => {
```
facuspagnuolo: calldata underflow maybe?

contracts/evmscript/executors/CallsScript.sol
sohkai: `isInitialized` checks against the slot at `INITIALIZATION_BLOCK_POSITION`, which is left false in this executor contract (executors are deployed as-is).

We added `isInitialized` in an attempt to make calling these scripts harder to do outside of Aragon apps (before anyone could make a call to them), so this protects against:

- Direct calls (imagine if this script changed some state; we would only want to do so if it was `delegatecall`ed into)
- Usage from uninitialized (or "petrified") Aragon apps

You could still make a contract that sets the storage slot and `delegatecall`s into this executor.

contracts/evmscript/EVMScriptRunner.sol
sohkai: Yes... I think so... We expect the returndata to be a `bytes` array, so the first element is the size of the array (and hence the "ABI decode").

contracts/evmscript/executors/CallsScript.sol
facuspagnuolo: Thanks for clarifying :)

contracts/evmscript/EVMScriptRunner.sol
sohkai: Updated in f955b3d; see https://github.com/aragon/aragonOS/pull/497#discussion_r268651597 for discussion.

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -50,17 +50,25 @@ contract EVMScriptRunner is AppStorage, Initializable, EVMScriptRegistryConstant
...
+
+            switch returndatasize
             case 0 {}
             default {
-                ret := mload(0x40) // free mem ptr get
-                mstore(0x40, add(ret, add(size, 0x20))) // free mem ptr set
-                returndatacopy(ret, 0x20, sub(size, 0x20)) // copy return data
+                // Copy return data
+                // Apply ABI decode by ignoring the first 32 bytes of the return data
+                returndatacopy(ret, 0x20, sub(returndatasize, 0x20))
```
sohkai: @facuspagnuolo Confirmed using `returndatasize` is much cheaper; diffing https://github.com/aragon/aragonOS/pull/496/commits/f73c45af3f40c1c6b4b84b1c88d15c6f7678a907 against its previous commit results in 4 less opcodes :).

test/helpers/assertThrow.js
```diff
@@ -1,26 +1,38 @@
-function assertError(error, s, message) {
```
sohkai: Pulled from https://github.com/aragon/aragon-apps/pull/741.

sohkai:
Added tests for the returned bytes in cecb70067b3d01d5224c27c093780a4610f69c3c.

Polished the tests:

- Separated the `EVMScriptRegistryFactory()` usage into its own test
- Uses the new `assertThrows` module from https://github.com/aragon/aragon-apps/pull/741
- Some other bits of clean up or renaming to clarify the existing tests

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -40,30 +39,44 @@ contract EVMScriptRunner is AppStorage, Initializable, EVMScriptRegistryConstant
...
+                add(data, 0x20),    // calldata start
+                mload(data),        // calldata length
+                0,                  // don't write output (we'll handle this ourselves)
+                0                   // don't write output
+            )
 
-        return output;
-    }
+            output := mload(0x40) // free mem ptr get
+            mstore(0x40, add(output, add(returndatasize, 0x20))) // free mem ptr set
```
izqui: This was already in the code, but I think we are allocating 32 bytes more of memory than we need when the call is successful, as we only copy `returndatasize - 0x20` bytes.

contracts/evmscript/executors/CallsScript.sol
```diff
@@ -50,11 +56,43 @@ contract CallsScript is BaseEVMScriptExecutor {
...
+                    case 0 {
+                        // No error data was returned, revert with "EVMCALLS_CALL_REVERTED"
+                        // See remix: doing a `revert("EVMCALLS_CALL_REVERTED")` always results in
+                        // this memory layout
+                        mstore(ptr, 0x08c379a000000000000000000000000000000000000000000000000000000000)
+                        mstore(add(ptr, 0x20), 0x0000002000000000000000000000000000000000000000000000000000000000)
+                        mstore(add(ptr, 0x40), 0x0000001645564d43414c4c535f43414c4c5f5245564552544544000000000000)
+                        mstore(add(ptr, 0x60), 0)
+
+                        revert(ptr, 100)
```
izqui: I'd add a comment for the `100` magic number.

```suggestion
                        revert(ptr, 100) // 100 = 4 + 3 * 32 (error identifier + 3 words for the ABI encoded error)
```

facuspagnuolo: It also may be clearer using correctly padded data with fixed indexes, it is more readable I think, and it somehow mimics what the EVM actually does.
```sol
mstore(ptr, 0x08c379a000000000000000000000000000000000000000000000000000000000)         // error identifier
mstore(add(ptr, 0x04), 0x0000000000000000000000000000000000000000000000000000000000000020) // starting offset
mstore(add(ptr, 0x24), 0x0000000000000000000000000000000000000000000000000000000000000016) // reason length
mstore(add(ptr, 0x44), 0x45564d43414c4c535f43414c4c5f524556455254454400000000000000000000) // reason
```

contracts/evmscript/EVMScriptRunner.sol
sohkai: Actually it turns out we were allocating 64 bytes more:

- Assume the current free mem ptr is at 0x80
- `returndatasize` is 0x60
- We copy 0x40 (0x60 - 0x20) bits onto 0x80
- Next free pointer should be 0xc0 (0x80, 0xa0 are taken by the copy)

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -70,7 +69,10 @@ contract EVMScriptRunner is AppStorage, Initializable, EVMScriptRegistryConstant
                 //    [ output content (N bytes) ]
                 //
                 // Perform the ABI decode by ignoring the first 32 bytes of the return data
-                returndatacopy(output, 0x20, sub(returndatasize, 0x20))
+                let copysize := sub(returndatasize, 0x20)
+                returndatacopy(output, 0x20, copysize)
+
+                mstore(0x40, add(output, copysize)) // free mem ptr set
```
sohkai: Note that this **will* fail the new tests if `sub(copysize, 0x20)` is used.

sohkai: Note that this **will** fail the new tests if `sub(copysize, 0x20)` is used.

PR #497, opened 2019-03-20
EVMScriptRunner, CallsScript: transparently revert errors rather than emit its own string

sohkai:
I think returning the actual execution error would be more clear for debugging, since it'd otherwise be impossible to get this revert string (unless we can see it in the trace?)

Bytecode diff against #496 (after removing the error strings):

```
AppStubScriptRunner.json      25800 less gas       -129               0
CallsScript.json              2600 more gas        +13                0
EVMScriptRegistryFactory.json Same                 0                  +13
ExecutionTarget.json          82400 more gas       +412               0
```


contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -40,33 +39,35 @@ contract EVMScriptRunner is AppStorage, Initializable, EVMScriptRegistryConstant
...
+                returndatacopy(output, 0, size) // copy full return data
+                revert(output, size)
+            }
             default {
-                returndatacopy(ret, 0x20, sub(size, 0x20)) // copy return data
+                switch size
+                case 0 {}
+                default {
+                    // Copy and ABI decode result, since `execScript()`'s return type is `bytes`
+                    mstore(output, size)
```
sohkai: I think this is actually something we missed before; we weren't setting the size correctly in the old output.

izqui: Good catch

facuspagnuolo: The first word is used to tell the return data size?

sohkai: See https://github.com/aragon/aragonOS/pull/496/files#r268555232.

Although now I'm thinking; maybe this "ABI decode" is unnecessary and we can just do

```solidity
returndatacopy(output, 0x0, size)
```

Since the bytes length would be the same (as opposed to something like a uint256 array)?

-----------

In any case, this is still wrong; it should be `sub(size, 0x20)`.

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -40,33 +39,35 @@ contract EVMScriptRunner is AppStorage, Initializable, EVMScriptRegistryConstant
...
-    }
-
-    /**
-    * @dev Copies and returns last's call data. Needs to ABI decode first
-    */
-    function returnedDataDecoded() internal pure returns (bytes) {
-        bytes memory ret;
+        bytes memory output;
         assembly {
             let size := returndatasize
```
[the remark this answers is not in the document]
sohkai: Cool! Didn't realize it was just `2 gas` to call!

contracts/evmscript/EVMScriptRunner.sol
facuspagnuolo: Exactly, I think we can just do `returndatacopy(output, 0x0, returndatasize)`

sohkai:
@facuspagnuolo @izqui Refactored the entire `delegatecall` sequence in af4b0f5, to be a little bit more optimized and similar to implementations we have elsewhere in the codebase (e.g. `SafeERC20`).

contracts/evmscript/EVMScriptRunner.sol
```diff
@@ -40,33 +39,35 @@ contract EVMScriptRunner is AppStorage, Initializable, EVMScriptRegistryConstant
...
+            output := mload(0x40) // free mem ptr get
+            mstore(0x40, add(output, add(size, 0x20))) // free mem ptr set
 
-            switch size
-            case 0 {}
+            // If the call returned error data, forward it
+            switch result
+            case 0 {
+                returndatacopy(output, 0, size) // copy full return data
+                revert(output, size)
```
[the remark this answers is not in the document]
sohkai: Hmm, not sure how safe it is to use the `0` memory slot, which solidity reserves.

contracts/evmscript/EVMScriptRunner.sol
izqui: Recording an offline conversation: because the returned data is ABI encoded, the actual payload looked like this:

```
[ position of the first dynamic length return value = 0x20 (32 bytes) ]
[ output length (32 bytes) ]
[ output content (N bytes) ]
```

Because of this we could copy into the output array all the returned bytes minus the first word (as we assume the return types) directly with: `returndatacopy(ret, 0x20, sub(returndatasize, 0x20))`

sohkai:
@izqui @facuspagnuolo Ready for review again.

Not sure about 8f8756fced87f36c268d5300ffa41fdfd5209d76; @izqui do you think it's better if we don't forward a failing call's revert data in `CallsScripts`?

izqui:
> do you think it's better if we don't forward a failing call's revert data in CallsScripts?

I think this is always better if there is error data to be forwarded. An option could be to return the default error if a call reverted without data, but unsure about the added complexity.

sohkai:
> An option could be to return the default error if a call reverted without data, but unsure about the added complexity.

Will do in #496.

PR #499, opened 2019-03-26
ACL: revert if attempting to save empty params

sohkai:


contracts/acl/ACL.sol
```diff
@@ -43,9 +43,10 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
 
     uint256 internal constant ORACLE_CHECK_GAS = 30000;
 
-    string private constant ERROR_AUTH_INIT_KERNEL = "ACL_AUTH_INIT_KERNEL";
     string private constant ERROR_AUTH_NO_MANAGER = "ACL_AUTH_NO_MANAGER";
     string private constant ERROR_EXISTENT_MANAGER = "ACL_EXISTENT_MANAGER";
+    string private constant ERROR_AUTH_INIT_KERNEL = "ACL_AUTH_INIT_KERNEL";
+    string private constant ERROR_SAVE_EMPTY_HASH = "ACL_SAVE_EMPTY_HASH";
```
facuspagnuolo: Are we using this one?

sohkai: Ehh yes... it looks like this suggestion commit did something funny: https://github.com/aragon/aragonOS/pull/499/commits/17fbaae47562f22d3c85a533b08635f0404c0fbf.

Fixed now in ded7a42.

PR #500, opened 2019-03-26
ACL: optimize checkOracle()

sohkai:
Updates the `ACL`'s staticcall to use a more-optimized one similar to `SafeERC20`'s `invokeAndCheckSuccess()`.

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -428,31 +428,37 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     function checkOracle(IACLOracle _oracleAddr, address _who, address _where, bytes32 _what, uint256[] _how) internal view returns (bool) {
         bytes4 sig = _oracleAddr.canPerform.selector;
 
-        // a raw call is required so we can return false if the call reverts, rather than reverting
         bytes memory checkCalldata = abi.encodeWithSelector(sig, _who, _where, _what, _how);
         uint256 oracleCheckGas = ORACLE_CHECK_GAS;
-
-        bool ok;
-        assembly {
-            ok := staticcall(oracleCheckGas, _oracleAddr, add(checkCalldata, 0x20), mload(checkCalldata), 0, 0)
-        }
-
-        if (!ok) {
-            return false;
-        }
-
-        uint256 size;
-        assembly { size := returndatasize }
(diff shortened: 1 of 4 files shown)


contracts/acl/ACL.sol
```diff
@@ -428,31 +428,37 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
+                oracleCheckGas,           // gas forwarded
+                _oracleAddr,              // address
+                add(checkCalldata, 0x20), // calldata start
+                mload(checkCalldata),     // calldata length
+                ptr,                      // write output over free memory
+                0x20                      // uint256 return
+            )
+
+            if gt(success, 0) {
+                // Check number of bytes returned from last function call
```
izqui:
```suggestion
                // Check number of bytes returned from last external call
```

contracts/acl/ACL.sol
```diff
@@ -428,31 +428,37 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
+            let success := staticcall(
+                oracleCheckGas,           // gas forwarded
+                _oracleAddr,              // address
+                add(checkCalldata, 0x20), // calldata start
+                mload(checkCalldata),     // calldata length
+                ptr,                      // write output over free memory
+                0x20                      // uint256 return
+            )
+
+            if gt(success, 0) {
```
izqui: Looks like coverage is crashing when instrumenting this line

sohkai: Ahh shoot, I remember having this issue with `SafeERC20` but ended up putting that in the skip list.

contracts/acl/ACL.sol
```diff
@@ -428,31 +428,39 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
+            result := staticcall(
+                oracleCheckGas,           // gas forwarded
+                _oracleAddr,              // address
+                add(checkCalldata, 0x20), // calldata start
+                mload(checkCalldata),     // calldata length
+                ptr,                      // write output over free memory
+                0x20                      // uint256 return
+            )
+
+            // solidity-coverage fails on assembly `if`
```
facuspagnuolo: 🤷‍♂️

PR #501, opened 2019-03-26
feat: add ConversionHelpers lib for converting memory types

sohkai:
Consolidates all the bytes<>uint256[] conversions into a library.

It's not _too_ costly to add, and hopefully makes us all feel a bit better about this bit 😄.

Also fixes the solidity test runner, which must've broke at some point along the way (due to the assert logs not being decoded properly) 😅.

Bytecode diff:

```
ACL.json                      5400 more gas        +27                0
APMRegistry.json              4400 more gas        +22                0
AragonApp.json                4000 more gas        +20                0
ENSSubdomainRegistrar.json    4000 more gas        +20                0
EVMScriptRegistry.json        4000 more gas        +20                0
EVMScriptRegistryFactory.json Same                 0                  +20
Kernel.json                   8000 less gas        -40                0
Repo.json                     4000 more gas        +20                0
TestACLInterpreter.json       5400 more gas        +27                0
```

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -1,6 +1,7 @@
 pragma solidity 0.4.24;
 
 import "../apps/AragonApp.sol";
+import "../common/ConversionHelpers.sol";
 import "../common/TimeHelpers.sol";
 import "./ACLSyntaxSugar.sol";
 import "./IACL.sol";
@@ -242,17 +243,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     * @return boolean indicating whether the ACL allows the role or not
     */
     function hasPermission(address _who, address _where, bytes32 _what, bytes memory _how) public view returns (bool) {
-        // Force cast the bytes array into a uint256[], by overwriting its length
-        // Note that the uint256[] doesn't need to be initialized as we immediately overwrite it
-        // with _how and a new length, and _how becomes invalid from this point forward
-        uint256[] memory how;
-        uint256 intsLength = _how.length / 32;
-        assembly {
-            how := _how
-            mstore(how, intsLength)
-        }
-
-        return hasPermission(_who, _where, _what, how);
+        return hasPermission(_who, _where, _what, ConversionHelpers.castBytesToUintArray(_how));
     }
 
     function hasPermission(address _who, address _where, bytes32 _what, uint256[] memory _how) public view returns (bool) {
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -243,7 +243,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     * @return boolean indicating whether the ACL allows the role or not
     */
     function hasPermission(address _who, address _where, bytes32 _what, bytes memory _how) public view returns (bool) {
-        return hasPermission(_who, _where, _what, ConversionHelpers.castBytesToUintArray(_how));
+        return hasPermission(_who, _where, _what, ConversionHelpers.dangerouslyCastBytesToUintArray(_how));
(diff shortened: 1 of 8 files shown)


contracts/acl/ACL.sol
```diff
@@ -242,17 +243,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
-        // with _how and a new length, and _how becomes invalid from this point forward
-        uint256[] memory how;
-        uint256 intsLength = _how.length / 32;
-        assembly {
-            how := _how
-            mstore(how, intsLength)
-        }
-
-        return hasPermission(_who, _where, _what, how);
+        return hasPermission(_who, _where, _what, ConversionHelpers.castBytesToUintArray(_how));
```
facuspagnuolo: Is there any downside if we take advantage of libraries syntax sugar to do sth like `_how.toUintArray()`? (ofc, declaring the usage of `ConversionHelpers` for `bytes`)

contracts/common/ConversionHelpers.sol
```diff
@@ -0,0 +1,26 @@
...
+        // Force cast the bytes array into a uint256[], by overwriting its length
+        // Note that the uint256[] doesn't need to be initialized as we immediately overwrite it
+        // with the input and a new length. The input becomes invalid from this point forward.
+        uint256 intsLength = _input.length / 32;
+        assembly {
+            output := _input
+            mstore(output, intsLength)
+        }
+    }
+}
```
facuspagnuolo: WDYT about having separate helper libraries for uint arrays and bytes? I'm just thinking in case we need to provide more helper functions, having them split by type may make sense.

contracts/acl/ACL.sol
sohkai: I actually want to avoid using the syntax sugar for these helpers, to make them more explicit so there's less chance for mistakes.

Particularly, the `cast` nomenclature may be helpful to communicate that the passed in variable shouldn't be used anymore.

sohkai: I've now renamed these cast functions to be `dangerouslyCastX()` in 410ad80, similar to how React has [`dangerouslySetInnerHTML()`](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml), to make the intent more explicit.

facuspagnuolo: I see and I liked the naming idea :)

contracts/test/TestConversionHelpers.sol
```diff
@@ -0,0 +1,139 @@
...
+
+        // Check memory position (conversion should be in place)
+        uint256 arrMemLoc;
+        uint256 arrBytesMemLoc;
+        assembly {
+            arrMemLoc := arr
+            arrBytesMemLoc := arrBytes
+        }
+        Assert.equal(arrMemLoc, arrBytesMemLoc, "should have same memory location after conversion");
+    }
```
facuspagnuolo: We could also check values here

```suggestion

        // Check values
        uint256 first;
        uint256 second;
        uint256 third;
        assembly {
            first := mload(add(arrBytesMemLoc, 0x20))
            second := mload(add(arrBytesMemLoc, 0x40))
            third := mload(add(arrBytesMemLoc, 0x60))
        }
        Assert.equal(first, FIRST, "should have correct first value");
        Assert.equal(second, SECOND, "should have correct second value");
        Assert.equal(third, THIRD, "should have correct third value");
    }
```

sohkai: Hehe yes... I was thinking about doing this and then got a bit lazy to do the mloads. It's a bit of a superfluous test since we're just accessing raw memory that we know we set, but still good to make sure the cast doesn't do something funny to the memory 👍.

PR #506, opened 2019-04-08
test: fix test parameterization variable binding

sohkai:
Oops, previously some parameterized tests would only run the last parameterization repeatedly.

**Learning**: We (especially me 😅) should remember to **ALWAYS** bind the parameterized variables locally when doing this type of parameterization. This can be done easily by using `const` or `let` in the `for/of` declaration, or doing it locally inside the block.

Using a `var` declaration (default) causes things to break because Mocha does two passes on the tests:

1. Collect test cases by running everything not in a test hook (e.g. `before()`, `beforeEach()`, `it()`, etc.)
1. Actually run the test cases.

Because of this, any for loop's `var` declaration will run through 1 (collection), and always end up as the array's last item when 2 (test running) starts.

Thanks to @osarrouy for noticing this 😄.

### DIFF
--- a/contracts/test/TestConversionHelpers.sol
+++ b/contracts/test/TestConversionHelpers.sol
@@ -129,13 +129,13 @@ contract TestConversionHelpers {
         Assert.equal(arrMemLoc, arrReconvertedMemLoc, "should have same memory location after reconverting");
     }
 
-    function assertValues(uint256[] memory _data) {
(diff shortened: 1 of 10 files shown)


sohkai:
@all-contributors please add @osarrouy for testing

facuspagnuolo:
Good to know :)

PR #510, opened 2019-04-15
New script deploy-apm-open

0xGabi:
Truffle script draft to deploy **open.aragonpm.eth** `APMRegistry`. 

Related discussion: https://github.com/aragon/aragen/issues/10

I'm having troubles to understand how to assign ENS name **open.aragonpm.eth** to `apmFactory` after the first `APMRegistry` (**aragonpm.eth**) was deployed. 

For what I understand the `ensSubdomainRegistrar`of the deployed `APMRegistry` is the owner of the node **aragonpm.eth** on the `ENS Registry`. But I have tried to update the owner both with `ens.setSubnodeOwner` and `ensSubdomainRegistrar.createName` but none seems to work.

### DIFF
--- a/scripts/deploy-apm-open.js
+++ b/scripts/deploy-apm-open.js
@@ -1,0 +1,93 @@
+const namehash = require('eth-ens-namehash').hash
+const keccak256 = require('js-sha3').keccak_256
+
+const deployAPM = require('./deploy-apm')
+
+const globalArtifacts = this.artifacts // Not injected unless called directly via truffle
+const globalWeb3 = this.web3 // Not injected unless called directly via truffle
+
+const defaultOwner = process.env.OWNER
+const defaultDaoFactoryAddress = process.env.DAO_FACTORY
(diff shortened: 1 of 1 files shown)


izqui:
Before calling `ensSubdomainRegistrar.createName` you need to grant the `CREATE_NAME_ROLE` to the account that is performing the transactions. After creating the subdomain, you can revoke the permission.

sohkai:
@0xGabi Perhaps we can move this script to aragen instead?

0xGabi:
@izqui I was trying to do that using `apm.address` as the account  ([this line](https://github.com/aragon/aragonOS/blob/deploy-apm-open/scripts/deploy-apm-open.js#L60)), still not fully understand why this is not possible cause the APM instance have this role granted when created by the `apmFactory` ([this line](https://github.com/aragon/aragonOS/blob/dev/contracts/factory/APMRegistryFactory.sol#L100))

@sohkai Make sense, I'll close this and open one on aragen

PR #513, opened 2019-04-18
Testing improvements

sohkai:
Adds some improvements to the test suite, as per the Consensys Diligence advisory.

### DIFF
--- a/test/helpers/runSolidityTest.js
+++ b/test/helpers/runSolidityTest.js
@@ -27,7 +27,7 @@ const HOOKS_MAP = {
   afterAll: 'afterAll',
 }
 
-const processResult = txReceipt => {
+const processResult = (txReceipt, mustAssert) => {
   if (!txReceipt || !txReceipt.receipt) {
     return
   }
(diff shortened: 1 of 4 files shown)


test/helpers/runSolidityTest.js
```diff
@@ -37,6 +37,9 @@ const processResult = txReceipt => {
       throw new Error(log.args.message)
     }
   })
+  if (mustAssert && !decodedLogs.length) {
+    throw new Error('No assertions made')
+  }
```
facuspagnuolo: Don't know if this is necessary, test fwks usually don't throw if there are no assertions, they simply throw in case there was at least one failure.

sohkai: Yes... it is kind of "special" behaviour but for our purposes I think having this check is more robust:

1. Tests without explicit assertions made are probably buggy in some way
1. This implicitly tests that the test framework is working (decoding `Assert`'s events), avoiding an earlier issue where the tests were always just passing regardless of actual test behaviour.

PR #517, opened 2019-04-30
Organize mocks and test files

facuspagnuolo:
Simply organizing a bit the mock contract files and test files to follow the same directories structure we have under `contracts/`


facuspagnuolo:
Sure @sohkai, I wrongly branched from `master`, rebase on top of `dev` and merge, thx!

PR #518, opened 2019-04-30
[WIP] Application kill switch

facuspagnuolo:
Follow up on #516 

Based on an offline discussion we had about the next-steps regarding the kill-switch topic, we decided to:
[x] Drop the kernel level kill-switch idea, at least for the MVP version
[x] Simplify the whole architecture to support only application level kill-switch using severities
[x] Support having one issue registry per application
[ ] Integrate with the rest of the aragonOS components
[ ] Keep investigating how an organization is able to recover (for now we are leaving on a side the idea of allowing DAOs to programmatically override the application decision of how it should behave when the kill-switch is turned on)


contracts/kill_switch/KillSwitch.sol
```diff
@@ -0,0 +1,122 @@
...
+        external
+        authP(SET_ISSUES_REGISTRY_ROLE, arr(_contract, msg.sender))
+    {
+        require(isContract(_issuesRegistry), ERROR_ISSUES_REGISTRY_NOT_CONTRACT);
+        contractSettings[_contract].issuesRegistry = _issuesRegistry;
+        emit IssuesRegistrySet(_contract, address(_issuesRegistry));
+    }
+
+    function shouldDenyCallingContract(address _contract) external returns (bool) {
+        // if the call should be denied, then deny given call
```
bingen: That sounds pretty obvious, haha ;-P Maybe first "call" should be "contract"?

contracts/kill_switch/KillSwitch.sol
```diff
@@ -0,0 +1,122 @@
...
+    function setIssuesRegistry(address _contract, IIssuesRegistry _issuesRegistry)
+        external
+        authP(SET_ISSUES_REGISTRY_ROLE, arr(_contract, msg.sender))
+    {
+        require(isContract(_issuesRegistry), ERROR_ISSUES_REGISTRY_NOT_CONTRACT);
+        contractSettings[_contract].issuesRegistry = _issuesRegistry;
+        emit IssuesRegistrySet(_contract, address(_issuesRegistry));
+    }
+
+    function shouldDenyCallingContract(address _contract) external returns (bool) {
```
bingen: It would be nice to have an estimation of how much gas overload means adding the modifier that calls this to a function.

contracts/kill_switch/KillSwitch.sol
```diff
@@ -0,0 +1,122 @@
...
+        // if the call should be denied, then deny given call
+        if (isContractDenied(_contract)) {
```
bingen: What's exactly the point of this? Wouldn't it be the same to add an issue of maximum severity for the contract?

contracts/kill_switch/IssuesRegistry.sol
```diff
@@ -0,0 +1,28 @@
...
+contract IssuesRegistry is IIssuesRegistry, AragonApp {
+    bytes32 constant public SET_ENTRY_SEVERITY_ROLE = keccak256("SET_ENTRY_SEVERITY_ROLE");
+
+    mapping (address => Severity) internal issuesSeverity;
+
+    function initialize() external onlyInit {
+        initialized();
+    }
+
+    function setSeverityFor(address entry, Severity severity) external authP(SET_ENTRY_SEVERITY_ROLE, arr(entry, msg.sender)) {
```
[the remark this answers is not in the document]
facuspagnuolo: I had a brain fart :)

contracts/kill_switch/KillSwitch.sol
facuspagnuolo: I'm thinking the case where a DAO don't have access to submit issues to the registry, then, once again you have the chance for a last call on the DAO's side

facuspagnuolo:
@bingen thanks for your review, I addressed all your comments in the last commit (https://github.com/aragon/aragonOS/pull/518/commits/ffff1702521eebb43f4fdfe5cdf4a27a4cac0723), feel free to take another look. Regarding your question about the idea of overriding the KillSwitch, yes, it's a wip, but I'm probably going to push a separate PR for that.

contracts/kernel/KernelConstants.sol
```diff
@@ -10,10 +10,12 @@ contract KernelAppIds {
     bytes32 internal constant KERNEL_CORE_APP_ID = apmNamehash("kernel");
     bytes32 internal constant KERNEL_DEFAULT_ACL_APP_ID = apmNamehash("acl");
     bytes32 internal constant KERNEL_DEFAULT_VAULT_APP_ID = apmNamehash("vault");
+    bytes32 internal constant KERNEL_DEFAULT_KILL_SWITCH_APP_ID = apmNamehash("killSwitch");
```
izqui: Nit: App names use snake case 

```suggestion
    bytes32 internal constant KERNEL_DEFAULT_KILL_SWITCH_APP_ID = apmNamehash("kill-switch");
```

(will need to update the actual hash as well)

contracts/kill_switch/IssuesRegistry.sol
```diff
@@ -0,0 +1,28 @@
...
+contract IssuesRegistry is IIssuesRegistry, AragonApp {
+    bytes32 constant public SET_ENTRY_SEVERITY_ROLE = keccak256("SET_ENTRY_SEVERITY_ROLE");
+
+    mapping (address => Severity) internal issuesSeverity;
+
+    function initialize() external onlyInit {
+        initialized();
+    }
+
+    function setSeverityFor(address entry, Severity severity) external authP(SET_ENTRY_SEVERITY_ROLE, arr(entry)) {
```
izqui: It would be clearer if `entry` was renamed to `baseContract` or `implementation`

izqui: This will allow to parametrize permissions depending on the severity being set

```suggestion
    function setSeverityFor(address entry, Severity severity) external authP(SET_ENTRY_SEVERITY_ROLE, arr(entry, uint256(severity))) {
```

contracts/kill_switch/KillSwitch.sol
```diff
@@ -0,0 +1,130 @@
...
+     */
+
+    bytes32 constant public SET_DEFAULT_ISSUES_REGISTRY_ROLE = 0xec32b556caaf18ff28362d6b89f3f678177fb74ae2c5c78bfbac6b1dedfa6b43;
+    bytes32 constant public SET_ISSUES_REGISTRY_ROLE = 0xc347b194ad4bc72077d417e05508bb224b4be509950d86cc7756e39a78fb725b;
+    bytes32 constant public SET_CONTRACT_ACTION_ROLE = 0xc7e0b4d70cab2a2679fe330e7c518a6e245cc494b086c284bfeb5f5d03fbe3f6;
+    bytes32 constant public SET_HIGHEST_ALLOWED_SEVERITY_ROLE = 0xca159ccee5d02309b609308bfc70aecedaf2d2023cd19f9c223d8e9875a256ba;
+
+    string constant private ERROR_ISSUES_REGISTRY_NOT_CONTRACT = "KS_ISSUES_REGISTRY_NOT_CONTRACT";
+
+    enum ContractAction { Allow, Check, Deny }
```
izqui: As we chatted offline, not all actions are useful on a per base/implementation contract basis.

With the current implementation, if checks are enabled for an app when it is upgraded the new implementation contract would need to be enabled for kill switch checks. IMO for most uses, it would be better if the `Check` action is applied per `appId`.

Similarly, the `Allow` action makes more sense to be applied for proxy addresses, which allows to bypass the kill switch check for a specific app. Marking a critical app proxy (like the root authority Voting instance in an organization) with `Allow` would be a way to prevent the organization from ever bricking (at the expense of not being protected by the kill switch)

contracts/kernel/Kernel.sol
```diff
@@ -54,6 +56,27 @@ contract Kernel is IKernel, KernelStorage, KernelAppIds, KernelNamespaceConstant
...
+    {
+        // Set and create ACL app
+        initialize(_baseAcl, _permissionsCreator);
+
+        // Set and create KillSwitch app
+        _setApp(KERNEL_APP_BASES_NAMESPACE, KERNEL_DEFAULT_KILL_SWITCH_APP_ID, _baseKillSwitch);
+        IKillSwitch killSwitch = IKillSwitch(newAppProxy(this, KERNEL_DEFAULT_KILL_SWITCH_APP_ID));
+        killSwitch.initialize(_issuesRegistry);
+        _setApp(KERNEL_APP_ADDR_NAMESPACE, KERNEL_DEFAULT_KILL_SWITCH_APP_ID, killSwitch);
+    }
```
sohkai: To avoid coupling the Kernel too much with the KillSwitch, I think this bit of logic could be moved higher up the stack to the `DAOFactory`.

I think this would also make it more explicit that your Kernel can function without a KillSwitch, and that you can install and remove the KillSwitch as wanted.

contracts/kill_switch/IKillSwitch.sol
```diff
@@ -0,0 +1,10 @@
+pragma solidity 0.4.24;
+
+import "./IIssuesRegistry.sol";
+
+
+contract IKillSwitch {
+    function initialize(IIssuesRegistry _issuesRegistry) external;
```
sohkai: Both here and `IIssuesRegistry`, I'd prefer removing `initialize()` from the interface.

It feels like an abstraction leak to have this here; it should only be the initializer / installer of these two apps that needs to know about its initialization protocol (another good reason to move the KillSwitch's initialization into the `DAOFactory`, as that's where both the ACL and KillSwitch are "real" rather than at the Kernel, where they're abstract).

contracts/kill_switch/IIssuesRegistry.sol
```diff
@@ -0,0 +1,16 @@
+pragma solidity 0.4.24;
+
+
+contract IIssuesRegistry {
+    enum Severity { None, Low, Mid, High, Critical }
+
+    event SeveritySet(address indexed entry, Severity severity, address sender);
```
sohkai: May want to index the sender

contracts/kill_switch/IssuesRegistry.sol
```diff
@@ -0,0 +1,28 @@
...
+        issuesSeverity[entry] = severity;
```
sohkai: Should we ensure that the new severity isn't lower than a previous one?

contracts/kill_switch/IssuesRegistry.sol
```diff
@@ -0,0 +1,28 @@
...
+        emit SeveritySet(entry, severity, msg.sender);
+    }
+
+    function isSeverityFor(address entry) public view isInitialized returns (bool) {
```
sohkai: We should come up with a better name for this, perhaps `hasSeverity()` or `isAwareOfSeverity()`

contracts/apps/AragonApp.sol
```diff
@@ -31,6 +32,12 @@ contract AragonApp is AppStorage, Autopetrified, VaultRecoverable, ReentrancyGua
         _;
     }
 
+    modifier killSwitched {
```
sohkai: Wondering if we should directly bundle this into `auth()` or something similar as well. It may be tedious for developers to place this on every external, state-changing call (similar to `isInitialized()`).

We may also want to find a better name, e.g. `killSwitchProtected`.

contracts/kill_switch/IssuesRegistry.sol
facuspagnuolo: The thing is that there is not way to roll it back then, I'm thinking of a false-positive scenario

contracts/apps/AragonApp.sol
facuspagnuolo: I think it may make sense but before coupling them we can keep working on other details and leave that decision for later. We will probably have a better understanding of how the whole kill-switch thing interacts with a DAO once we have finished some of the pending stuff.

OTOH, I like the new name you propose :)

facuspagnuolo:
@izqui @bingen @sohkai thanks for your reviews, I already addressed all your comments, and I really like how this is looking :)

OTOH, I've been exploring many different edge cases about recoverability, and unfortunately my conclusion is that we will need to keep working on further models to have a robust implementation. I filed an [issue](https://github.com/aragon/aragonOS/issues/523) sharing my thoughts on the matter, feel free to comment and leave feedback there. 
However, based on the last changes, I think we can are pretty to close to have a full MVP version of the kill-switch, we just need to decide the MVP version of the recoverability process (please follow-up in the [issue](https://github.com/aragon/aragonOS/issues/523) mentioned above).

contracts/apps/AragonApp.sol
bingen: Yes, makes sense, although I wonder if, given the gas overhead, it could be useful to have that granularity to only apply the killSwitch to very critical functions.
Nevertheless, one could make the argue that if the function has an ACL protection is likely critical. :thinking:

contracts/kill_switch/KillSwitch.sol
```diff
@@ -0,0 +1,142 @@
...
+        if (isInstanceAllowed(_instance)) {
+            return false;
+        }
+
+        // if the base implementation is denied, then deny given call
+        if (isBaseImplementationDenied(_base)) {
+            return true;
+        }
+
+        // if the app severity found is ignored, then allow given call
```
izqui: I would clarify that this is the part of the code where we check against the issues registry, maybe we could rename `isSeverityIgnored` to something more explicit.

facuspagnuolo: I improved the comment there and rename the function to `checkSeverityIssuesFor`, WDYT?

contracts/test/mocks/kill_switch/FailingKillSwitchMock.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity 0.4.24;
+
+import "../../../kill_switch/KillSwitch.sol";
+
+
+contract FailingKillSwitchMock is KillSwitch {
```
sohkai: Nitpick: perhaps we can reword the name and error to be "revert" for clarity ("failing" is ambiguous as to if it doesn't work correctly by returning bad values or by reverting).

test/contracts/kill_switch/issues_registry.test.js
```diff
@@ -0,0 +1,129 @@
...
+      })
+
+      context('when the issues was not fixed yet', () => {
+        it('returns true', async () => {
+          assert.isTrue(await issuesRegistry.hasSeverity(implementation), 'did not expect severity for given implementation')
+        })
+      })
+
+      context('when the issues was already fixed', () => {
+        beforeEach('set medium severity', async () => {
```
sohkai:
```suggestion
        beforeEach('roll back severity', async () => {
```

test/contracts/kill_switch/issues_registry.test.js
```diff
@@ -0,0 +1,129 @@
...
+        await issuesRegistry.setSeverityFor(implementation, SEVERITY.LOW, { from: owner })
+      })
+
+      context('when the issues was not fixed yet', () => {
+        it('returns true', async () => {
+          assert.isTrue(await issuesRegistry.hasSeverity(implementation), 'did not expect severity for given implementation')
+        })
+      })
+
+      context('when the issues was already fixed', () => {
```
sohkai:
```suggestion
      context('when the issues was a false positive', () => {
```

It's unrealistic that the issue would be fixed without updating the base implementation contract (it would have to be an dynamically linked external contract that got changed), so this might be a more realistic scenario 😄

contracts/kill_switch/KillSwitch.sol
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * @dev Note that we are not checking if the appId, base address and instance address are valid and if they correspond
+     *      to each other in order to reduce extra calls. However, since this is only a query method, wrong input
+     *      can only result in invalid output. Internally, this method is used from the Kernel to stop calls if needed,
+     *      and we have several tests to make sure its usage is working as expected.
+     */
+    function shouldDenyCallingApp(bytes32 _appId, address _base, address _instance) external returns (bool) {
+        // if the instance is allowed, then allow given call
+        if (isInstanceAllowed(_instance)) {
```
sohkai: Wonder if we should a check for `_instance == address(this)` here, so that all KillSwitch instances implicitly can't turn themselves off.

contracts/apps/AragonApp.sol
```diff
@@ -57,6 +64,36 @@ contract AragonApp is AppStorage, Autopetrified, VaultRecoverable, ReentrancyGua
         );
     }
 
+    /**
+    * @dev Check whether a call to the current app can be executed or not based on the kill-switch settings
+    * @return Boolean indicating whether the call could be executed or not
+    */
+    function canExecuteCall() public view returns (bool) {
```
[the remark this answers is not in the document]
facuspagnuolo: I like the idea, I'll go with `isCallEnabled`

contracts/kernel/IKernel.sol
```diff
@@ -20,4 +20,5 @@ contract IKernel is IKernelEvents, IVaultRecoverable {
 
     function setApp(bytes32 namespace, bytes32 appId, address app) public;
     function getApp(bytes32 namespace, bytes32 appId) public view returns (address);
+    function shouldDenyCallingContract(bytes32 appId, address _instance) public returns (bool);
```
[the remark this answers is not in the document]
facuspagnuolo: Sure, I like the idea :)

contracts/factory/DAOFactory.sol
```diff
@@ -1,33 +1,46 @@
...
 
 
 contract DAOFactory {
+    string private constant ERROR_MISSING_BASE_KILL_SWITCH = "DF_MISSING_BASE_KILL_SWITCH";
+
     IKernel public baseKernel;
     IACL public baseACL;
-    EVMScriptRegistryFactory public regFactory;
+    KillSwitch public baseKillSwitch;
+    EVMScriptRegistryFactory public scriptsRegistryFactory;
```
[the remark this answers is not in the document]
facuspagnuolo: I see, I'll roll it back

contracts/kill_switch/KillSwitch.sol
```diff
@@ -0,0 +1,142 @@
...
+
+    function isInstanceAllowed(address _instance) public view returns (bool) {
+        return allowedInstances[_instance];
+    }
+
+    function isBaseImplementationDenied(address _base) public view returns (bool) {
+        return deniedBaseImplementations[_base];
+    }
+
+    function checkSeverityIssuesFor(bytes32 _appId, address _base) public view returns (bool) {
```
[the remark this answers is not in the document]
facuspagnuolo: I like `hasExceededAllowedSeverity` 👍

contracts/kill_switch/KillSwitch.sol
```diff
@@ -0,0 +1,142 @@
...
+    string constant private ERROR_ISSUES_REGISTRY_NOT_CONTRACT = "KS_ISSUES_REGISTRY_NOT_CONTRACT";
+
+    struct IssuesSettings {
+        IIssuesRegistry issuesRegistry;
+        IIssuesRegistry.Severity highestAllowedSeverity;
+    }
+
+    IIssuesRegistry public defaultIssuesRegistry;
+    mapping (address => bool) internal allowedInstances;
+    mapping (address => bool) internal deniedBaseImplementations;
```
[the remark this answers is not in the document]
facuspagnuolo: We can, would you change the interface as well? I mean, `isInstanceAllowed` to `isInstanceWhitelisted` for example?

contracts/kill_switch/KillSwitch.sol
facuspagnuolo: Yea, I was wondering that too, but I finally decided to whitelist the kill switch instance from the factory instead. However, we could change this function as well, WDYT?

contracts/kill_switch/KillSwitch.sol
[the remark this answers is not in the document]
facuspagnuolo: Nvm, I already changed them. I think it sounds more robust to refer them as whitelist/blacklist instead 👍

contracts/apps/AragonApp.sol
```diff
@@ -57,6 +64,36 @@ contract AragonApp is AppStorage, Autopetrified, VaultRecoverable, ReentrancyGua
...
+        IKernel _kernel = kernel();
+        bytes4 selector = _kernel.shouldDenyCallingContract.selector;
+        bytes memory callData = abi.encodeWithSelector(selector, appId(), address(this));
+        bool success = address(_kernel).call(callData);
+
+        // if the call to `kernel.shouldDenyCallingApp` reverts (using an old Kernel) we consider that
```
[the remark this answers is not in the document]
facuspagnuolo: Let's add it to the `canPerform` method 👍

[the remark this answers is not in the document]
facuspagnuolo: Another thing to have in mind is that we will be hiding too much information behind the  `APP_AUTH_FAILED` error message, it could be somehow misleading.

test/contracts/kill_switch/kill_switch.test.js
```diff
@@ -0,0 +1,963 @@
...
+    SET_DEFAULT_ISSUES_REGISTRY_ROLE = await killSwitchBase.SET_DEFAULT_ISSUES_REGISTRY_ROLE()
+    SET_ISSUES_REGISTRY_ROLE = await killSwitchBase.SET_ISSUES_REGISTRY_ROLE()
+    SET_ALLOWED_INSTANCES_ROLE = await killSwitchBase.SET_ALLOWED_INSTANCES_ROLE()
+    SET_DENIED_BASE_IMPLS_ROLE = await killSwitchBase.SET_DENIED_BASE_IMPLS_ROLE()
+    SET_HIGHEST_ALLOWED_SEVERITY_ROLE = await killSwitchBase.SET_HIGHEST_ALLOWED_SEVERITY_ROLE()
+  })
+
+  context('when the kernel version does not support kill-switch logic', async () => {
+    before('create DAO factory', async () => {
+      daoFactory = await DAOFactory.new(kernelWithoutKillSwitchBase.address, aclBase.address, ZERO_ADDRESS, registryFactory.address)
```
sohkai: What do you think about no using the `DAOFactory` in these tests, but installing a `KillSwitch` manually, so that it's really only unit testing the `KillSwitch` (then the `DAOFactory` tests can act as more "integration" type tests).

contracts/apps/AragonApp.sol
```diff
@@ -57,6 +64,36 @@ contract AragonApp is AppStorage, Autopetrified, VaultRecoverable, ReentrancyGua
...
+        // there is no kill switch and the call can be executed be allowed to continue
+        if (!success) return true;
+
+        // if not, first ensure the returned value is 32 bytes length
+        uint256 _outputLength;
+        assembly { _outputLength := returndatasize }
+        require(_outputLength == 32, ERROR_UNEXPECTED_KERNEL_RESPONSE);
```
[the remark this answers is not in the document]
facuspagnuolo: At least for now, we are using this check within a `require` statement, i.e. to raise or not an error. Thus, returning false or reverting here will result in the same effect. And I think that between having more context to know why a call reverted and not, I prefer having it. Do you have in mind some particular scenario where it could be useful swallowing that scenario?

test/contracts/kill_switch/kill_switch.test.js
```diff
@@ -0,0 +1,963 @@
...
+    APP_MANAGER_ROLE = await kernelBase.APP_MANAGER_ROLE()
+    SET_SEVERITY_ROLE = await issuesRegistryBase.SET_SEVERITY_ROLE()
+    SET_DEFAULT_ISSUES_REGISTRY_ROLE = await killSwitchBase.SET_DEFAULT_ISSUES_REGISTRY_ROLE()
+    SET_ISSUES_REGISTRY_ROLE = await killSwitchBase.SET_ISSUES_REGISTRY_ROLE()
+    SET_ALLOWED_INSTANCES_ROLE = await killSwitchBase.SET_ALLOWED_INSTANCES_ROLE()
+    SET_DENIED_BASE_IMPLS_ROLE = await killSwitchBase.SET_DENIED_BASE_IMPLS_ROLE()
+    SET_HIGHEST_ALLOWED_SEVERITY_ROLE = await killSwitchBase.SET_HIGHEST_ALLOWED_SEVERITY_ROLE()
+  })
+
+  context('when the kernel version does not support kill-switch logic', async () => {
```
sohkai: Doesn't this set of tests check a `KillSwitch`-compatible Kernel with no `KillSwitch` installed? I would say `KernelWithoutKillSwitchMock` is the one that best mimics the behaviour of old Kernels.

test/contracts/kill_switch/kill_switch.test.js
```diff
@@ -0,0 +1,963 @@
...
+    })
+
+    it('executes the call', async () => {
+      await app.write(10, { from: owner })
+      assert.equal(await app.read(), 10)
+    })
+  })
+
+  context('when the kernel version does support kill-switch logic', async () => {
+    const SAMPLE_APP_ID = '0x1236000000000000000000000000000000000000000000000000000000000000'
```
sohkai: Perhaps this constant could be lifted up, I notice we use a similar one in https://github.com/aragon/aragonOS/pull/518/files#diff-20fa58b7946252b898fd2dc455170cf6R65.

test/contracts/kill_switch/kill_switch.test.js
facuspagnuolo: Maybe the context message is misleading, but it is actually testing an "old kernel" version, that's why I mention that it "does not support kill-switch logic". Actually right below is a `beforeEach` statement deploying an "old" kernel using `KernelWithoutKillSwitchMock` (see [here](https://github.com/aragon/aragonOS/pull/518/files/ce7a29bf9d359abeab95c9a6b9075fe90ed8722d#diff-20fa58b7946252b898fd2dc455170cf6R56))

test/contracts/kill_switch/kill_switch.test.js
```diff
@@ -0,0 +1,963 @@
...
+
+                itExecutesTheCall()
+              })
+
+              context('when the base implementation is denied', () => {
+                beforeEach('deny base implementation', async () => {
+                  await killSwitch.setDeniedBaseImplementation(appBase.address, true, { from: owner })
+                })
+
+                itExecutesTheCall()
```
sohkai: Could we add a comment that allowing the instance has higher precedence than denying the base implementation?

test/contracts/kill_switch/kill_switch.test.js
```diff
@@ -0,0 +1,963 @@
...
+              })
+            })
+
+            context('when the instance being called is not marked as allowed', () => {
+              beforeEach('dot not allow instance', async () => {
```
sohkai:
```suggestion
              beforeEach('do not allow instance', async () => {
```

test/contracts/kill_switch/kill_switch.test.js
facuspagnuolo: We could but I think we are covering more contracts and avoiding code duplication if we use it. However, if you prefer to follow that pattern I can refactor them.

test/contracts/kill_switch/kill_switch.test.js
```diff
@@ -0,0 +1,963 @@
...
+
+                itExecutesTheCall()
+              })
+
+              context('when the base implementation is denied', () => {
+                beforeEach('deny base implementation', async () => {
+                  await killSwitch.setDeniedBaseImplementation(appBase.address, true, { from: owner })
+                })
+
+                itExecutesTheCall()
```
sohkai: Shouldn't this revert?

contracts/factory/DAOFactory.sol
```diff
@@ -1,33 +1,46 @@
...
-        if (address(_regFactory) != address(0)) {
-            regFactory = _regFactory;
+        if (address(_scriptsRegistryFactory) != address(0)) {
+            scriptsRegistryFactory = _scriptsRegistryFactory;
+        }
+        if (address(_baseKillSwitch) != address(0)) {
+            baseKillSwitch = _baseKillSwitch;
         }
 
         baseKernel = _baseKernel;
```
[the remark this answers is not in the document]
facuspagnuolo: The problem is that you won't notice any further changes in the contracts that implements those IDs in the future, we could cache it for a single transaction, but caching it "forever" seems a lot to me. Note that caching it for a single transaction will result in a lot of code duplication, it is the same trade-off we will always face, code simplicity vs. gas optimization. In this case I opted for the first one since it is a single call per DAO, but I'll try to optimize its costs as much as possible.

test/contracts/kill_switch/kill_switch.test.js
facuspagnuolo: Note that the `context` for this test is `"when the kernel is initialized with a failing kill-switch implementation"`. We decided that if the kill-switch was failing then we will allow calling the app being queried, it would be the same case as an "old-version" kernel

contracts/factory/DAOFactory.sol
```diff
@@ -40,38 +53,114 @@ contract DAOFactory {
...
+    function _transferCreatePermissionsRole(Kernel _dao, ACL _acl, address _from, address _to) internal {
+        bytes32 createPermissionsRole = _acl.CREATE_PERMISSIONS_ROLE();
+        _acl.revokePermission(_from, _acl, createPermissionsRole);
+        if (_from != address(this)) {
+            _acl.revokePermission(address(this), _acl, createPermissionsRole);
+        }
+
+        _acl.grantPermission(_to, _acl, createPermissionsRole);
+        _acl.setPermissionManager(_to, _acl, createPermissionsRole);
     }
```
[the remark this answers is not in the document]
facuspagnuolo: Actually we are granting that permission only for the kill-switch case.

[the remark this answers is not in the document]
facuspagnuolo: As said above, I think we can optimize it a bit but contexts are different, we are granting those permissions to different entities

facuspagnuolo:
Thanks for the review @sohkai! I already addressed/answered all your comments. Feel free to take another look if you want.

test/contracts/kill_switch/kill_switch.test.js
sohkai: :+1: Ahh yes, missed that this was under the "failing kill-switch" context!

test/contracts/kill_switch/kill_switch.test.js
```diff
@@ -0,0 +1,963 @@
...
+
+            context('when the highest allowed severity is greater than the reported bug severity', () => {
+              beforeEach('set highest allowed severity', async () => {
+                await killSwitch.setHighestAllowedSeverity(SAMPLE_APP_ID, SEVERITY.CRITICAL, { from: owner })
+              })
+
+              it('returns true', async () => {
+                assert.isTrue(await killSwitch.checkSeverityIssuesFor(SAMPLE_APP_ID, appBase.address))
+              })
+            })
```
sohkai: Perhaps we can also write a test or two here that checks that `checkSeverityIssuesFor()` reflects the latest issue registry set if the specific app gets set to use a different issues registry (or the `KillSwitch` sets a different default registry).

contracts/apps/AragonApp.sol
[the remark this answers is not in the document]
sohkai: Yeah, the thinking behind putting it inside `Kernel.hasPermission()` was to streamline the API and reduce a bit of the gas overhead, as I imagine it would be _very_ common to put the `auth` and `killswitchProtected` modifiers together for most apps.

contracts/apps/AragonApp.sol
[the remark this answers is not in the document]
sohkai: If someone were to call this function directly, outside of the modifier (e.g. the app itself or another external contract calling this function to see if it was killswitched or not).

One example is with the Token Manager's [`canForward()`](https://github.com/aragon/aragon-apps/blob/master/apps/token-manager/contracts/TokenManager.sol#L269), where if we don't bake-in the killswitch check at the `EVMScriptRunner` level, we'd have to do a query to `canExecuteCall()` there.

I'd also like to make `canPerform()` and `canExecuteCall()` as consistent as possible, so users interact with them using the same error handling model. `canPerform()` masks incorrect oracle param responses with `false`, rather than reverting.

contracts/factory/DAOFactory.sol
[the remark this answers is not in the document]
sohkai: > you won't notice any further changes in the contracts that implement those IDs

Yes, but in the case of `DAOFactory`, there's no way to change the base implementations we get of the `Kernel`, `ACL`, and `KillSwitch`. Outside of funny business with CREATE2, these IDs shouldn't change.

If we want to handle the possible CREATE2 scenario where these bases are replaced, then let's keep the code simpler and just query them when we need them 👍

contracts/kill_switch/KillSwitch.sol
sohkai: I'm definitely thinking making it implicit here sounds like a better abstraction; a `KillSwitch` turning itself off could result in some weird nasty business (although again, right now nothing in `KillSwitch` is `killSwitchProtected` so nothing would happen, which is also quite awkward).

test/contracts/kill_switch/kill_switch.test.js
sohkai: 🙏 I might've gotten confused with the context below, this test looks good!

test/contracts/kill_switch/kill_switch.test.js
sohkai: I'm wondering if it might be worth splitting the tests into `kill-switch/kill_switch.js`, which are mostly unit tests for the exposed `KillSwitch` functionality, and `kernel/kernel_kill_switch.js`, which are then the full suite of integration tests that interact with the various `Kernel`s / mocks.

contracts/factory/DAOFactory.sol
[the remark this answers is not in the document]
facuspagnuolo: Anyway, I refactored it in a way we query roles only once per new DAO

contracts/apps/AragonApp.sol
```diff
@@ -47,7 +41,7 @@ contract AragonApp is AppStorage, Autopetrified, VaultRecoverable, ReentrancyGua
     *         Always returns false if the app hasn't been initialized yet.
     */
     function canPerform(address _sender, bytes32 _role, uint256[] _params) public view returns (bool) {
-        if (!hasInitialized()) {
+        if (!isCallEnabled()) {
             return false;
         }
```
sohkai: We can move the `address(linkedKernel) == address(0)` check to be above this, as `isCallEnabled()` will always fail if the kernel doesn't exist. It may also save a _tiny_ bit of gas to inline the two conditions with an `||` 🤷‍♂

contracts/kernel/Kernel.sol
```diff
@@ -164,11 +164,13 @@ contract Kernel is IKernel, KernelStorage, KernelAppIds, KernelNamespaceConstant
     }
 
     /**
-    * @dev Tells whether a call to an instance of an app should be denied or not based on the kill-switch settings
+    * @dev Tells whether a call to an instance of an app should be denied or not based on the kill-switch settings.
+    *      Note that we don't need to perform an initialization check since having or not a kill-switch installed
+    *      implicitly means that.
```
sohkai: English nitpick: would rephrase to

```
Initialization check is implicitly provided by the KillSwitch's existence, as apps can only be installed after initialization
```

contracts/kill-switch/KillSwitch.sol
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    function isBaseImplementationBlacklisted(address _base) public view returns (bool) {
+        return blacklistedBaseImplementations[_base];
+    }
+
+    function hasExceededAllowedSeverity(bytes32 _appId, address _base) public view returns (bool) {
+        IIssuesRegistry.Severity severityFound = getIssuesRegistry(_appId).getSeverityFor(_base);
+        IIssuesRegistry.Severity highestAllowedSeverity = getHighestAllowedSeverity(_appId);
+        return highestAllowedSeverity >= severityFound;
```
sohkai: Nitpick: this should probably be `highestAllowedSeverity < severityFound` (and `shouldDenyCallingApp()` should return true if this function is true), to stay true to this function's name (**exceeding** allowed severity).

Alternatively we could rename this to `isWithinAllowedSeverity()` to keep the logic the same.

test/contracts/kill-switch/kill_switch.js
```diff
@@ -0,0 +1,972 @@
...
+          await killSwitch.setHighestAllowedSeverity(SAMPLE_APP_ID, SEVERITY.MID, { from: owner })
+          await defaultIssuesRegistry.setSeverityFor(appBase.address, SEVERITY.LOW, { from: securityPartner })
+        })
+
+        it('kill switch should overload ~27k of gas to a function', skipCoverage(async () => {
+          const { receipt: { cumulativeGasUsed: gasUsedWithKillSwitch } } = await app.write(10, { from: owner })
+          const { receipt: { cumulativeGasUsed: gasUsedWithoutKillSwitch } } = await app.writeWithoutKillSwitch(10, { from: owner })
+
+          const killSwitchCost = gasUsedWithKillSwitch - gasUsedWithoutKillSwitch
+          assert(killSwitchCost <= 27000, 'kill switch should overload ~27k of gas')
```
sohkai: Nitpick: wondering if there was a reason to not use [`assert.isAtMost()`](https://www.chaijs.com/api/assert/#method_isatmost) here, so the tests would log the actual gas used on failure?

contracts/kill-switch/KillSwitch.sol
facuspagnuolo: Good catch, I completely forgot that after renaming it :)

facuspagnuolo:
The coverage task is failing because there are some tests that fail but only when measuring coverage. I've trying to debug it locally but it was completely cumbersome and I couldn't get to know what was going on. After that, I tried running `solidity-coverage` on using `ganache-cli` instead of `testrpc-sc` using [frangio's fork](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/v2.3.0-rc.3/package.json#L57) and it passed. Definitely `solidity-coverage` is an outdated tool that has already been giving us several headaches 😕

PR #525, opened 2019-05-08
Tests: Refactors and enhancements

facuspagnuolo:
This PR carries some test suite enhancements:
- Use new `assertRevert` syntax
- Provide better way to assert events
- Remove unused variables
- Improve imports


test/helpers/assertEvent.js
```diff
@@ -1,5 +1,35 @@
-module.exports = (receipt, eventName, instances = 1) => {
-    const events = receipt.logs.filter(x => x.event == eventName)
-    assert.equal(events.length, instances, `'${eventName}' event should have been fired ${instances} times`)
-    return events
+module.exports = web3 => {
+    const getEvents = ({ logs = [] }, event) => logs.filter(l => l.event === event)
+    const getEventAt = (receipt, event, index = 0) => getEvents(receipt, event)[index]
+    const getEventArgument = (receipt, event, arg, index = 0) => getEventAt(receipt, event, index).args[arg]
+    const getNewProxyAddress = (receipt) => getEventArgument(receipt, 'NewAppProxy', 'proxy')
```
sohkai: What do you think about moving these getters to a separate file, perhaps `events.js`? There's a few now, so it might be nice to separate the utilities from the custom assertions.

PR #526, opened 2019-05-09
[PoC] Meta Txs: Measure gas ovearload

facuspagnuolo:
This PR is simply to show different approaches I've been exploring to support meta transactions in order to allow any member of a DAO to interact with its apps without requiring them to have funds. The only concept shared among all the alternatives described in this PR is the idea of having a shared trusted off-chain service in charge of submitting the transactions on-chain that will get refunded by the DAOs for every transaction it relays to make sure it doesn't run out of funds. 

That said, the different alternatives explored are the possible combinations between the following patterns:
1. Having an on-chain relayer in charge of signature validation, nonce validation, and refunds, that will act as a middleman between the off-chain service and the target app, vs. integrating all these responsibilities within the `AragonApp` itself.
2. Parameterizing the transaction signer to the functions of the target app, vs. using the [volatile storage approach](https://github.com/aragon/aragonOS/pull/442) to query it when needed without changing the app interface

The following table shows the gas overloads to a single transaction for the combinations mentioned above:

|                           | On-chain Relayer | App Relayer | 
| --------------- |------------------|------------- |
| Volatile              |            114k           |        84k         |   
| Parameterized  |             94k           |        75k         |

The measures showed above where taken while storing all the nonces used by a user, if can manage storing only the last used nonce we have get the following numbers:

|                           | On-chain Relayer | App Relayer | 
| --------------- |------------------|------------- |
| Volatile              |             98k           |        83k         |   
| Parameterized  |             78k           |        74k         |

The following table describes the different costs of the tasks that are inherent to meta transactions in general:

|                                           | Gas cost | 
| -------------------------|----------|
| Nonce validation              |     0.4k    |
| Signature validation         |     6.1k    | 
| Off-chain service refund |     7.5k    | 
| Off-chain service auth    |    18.3k   |


contracts/relayer/BaseRelayer.sol
```diff
@@ -0,0 +1,61 @@
...
+    string private constant ERROR_NONCE_ALREADY_USED = "RELAYER_NONCE_ALREADY_USED";
+    string private constant ERROR_INVALID_SENDER_SIGNATURE = "RELAYER_INVALID_SENDER_SIGNATURE";
+
+    event FundsReceived(address indexed sender, uint256 amount);
+    event TransactionRelayed(address indexed from, address indexed to, uint256 nonce, bytes calldata);
+
+    modifier refundGas() {
+        uint256 startGas = gasleft();
+        _;
+        uint256 refund = EXTERNAL_TX_COST + startGas - gasleft();
```
izqui: Two comments about `EXTERNAL_TX_COST`:

- If the sender is a contract, they would be able to call into this contract multiple times in one transaction, so we would be refunding more than necessary.
- Assuming that the call is performed by an EOA, the cost will always be higher than 21k gas as the contract needs to read the calldata and find the function that is being called before the modifier sets the `startGas`.

contracts/relayer/BaseRelayer.sol
```diff
@@ -0,0 +1,61 @@
...
+        require(!isNonceUsed(from, nonce), ERROR_NONCE_ALREADY_USED);
+        require(isValidSignature(from, messageHash(calldata, nonce), signature), ERROR_INVALID_SENDER_SIGNATURE);
+    }
+
+    function isValidSignature(address sender, bytes32 hash, bytes signature) internal pure returns (bool) {
+        address signer = hash.toEthSignedMessageHash().recover(signature);
+        return sender == signer;
+    }
+
+    function messageHash(bytes calldata, uint256 nonce) internal pure returns (bytes32) {
```
izqui: Probably want to include the target contract here as well, otherwise calls could be relayed to any app.

contracts/relayer/BaseRelayer.sol
```diff
@@ -0,0 +1,61 @@
...
+
+import "../lib/sig/ECDSA.sol";
+import "../apps/AragonApp.sol";
+import "../common/DepositableStorage.sol";
+
+
+contract BaseRelayer is AragonApp, DepositableStorage {
+    using ECDSA for bytes32;
+
+    bytes32 public constant OFF_CHAIN_RELAYER_SERVICE_ROLE = keccak256("OFF_CHAIN_RELAYER_SERVICE_ROLE");
```
izqui: What's the rationale for having a role for this? The auth check is definitely adding a decent amount of gas overhead.

contracts/relayer/BaseRelayer.sol
```diff
@@ -0,0 +1,61 @@
...
+        require(msg.sender.send(refund), ERROR_GAS_REFUND_FAIL);
+    }
+
+    function () external payable {
+        emit FundsReceived(msg.sender, msg.value);
+    }
+
+    function initialize() public onlyInit {
+        initialized();
+        setDepositable(true);
```
izqui: Note that by making it depositable, the call can be intercepted by the proxy depending on the gas the call has left: https://github.com/aragon/aragonOS/blob/dev/contracts/common/DepositableDelegateProxy.sol#L12-L15

To ensure this contract's fallback is called when sending ETH (besides ETH being deposited with `SELFDESTRUCT`) is by not making it depositable.

contracts/relayer/RelayerAragonAppWithParameterizedSender.sol
```diff
@@ -0,0 +1,29 @@
...
+        _;
+    }
+
+    modifier relayedAuthP(address _sender, bytes32 _role, uint256[] _params) {
+        require(canPerform(_sender, _role, _params), ERROR_AUTH_FAILED);
+        _;
+    }
+
+    function exec(address from, uint256 nonce, bytes calldata, bytes signature) external refundGas auth(OFF_CHAIN_RELAYER_SERVICE_ROLE) {
+        assertValidTransaction(from, nonce, calldata, signature);
```
izqui: The issue that I see with this approach is that ensuring that the sender in the calldata is the signer of the message would require inspecting the calldata to extract the sender parameter and checking that it equals to `from`. 

With the current implementation, any address can sign a valid message that would allow them to impersonate someone else.

contracts/relayer/BaseRelayer.sol
facuspagnuolo: Yea, I just added it because it feels more safe this way, but it may not be dangerous to accept relayed calls from any off-chain service. The only possible attack we could face is miners as relayers where they could set a high gas price for example. 

An optimization to using the ACL could be to keep a whitelisted list of off-chain services, and use the ACL to write/remove addresses from that list instead.

contracts/relayer/BaseRelayer.sol
facuspagnuolo: Sure, I just wanted to provide the closest  implementation to have an idea of its costs. We will need to do the maths and write some tests as well.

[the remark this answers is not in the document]
facuspagnuolo: Sure, I just wanted to provide the closest implementation to have an idea of its costs. We will need to do the maths and write some tests as well.

contracts/relayer/RelayerAragonAppWithParameterizedSender.sol
facuspagnuolo: Yea, being honest, this approach doesn't look good at all to me.

contracts/relayer/RelayedAragonAppWithParameterizedSender.sol
```diff
@@ -0,0 +1,24 @@
+pragma solidity ^0.4.24;
+
+import "../apps/AragonApp.sol";
+
+
+contract RelayedAragonAppWithParameterizedSender is AragonApp {
```
bingen: Is this supposed to be the 4th case in the description? (The "Parameterized App Relayer") Isn't it missing the `exec` bits?

contracts/relayer/RelayerAragonAppWithParameterizedSender.sol
```diff
@@ -0,0 +1,29 @@
...
+
+        setUsedNonce(from, nonce, true);
+        bool success = address(this).call(calldata);
```
bingen: Are this and the next one supposed to be the cases in the first column of the description? The "on-chain relayers"? In that case, shouldn't they call to an external app, as they are just a "middleman", instead to itself?

contracts/relayer/RelayerAragonAppWithVolatileSender.sol
```diff
@@ -0,0 +1,29 @@
...
+
+
+contract RelayerAragonAppWithVolatileSender is BaseRelayer {
+    function exec(address from, uint256 nonce, bytes calldata, bytes signature) external refundGas auth(OFF_CHAIN_RELAYER_SERVICE_ROLE) {
+        assertValidTransaction(from, nonce, calldata, signature);
+
+        setVolatileStorageSender(from);
+        setUsedNonce(from, nonce, true);
+
+        bool success = address(this).call(calldata);
```
bingen: Same comment as above about the external app.

contracts/relayer/BaseRelayer.sol
facuspagnuolo: I'm actually working on another approach cause this one has a ton of edge cases and it actually costs a lot of gas to calculate how much gas we need to refund

contracts/relayer/RelayerAragonAppWithParameterizedSender.sol
facuspagnuolo: No, this example would be the app relayer, i.e. an AragonApp including relayer logic. It is forwarding the call  to itself to enter the target function actually.

contracts/relayer/RelayerAragonAppWithVolatileSender.sol
facuspagnuolo: Answer above

contracts/relayer/RelayedAragonAppWithParameterizedSender.sol
facuspagnuolo: No, this is a relayed app, i.e. an app that is being relayed by an on-chain relayer.

facuspagnuolo:
Thanks for your review @bingen! Although, I think it is more interesting to review directly the last approach we decided to follow :)

facuspagnuolo:
@izqui I decided to drop the idea of calculating the amount of gas to be refunded to the off-chain service since it was actually taking a lot of gas to perform such calculation (I reached an overload of 100k gas for some edge cases). That said, I came up with the idea of delegating the decision of calculating the gas refund to the DAOs' members. Note that off-chain services can tell whether the submitted amount of gas actually covers the costs of a transaction before relaying it. Therefore, DAOs' members will be incentivized to estimate gas values correctly. 

To achieve this, I propose using a monthly quota on the relayer side to limit the amount of txs users can relay. And ofc, this triggers some new questions we should decide on:
- Should we allow to change quotas amount?
- Should we allow to change quotas based period of time?
- Should we have different quotas per address or is it okay to have the same one for all members?
- Should we catch reverting relayed calls to ensure we always refund gas costs or should the off-chain service check if the tx will revert before relaying it?

izqui:
> Therefore, DAOs' members will be incentivized to estimate gas values correctly.

This could be an issue if the 'service provider' can relay their own transactions (and in the current implementation it seems like any account can use the relayer), as their incentive would be to use a high gas limit/price.

> Should we allow to change quotas amount and how?

I think we should, as the amount of activity that happens in the organization will change over time and the gas price can fluctuate.

> Should we allow to change quotas based period of time and how?

I don't think this is really important if the amount can be changed.

> Should we have different quotas per address or is it okay to have the same one for all members?

I think that having a unique quota amount for everyone should be fine for a first version.

> Should we catch reverting relayed calls to ensure we always refund gas costs or should the off-chain service check if the tx will revert before relaying it?

I'd say the off-chain service should make sure that the transaction won't revert.

facuspagnuolo:
@izqui I already addressed your comments, feel free to take another look :)
I'll publish all the off-chain service stuff on a separate PR

contracts/relayer/Relayer.sol
```diff
@@ -0,0 +1,349 @@
...
+    string private constant ERROR_GAS_QUOTA_EXCEEDED = "RELAYER_GAS_QUOTA_EXCEEDED";
+    string private constant ERROR_SENDER_NOT_ALLOWED = "RELAYER_SENDER_NOT_ALLOWED";
+    string private constant ERROR_NONCE_ALREADY_USED = "RELAYER_NONCE_ALREADY_USED";
+    string private constant ERROR_SERVICE_NOT_ALLOWED = "RELAYER_SERVICE_NOT_ALLOWED";
+    string private constant ERROR_INVALID_SENDER_SIGNATURE = "RELAYER_INVALID_SENDER_SIGNATURE";
+
+    // Constant values used to identify the domain for the current app following EIP 712 spec
+    string private constant EIP_712_DOMAIN_NAME = "Aragon Relayer";
+    string private constant EIP_712_DOMAIN_VERSION = "1";
+    uint256 private constant EIP_712_DOMAIN_CHAIN_ID = 1;
```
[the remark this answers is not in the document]
facuspagnuolo: Sure, we can do that :)

contracts/relayer/Relayer.sol
```diff
@@ -0,0 +1,349 @@
...
+    event MonthlyRefundQuotaSet(address indexed who, uint256 previousQuota, uint256 newQuota);
+
+    // Timestamp to start counting monthly refunds quotas for each member
+    uint256 internal startDate;
+
+    // Monthly refunds quota in ETH for each member
+    uint256 internal monthlyRefundQuota;
+
+    // Mapping that indicates whether a given address is allowed to use the relay service
+    mapping (address => bool) internal allowedSenders;
```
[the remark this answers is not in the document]
facuspagnuolo: Initially we did, but given that using the `ACL` increases gas costs in `~15k`, we decided to use it but in a more "async" way. As you can see, we are using it to edit the whitelist of allowed senders, but avoiding it when we need to check if someone is in included in the list.

contracts/relayer/Relayer.sol
```diff
@@ -0,0 +1,349 @@
...
+
+    // Mapping that indicates whether a given address is allowed as off-chain service to relay transactions
+    mapping (address => bool) internal allowedServices;
```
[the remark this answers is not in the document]
facuspagnuolo: Explained above

contracts/relayer/Relayer.sol
```diff
@@ -0,0 +1,349 @@
...
+
+    // Mapping from members to nonce numbers that indicates the last nonce used by each member
+    mapping (address => uint256) internal lastUsedNonce;
+
+    // Mapping from members to monthly refunds that indicates the refunds requested per member per month
+    mapping (address => mapping (uint256 => uint256)) internal monthlyRefunds;
```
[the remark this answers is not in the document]
facuspagnuolo: Sure, I can give it a try

contracts/relayer/Relayer.sol
```diff
@@ -0,0 +1,349 @@
...
+        allowedSenders[sender] = false;
+        emit SenderDisallowed(sender);
+    }
+
+    /**
+     * @notice Set new monthly refunds quota per address of `@tokenAmount(newQuota, 0x00)`.
+     * @param newQuota New monthly refunds quota in ETH for each allowed member
+     */
+    function setMonthlyRefundQuota(uint256 newQuota) external authP(SET_MONTHLY_REFUND_QUOTA_ROLE, arr(newQuota)) {
+        emit MonthlyRefundQuotaSet(msg.sender, monthlyRefundQuota, newQuota);
```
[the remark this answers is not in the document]
facuspagnuolo: Just because I'm editing the monthly refund quota in the next line, otherwise I'll need to cache it first

contracts/relayer/Relayer.sol
```diff
@@ -0,0 +1,349 @@
...
+        return monthRefunds.add(amount) <= monthlyRefundQuota;
+    }
+
+    function _isValidSignature(address sender, address to, uint256 nonce, bytes data, uint256 gasRefund, uint256 gasPrice, bytes signature)
+        internal
+        view
+        returns (bool)
+    {
+        bytes32 messageHash = _messageHash(to, nonce, data, gasRefund, gasPrice);
+        address signer = messageHash.recover(signature);
```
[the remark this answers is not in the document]
facuspagnuolo: No, that's the idea of supporting [EIP 712](https://eips.ethereum.org/EIPS/eip-712) 😉

contracts/relayer/Relayer.sol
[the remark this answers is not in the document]
bingen: Oh, of course! Haha. :sweat_smile:

contracts/relayer/Relayer.sol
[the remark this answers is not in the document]
bingen: I thought it may be related to it, but didn't read all the details of that EIP, and I was confused by the fact that `toEthSignedMessageHash` is still in ECDSA lib. Do we need it there then?

contracts/relayer/Relayer.sol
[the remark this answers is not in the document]
bingen: Ok, then I think this works for services, as I don't expect having so many, but imagine Aragon wants to subsidize some voting (or whatever) to all token holders. According to etherscan, that would be like ~21k holders, i.e., allowed senders. Adding them to that mapping would be quite an amount of gas (and there will probably be bigger orgs with this problem), besides we should monitor ANT transactions to edit allowances.
If we used such an `IGroup` interface, we could have Token Manager implementing it, or if we don't want to update it, just have a wrapper small contract implementing which would get the balance and return true if `> 0`.

contracts/relayer/Relayer.sol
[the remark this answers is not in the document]
facuspagnuolo: Not really, it is part of the library I picked from OpenZeppelin

contracts/relayer/Relayer.sol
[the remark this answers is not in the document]
facuspagnuolo: In that case it would be the same costs even if we use the `ACL` right? I mean, if we want to control each member of a DAO we will need to grant permissions for each of them right?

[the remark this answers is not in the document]
facuspagnuolo: Yea, having groups would be amazing :)

PR #527, opened 2019-05-17
Meta Txs: Adding support for meta transactions in aragon apps (Part 2)

facuspagnuolo:
*Continuation of https://github.com/aragon/aragonOS/pull/526*

This PR aims to provide the required JS libraries to implement an meta-transactions server that will act as the off-chain part of the relayer for the model described in the previous PR.


package.json
```diff
@@ -37,7 +37,7 @@
     "eth-ens-namehash": "^2.0.8",
     "eth-gas-reporter": "^0.1.1",
     "ethereumjs-abi": "^0.6.5",
-    "ganache-cli": "^6.4.2",
+    "ganache-cli": "~6.2.0",
```
sohkai: Interesting, was the latest version failing due to the gas estimation?

lib/relayer/GasPriceOracle.js
```diff
@@ -0,0 +1,25 @@
...
+module.exports = {
+  async fetch(networkId) {
+    if (MAINNET_ID === networkId) return this._fetchMainnetGasPrice()
+    if (TESTNET_IDS.includes(networkId)) return DEFAULT_TESTNET_GAS_PRICE
+    return DEFAULT_DEVNET_GAS_PRICE
+  },
+
+  async _fetchMainnetGasPrice() {
+    try {
+      const axios = require('axios')
```
sohkai: Might as well lift this up as a top-level import

lib/relayer/RelayTransactionSigner.js
```diff
@@ -0,0 +1,118 @@
...
+  Transaction: [
+    { name: 'to', type: 'address' },
+    { name: 'nonce', type: 'uint256' },
+    { name: 'data', type: 'bytes' },
+    { name: 'gasRefund', type: 'uint256' },
+    { name: 'gasPrice', type: 'uint256' }
+  ],
+}
+
+module.exports = web3 => class RelayTransactionSigner {
```
sohkai: What do you think about either placing the web3 instance inside these classes, or extracting the web3 instance used into its own util (e.g. [getWeb3()](https://github.com/aragon/aragon/blob/master/src/web3-utils.js#L134)).

Generally we don't run services in the context of a truffle environment (maybe we should? I'm a bit scared at needing to depend on them for the service to stay alive), so there's no global web3 instance available.

-------

It'd also avoid defining a class in a function, which is a bit awkward, since you can have different definitions if you invoke the function more than once. This then starts to violate the `instanceof` checks and prototype chains, and gets very awkward if you ever need to compare / manipulate more than one object derived from what should be same class.

package.json
facuspagnuolo: Exactly, I was facing this [issue](https://github.com/trufflesuite/ganache-core/issues/417)

lib/relayer/RelayTransactionSigner.js
facuspagnuolo: Completely agree, it should be part of the constructor o at least a known object that resolves that dependency

PR #529, opened 2019-05-20
Add development network to truffle config

### LINKED ISSUE #319: Add development network to truffle config
Now that `aragon-cli` is forcing users to use a `development` network, should we also include one here?

haythem96:
Fixes #319

### DIFF
--- a/truffle-config.js
+++ b/truffle-config.js
@@ -98,6 +98,13 @@ module.exports = {
       gas: 0xffffffffff,
       gasPrice: 0x01
     },
+    development: {
+      host: 'localhost',
+      network_id: '*',
+      port: 7545,
+      gas: 6.9e6,
+      gasPrice: 15000000001
+    },
   },
   build: {},
   mocha,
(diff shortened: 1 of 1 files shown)


sohkai:
Hmm, actually I'm not sure what parameters we should use. I noticed in [aragon-react-boilerplate](https://github.com/aragon/aragon-react-boilerplate/blob/master/truffle.js#L46) that it uses 8545 (the default), but sets `network_id: *`, as I believe we use a random ID for `aragon devchain`.

cc @0x6431346e @0xGabi.

0x6431346e:
> Hmm, actually I'm not sure what parameters we should use. I noticed in [aragon-react-boilerplate](https://github.com/aragon/aragon-react-boilerplate/blob/master/truffle.js#L46) that it uses 8545 (the default), but sets `network_id: *`, as I believe we use a random ID for `aragon devchain`.
> 
> cc @0x6431346e @0xGabi.

Yes, that's why.

sohkai:
@haythem96 Thanks for updating this!

PR #532, opened 2019-06-10
APMRegistry Ruler

0xGabi:
First attempt to solve a limitation with the current `APMRegistry` and `APMRegistryFactory` when developing an organization with the permission structure of [AGP-28](https://github.com/aragon/AGPs/blob/master/AGPs/AGP-28.md#permissions).

In this organization we have an entity that is the "ruler" of all the Repos (`voting` app in our case). With the current smart contracts is not possible to be achieved without updating the permissions after the creation of each Repo.

I have a couple of concerns about the current code:
1. In the `APMRegistryFactory` contract when I included the `_initAPM ` variable I got `Solidity stack too deep` error at compilation and was not sure how to solve it.
2. In the `APMRegistry` contract, I got a possible solution that include two initialize functions. Is this a correct solution? Why or why not? 

TODO:
- [ ] Update tests

- [ ] Fix solidity stack too deep for `APMRegistryFactory`

- [ ] Validate `APMRegistry` improvement

PS: I went ahead and removed the `_initAPM` variable and commented part of the code of the `APMRegistryFactory` (lines 113-127) to deploy a [testing organization on Rinkeby following the AGP-28 structure](https://rinkeby.aragon.org/#/0xc8d9efc88c98680ed8a855e55679d661dd41f067/permissions).

### DIFF
--- a/contracts/factory/APMRegistryFactory.sol
+++ b/contracts/factory/APMRegistryFactory.sol
@@ -54,9 +54,10 @@ contract APMRegistryFactory is APMInternalAppNames {
     * @param _tld The parent node of the controlled subdomain
     * @param _label The subdomain label
     * @param _root Manager for the new aragonPM DAO
+    * @param _initAPM Wheter to initialize the APMRegistry or not
     * @return The new aragonPM's APMRegistry app
     */
-    function newAPM(bytes32 _tld, bytes32 _label, address _root) public returns (APMRegistry) {
+    function newAPM(bytes32 _tld, bytes32 _label, address _root, bool _initAPM) public returns (APMRegistry) {
         bytes32 node = keccak256(abi.encodePacked(_tld, _label));
 
         // Assume it is the test ENS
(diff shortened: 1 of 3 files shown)


0xGabi:
After discussing with @lkngtn we might also want to have an `updateRuler` function behind a role on the `APMRegistry`.

scripts/deploy-apm.js
```diff
@@ -23,11 +24,14 @@ module.exports = async (
     ensAddress = defaultENSAddress,
     owner = defaultOwner,
     daoFactoryAddress = defaultDaoFactoryAddress,
-    verbose = true
+    initAPM = true && defaultInitAPM !== 'false',
```
bingen: Is `true &&` needed here?

contracts/factory/APMRegistryFactory.sol
```diff
@@ -108,18 +109,22 @@ contract APMRegistryFactory is APMInternalAppNames {
...
-        apm.newRepoWithVersion(REPO_APP_NAME, _root, firstVersion, repoBase, b("ipfs:repo"));
+            acl.createPermission(this, apm, apm.CREATE_REPO_ROLE(), this);
 
-        configureAPMPermissions(acl, apm, _root);
+            apm.newRepoWithVersion(APM_APP_NAME, _root, firstVersion, registryBase, b("ipfs:apm"));
+            apm.newRepoWithVersion(ENS_SUB_APP_NAME, _root, firstVersion, ensSubdomainRegistrarBase, b("ipfs:enssub"));
+            apm.newRepoWithVersion(REPO_APP_NAME, _root, firstVersion, repoBase, b("ipfs:repo"));
+
+            configureAPMPermissions(acl, apm, _root);
+        }
```
facuspagnuolo: I don't get why you would like to leave the APM uninitialized. If that's to allow initializing it later using the new syntax implemented above, I'd add a new function here to allow creating a new APM registry using the new initialization logic.

scripts/deploy-apm.js
```diff
@@ -41,7 +45,7 @@ module.exports = async (
   const tldName = 'eth'
   const labelName = 'aragonpm'
   const tldHash = namehash(tldName)
-  const labelHash = '0x'+keccak256(labelName)
+  const labelHash = '0x' + keccak256(labelName)
```
facuspagnuolo:
```suggestion
  const labelHash = `0x${keccak256(labelName)}`
```

scripts/deploy-apm.js
```diff
@@ -51,7 +55,9 @@ module.exports = async (
   const accounts = await getAccounts(web3)
   if (!owner) {
     owner = accounts[0]
-    log('OWNER env variable not found, setting APM owner to the provider\'s first account')
+    log(
+      "OWNER env variable not found, setting APM owner to the provider's first account"
+    )
```
facuspagnuolo: I think it was nicer before 😛

0xGabi:
@facuspagnuolo, @bingen thanks for reviewing this first draft 🙏🏻

I'll continue with both your suggestions over the week. I'm almost done with the new functionalities on aragon CLI related with this change that will allow a better publish workflow for AragonPM

sohkai:
Hmm, not sure we need this change now. The changes to the contract are also a bit touchy and would force an on-chain upgrade (we may want to wait for larger changes from Aragon Black before doing such an upgrade).

For AGP-28, wouldn't we just set the voting instance to be the only one that can create repos and every time we create repos, make sure to set `dev` to be the voting instance? We can add `dev` to the radspec string to make this more clear from a UI-perspective.

sohkai:
Closing this; I'm not sure we need this functionality but if we do, we can consider it when building a new version of aragonPM.

PR #534, opened 2019-06-11
Libs: Provide address utils library

facuspagnuolo:
This PR proposes turning the `IsContract` contract into a Solidity library per se. 

The motivation for this change is underlaid by the idea of discouraging inheritance as a way to reuse helpers. Solidity basically provides two ways of reusing code besides composition, one of those is inheritance and the other is through libraries. Inheritance means a strong relation and in a multiple-inheritance universe it affects the linearization graph. Personally, I think it's better to use libraries if we are simply willing to reuse a helper functionality. For example, IMO, this is why we use `SafeMath` as library and not as an inheritable contract.

Additionally, I examined how much does a contract's bytecode increase applying this change vs as they are now. To do this I crafted a sample contract extending from Kernel (which is already using the `isContract` helper) and reusing said library in the sub contract as well. Using the library in the Kernel contract increases it's deployed bytecode length in 191 bytes, and it increases the deployed bytecode length of the sub Kernel sample in 200 bytes, which means it is not duplicating the whole library in the inherited contract. 

I know we may have different opinions on this topic, but I think we should encourage a single way of doing things, in this case, sharing helper functionalities between contracts.

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -1,16 +1,19 @@
 pragma solidity 0.4.24;
 
-import "../apps/AragonApp.sol";
-import "../common/ConversionHelpers.sol";
-import "../common/TimeHelpers.sol";
-import "./ACLSyntaxSugar.sol";
 import "./IACL.sol";
 import "./IACLOracle.sol";
+import "./ACLSyntaxSugar.sol";
+import "../apps/AragonApp.sol";
+import "../common/TimeHelpers.sol";
+import "../common/AddressUtils.sol";
+import "../common/ConversionHelpers.sol";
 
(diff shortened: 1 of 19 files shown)


sohkai:
I'm not sure we need the `.isZero()` and its counterpart, since those cost a bit too much overhead compared to doing a `== address(0)`.

We were going back and forth on this in #382, and decided to table it for now as a way to save gas. In any case, if we do do this, it'd be in aragonOS 5 and perhaps we can see if a pragma upgrade makes any difference in the gas.

sohkai:
Going to close this for now and remove it from consideration of aragonOS v5.

PR #536, opened 2019-06-21
Add IForwarderFee

sohkai:
Should go into a 4.2.1 release for https://github.com/aragon/tollgate.

### DIFF
--- a/contracts/common/IForwarderFee.sol
+++ b/contracts/common/IForwarderFee.sol
@@ -1,0 +1,10 @@
+/*
+ * SPDX-License-Identitifer:    MIT
+ */
+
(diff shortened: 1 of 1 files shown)


sohkai:
> Looks good but I don't get why this should be part of aragonOS

@facuspagnuolo Good question. I've placed it here to standardize on the forwarding API (so we can write it in the docs and etc.), as aragon.js will need to use this hook to figure out how to do the token approval flow correctly. Other apps may need the same functionality so it felt natural to put it here as another ["app capability"](https://hack.aragon.org/docs/aragonos-ref#application-capabilities).

PR #540, opened 2019-07-13
SafeERC20: add extra returndatasize check to staticInvoke()

sohkai:
For good measure :).

### DIFF
--- a/contracts/common/SafeERC20.sol
+++ b/contracts/common/SafeERC20.sol
@@ -75,7 +75,16 @@ library SafeERC20 {
             )
 
             if gt(success, 0) {
-                ret := mload(ptr)
+                switch returndatasize
+
+                // 32 bytes returned; is valid return
+                case 0x20 {
+                    ret := mload(ptr)
+                }
+                // Else, mark call as failed
+                default {
+                    success := 0
+                }
             }
         }
         return (success, ret);
--- a/contracts/test/mocks/lib/token/TokenBalanceOfAllowanceMissingReturnMock.sol
+++ b/contracts/test/mocks/lib/token/TokenBalanceOfAllowanceMissingReturnMock.sol
@@ -1,0 +1,106 @@
+// Non-standards compliant token that is missing return values for
+// `allowance()` and `balanceOf()`.
+// Modified from https://github.com/OpenZeppelin/openzeppelin-solidity/blob/a9f910d34f0ab33a1ae5e714f69f9596a02b4d91/contracts/token/ERC20/StandardToken.sol
+
+pragma solidity 0.4.24;
+
+import "../../../../lib/math/SafeMath.sol";
+
+
+contract TokenBalanceOfAllowanceReturnMissingMock {
+    using SafeMath for uint256;
+    mapping (address => uint256) private balances;
+    mapping (address => mapping (address => uint256)) private allowed;
+    uint256 private totalSupply_;
+    bool private allowTransfer_;
+
+    event Approval(address indexed owner, address indexed spender, uint256 value);
+    event Transfer(address indexed from, address indexed to, uint256 value);
+
+    // Allow us to set the inital balance for an account on construction
+    constructor(address initialAccount, uint256 initialBalance) public {
+        balances[initialAccount] = initialBalance;
+        totalSupply_ = initialBalance;
+        allowTransfer_ = true;
+    }
+
+    function totalSupply() public view returns (uint256) { return totalSupply_; }
+
+    /**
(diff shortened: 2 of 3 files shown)


contracts/test/mocks/lib/token/TokenBalanceOfAllowanceMissingReturnMock.sol
```diff
@@ -0,0 +1,106 @@
...
+    event Transfer(address indexed from, address indexed to, uint256 value);
+
+    // Allow us to set the inital balance for an account on construction
+    constructor(address initialAccount, uint256 initialBalance) public {
+        balances[initialAccount] = initialBalance;
+        totalSupply_ = initialBalance;
+        allowTransfer_ = true;
+    }
+
+    function totalSupply() public view returns (uint256) { return totalSupply_; }
```
facuspagnuolo: Now that I'm thinking of... shouldn't we provide a static version of `totalSupply` in `SafeERC20`?

contracts/test/mocks/lib/token/TokenBalanceOfAllowanceMissingReturnMock.sol
```diff
@@ -0,0 +1,106 @@
+// Non-standards compliant token that is missing return values for
+// `allowance()` and `balanceOf()`.
+// Modified from https://github.com/OpenZeppelin/openzeppelin-solidity/blob/a9f910d34f0ab33a1ae5e714f69f9596a02b4d91/contracts/token/ERC20/StandardToken.sol
+
+pragma solidity 0.4.24;
+
+import "../../../../lib/math/SafeMath.sol";
+
+
+contract TokenBalanceOfAllowanceReturnMissingMock {
```
facuspagnuolo: WDYT about modeling this case within `TokenReturnMissingMock` as well?

contracts/common/SafeERC20.sol
```diff
@@ -75,7 +75,16 @@ library SafeERC20 {
...
+                switch returndatasize
+
+                // 32 bytes returned; is valid return
+                case 0x20 {
+                    ret := mload(ptr)
+                }
+                // Else, mark call as failed
+                default {
+                    success := 0
+                }
```
facuspagnuolo: Now that we upgraded `solidity-coverage` maybe we can stop using `switch` in favor of `if-else` statements, we can tackle that in a separate PR tho and migrate all at once

contracts/test/mocks/lib/token/TokenBalanceOfAllowanceMissingReturnMock.sol
facuspagnuolo: JIC https://github.com/aragon/aragonOS/pull/543

sohkai: Makes sense; we didn't have a use case for it before but it doesn't hurt to include it :).

contracts/test/mocks/lib/token/TokenBalanceOfAllowanceMissingReturnMock.sol
sohkai: I was thinking about that... but what stopped me was in the `TokenReturnMissingMock` tests we need a working `balanceOf()` to check results against.

contracts/common/SafeERC20.sol
sohkai: Unfortunately there's no `else` to go along with the if ([docs](https://solidity.readthedocs.io/en/v0.4.24/assembly.html#if)) :(

contracts/common/SafeERC20.sol
```diff
@@ -75,7 +75,16 @@ library SafeERC20 {
...
-                ret := mload(ptr)
+                switch returndatasize
+
+                // 32 bytes returned; is valid return
+                case 0x20 {
+                    ret := mload(ptr)
+                }
+                // Else, mark call as failed
+                default {
+                    success := 0
```
izqui: Nitpick: The revert reasons in `staticBalanceOf` and `staticAllowance` denote that the call to the token reverted, which is no longer always the case with this change. I don't think we should change the revert reason but we should at least make a note for this.

izqui: The comment in the function header says "Reverts if the call fails for some reason (should never fail)", we could just update it here.

sohkai: Modified to "(should never fail for correctly implemented ERC20s)"

Issue #514, opened 2019-04-19
Testing improvements


sohkai:
A few items left from the test suite advisory:

- Explicitly test `AppProxyFactory.newPinnedAppProxy(IKernel, bytes32)`
- Explicitly test `EVMScriptRunner.protectState` modifier
- Explicitly test auth modifiers in apps (`APMRegistry`, `Repo`, `EVMScriptRegistry`)

dapplion:
@sohkai I have reviewed the items that you proposed to test and I have a few doubts:
- Regarding `AppProxyFactory.newPinnedAppProxy(IKernel, bytes32)` are you referring to `newAppProxyPinned`?
https://github.com/aragon/aragonOS/blob/d10d195387269e8d51d0aa3e608ec44ecb9b8173/contracts/factory/AppProxyFactory.sol#L38
- Regarding `EVMScriptRunner.protectState`, since [DelegateScript and DeployDelegateScript executors removed](https://hack.aragon.org/docs/aragonos-4-migration#delegatescript-and-deploydelegatescript-executors-removed), is it possible to change the kernel address without a malicious executor?
- Could you provide an example of an existing test that explicitly tests a modifier as a reference?

sohkai:
@dapplion 

> Regarding `AppProxyFactory.newPinnedAppProxy(IKernel, bytes32)` are you referring to `newAppProxyPinned`?

Oops, yes! I've edited the issue :).

> Regarding EVMScriptRunner.protectState, since DelegateScript and DeployDelegateScript executors removed, is it possible to change the kernel address without a malicious executor?

Not with the current set of executors, but this is something that _may_ be possible if alternative executors get installed.

> Could you provide an example of an existing test that explicitly tests a modifier as a reference?

https://github.com/aragon/aragonOS/blob/dev/test/contracts/apps/app_acl.js is a generic test against the ACL's `auth` modifier, but we should also add explicit tests for those named AragonApps to make sure their authentication works as expected.

dapplion:
Thank you so much for your answer @sohkai. I have opened a PR https://github.com/aragon/aragonOS/pull/542 testing item 2 `EVMScriptRunner.protectState`. I still have doubts regarding the other items:
[...]

PR #541, opened 2019-07-13
chore: upgrade solidity-coverage to v0.6+

sohkai:
Upgrades `solidity-coverage`.

A number of tests invoking the fallback function are now failing with an out of gas on the new version, likely due to new instrumentation rules. Previously they were operating within the limit, even under instrumentation.

cc @cgewecke


cgewecke:
@sohkai Taking a look today... `send` and `transfer` continue to fail on our tests as well even though the updated coverage client (6.4.5) uses a new vm option which should make  instrumentation free. For example [these tests]((https://github.com/OpenZeppelin/openzeppelin-solidity/blob/489d2e85f10c68ff977bdd983360b361c4422423/test/introspection/SupportsInterface.behavior.js#L48-L54)) at zeppelin run a gas check for ERC165 interfaces and seem okI.

cgewecke:
@sohkai Could you retry with `0.6.2` when you have a chance? Our tests for this case are finally passing and hopefully you should be able to unskip all send / transfer tests.

sohkai:
@cgewecke Can confirm with the new version that all the ETH transfers are working under coverage 🎉.

Thanks so much for helping out on this!

facuspagnuolo:
Let's create an issue to tackle the uncovered lines

sohkai:
@facuspagnuolo Not sure why coverage decreased (if anything, not skipping those tests should increase it...), but we do have #514 for the few areas of coverage we previously missed :).

PR #547, opened 2019-07-31
Add gas check report in CI

### LINKED ISSUE #370: Add tests to compare size of compiled byte code between changesets
Not quite sure how we'd get the old compiled byte code to diff against the new one... but should be possible.

Could be done via a custom Github app, Danger (e.g. see how [react's dangerfile does the git actions](https://github.com/facebook/react/blob/master/dangerfile.js), or as part of a travis build.

### LINKED ISSUE #544: Add gas check in CI via codecheck
See [set up guide for `eth-gas-reporter`'s integration with CodeCheck](https://github.com/cgewecke/eth-gas-reporter/blob/master/docs/codechecks.md).

Not sure how well this works with monorepos, so we can try to set it up here first.

cgewecke:
Fixes #544, fixes #370.

Hi, opening for further discussion...

Adds a "coveralls for gas" CI hook which reports the changes in gas consumption between a PR and it's target branch. (It's agnostic about whether gas increases are good/bad and always succeeds.)
+ Updates eth-gas-reporter to `^0.2.9`
+ Adds @codechecks/client dev dep (the [CI service provider](https://github.com/codechecks/monorepo/tree/master/packages/client) for the hook)
+ Adds an additional CI job *as a fast-finish / optional failure*. (Because gas monitoring causes tests to run > 2X more slowly and extra info is not critical)

**What this looks like:**
+ [aragonOS new report on GitHub](https://github.com/cgewecke/aragonOS/pull/1/checks?check_run_id=181070157)
+ [example report with changes on GitHub](https://github.com/cgewecke/conditional-tokens-contracts/pull/2/checks?check_run_id=181186156)
+ example Github CI UI
Screen Shot 2019-06-18 at 12 25 49 PM 

**Further installation steps**

+ enable this project at [codechecks.io](https://codechecks.io/) (free for open source)
+ paste a `CC_SECRET` env variable into the Travis settings for this repo

**Things to consider...**
The test suite here is industrial sized and this PR increases your CI overhead a bit. The gas reporter is a little slow because it makes extra data collection calls to the client. Hoping to [change this soon](https://github.com/cgewecke/eth-gas-reporter/issues/171) and get it running as fast as normal tests.  Even if you think this report would be useful you might prefer to wait for performance improvements, idk.


sohkai:
@cgewecke This is awesome! Thank you so much for setting this up, the [results](https://github.com/cgewecke/aragonOS/pull/1/checks?check_run_id=181070157) look incredible!

The deployments section is super interesting, it should resolve our need for something like #370 as well :).

> The test suite here is industrial sized and this PR increases your CI overhead a bit.

Haha I wouldn't call this test suite "industrial" by any means 😉 ([this one is](https://travis-ci.com/aragon/aragon-apps/jobs/220484804)). It seems to run in ~15min right now, and I think that's very reasonable as an additional check.

PR #551, opened 2019-08-27
DepositableDelegateProxy: optimize for EIP-1884

### LINKED ISSUE #549: Istanbul hardfork will break DepositableDelegateProxy
### Background

The upcoming [Istanbul](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1679.md) Ethereum hardfork contains the [EIP-1884](https://eips.ethereum.org/EIPS/eip-1884). EIP-1884 will, besides other things, raise the gas cost of the EVM `SLOAD` instruction from 200 to 800 gas.

### Issue Description

The [fallback function](https://github.com/aragon/aragonOS/blob/c3508233f0a885579d17724b5d2ff47c3cacb897/contracts/common/DepositableDelegateProxy.sol#L10) of `DepositableDelegateProxy`:
```js
    function () external payable {
        // send / transfer
        if (gasleft() < FWD_GAS_LIMIT) {
            require(msg.value > 0 && msg.data.length == 0);
            require(isDepositable());
            emit ProxyDeposit(msg.sender, msg.value);
        } else { // all calls except for send or transfer
            address target = implementation();
            delegatedFwd(target, msg.data);
        }
    }
```
and the relevant `isDepositable()` function:
```js
    function isDepositable() public view returns (bool) {
        return DEPOSITABLE_POSITION.getStorageBool();
    }
```
Basically, as the code comment says, in case the fallback function was triggered by a `.send()` or `.transfer()`, which only send along 2300 gas, the first branch will be executed. Then, `isDepositable()` will be executed and a `ProxyDeposit` event will be emitted.

Currently, the execution of this function consumes 1759 gas. During the execution one `SLOAD` is executed due to `isDepositable()`. Hence, the gas cost **after the hardfork** will be `1759 + 600 = 2359` which will **run out of gas**. Hence, the transaction will generally revert.

### Example

The contract is deployed at address [0x0a74d136fafed0f8d58ce4b7307283695ec7a0b6](https://etherscan.io/address/0x0a74d136fafed0f8d58ce4b7307283695ec7a0b6) and its fallback function was recently executed with 2300 gas as part of the transaction [0x14a4a43b4e9759aac86bb0ae7e5926850406ff1c43ea571239563ff781474ae0](https://etherscan.io/tx/0x14a4a43b4e9759aac86bb0ae7e5926850406ff1c43ea571239563ff781474ae0). 

### Additional Information

[ChainSecurity](https://chainsecurity.com/) has conducted some analysis on the effects of EIP-1884. The intermediate results can be found [here](https://gist.github.com/ritzdorf/1c6bd72955391e831f8a397d3152b4e0).

Martin Holst Swende has collected information on the effects of EIP-1884 [here](https://github.com/holiman/eip-1884-security).

### Contact

Feel free to contact hubert@chainsecurity.com or `@HRitzdorf` on Telegram in case of questions or comments.

izqui:
Optimizes the fallback function in the proxy so it can receive ETH after EIP-1884 is implemented in Istanbul. 

With these changes, the gas cost for executing the fallback function has gone down **from 1759 gas to 1619**, which after the SLOAD opcode is repriced to 800 gas will still execute under 2300 gas (2219 gas).

Closes #549 (even though this issue doesn't tackle the potential migration for all existing DAOs whose Vaults would break)

### DIFF
--- a/contracts/common/DepositableDelegateProxy.sol
+++ b/contracts/common/DepositableDelegateProxy.sol
@@ -17,8 +17,8 @@ contract DepositableDelegateProxy is DepositableStorage, DelegateProxy {
             // Continue only if the gas left is lower than the threshold for forwarding to the implementation code,
             // otherwise continue outside of the assembly block.
             if lt(gas, forwardGasThreshold) {
-                // If the proxy accepts deposits, msg.data.length == 0 and msg.value > 0,
-                // accept the deposit and emit an event
+                // Only accept the deposit and emit an event if all of the following are true:
+                // the proxy accepts deposits (isDepositable), msg.data.length == 0, and msg.value > 0
                 if and(and(sload(isDepositablePosition), iszero(calldatasize)), gt(callvalue, 0)) {
                     let logData := mload(0x40) // free memory pointer
                     mstore(logData, caller) // add 'msg.sender' to the log data (first event param)
--- a/contracts/common/DepositableDelegateProxy.sol
+++ b/contracts/common/DepositableDelegateProxy.sol
@@ -14,7 +14,7 @@ contract DepositableDelegateProxy is DepositableStorage, DelegateProxy {
         // Optimized assembly implementation to prevent EIP-1884 from breaking deposits, reference code in Solidity:
         // https://github.com/aragon/aragonOS/blob/v4.2.1/contracts/common/DepositableDelegateProxy.sol#L10-L20
         assembly {
-            // If the gas left is lower than the threshold for forwarding to the implementation code,
+            // Continue only if the gas left is lower than the threshold for forwarding to the implementation code,
             // otherwise continue outside of the assembly block.
             if lt(gas, forwardGasThreshold) {
                 // If the proxy accepts deposits, msg.data.length == 0 and msg.value > 0,
--- a/contracts/common/DepositableDelegateProxy.sol
+++ b/contracts/common/DepositableDelegateProxy.sol
@@ -8,14 +8,34 @@ contract DepositableDelegateProxy is DepositableStorage, DelegateProxy {
     event ProxyDeposit(address sender, uint256 value);
 
     function () external payable {
-        // send / transfer
-        if (gasleft() < FWD_GAS_LIMIT) {
-            require(msg.value > 0 && msg.data.length == 0);
-            require(isDepositable());
-            emit ProxyDeposit(msg.sender, msg.value);
-        } else { // all calls except for send or transfer
-            address target = implementation();
-            delegatedFwd(target, msg.data);
+        uint256 forwardGasThreshold = FWD_GAS_LIMIT;
+        bytes32 isDepositablePosition = DEPOSITABLE_POSITION;
+
+        // Optimized assembly implementation to prevent EIP-1884 from breaking deposits, reference code in Solidity:
+        // https://github.com/aragon/aragonOS/blob/v4.2.1/contracts/common/DepositableDelegateProxy.sol#L10-L20
+        assembly {
+            // If the gas left is lower than the threshold for forwarding to the implementation code,
+            // otherwise continue outside of the assembly block.
+            if lt(gas, forwardGasThreshold) {
+                // If the proxy accepts deposits, msg.data.length == 0 and msg.value > 0,
+                // accept the deposit and emit an event
+                if and(and(sload(isDepositablePosition), iszero(calldatasize)), gt(callvalue, 0)) {
+                    let logData := mload(0x40) // free memory pointer
+                    mstore(logData, caller) // add 'msg.sender' to the log data (first event param)
+                    mstore(add(logData, 0x20), callvalue) // add 'msg.value' to the log data (second event param)
+
+                    // Emit an event with one topic to identify the event: keccak256('ProxyDeposit(address,uint256)') = 0x15ee...dee1
+                    log1(logData, 0x40, 0x15eeaa57c7bd188c1388020bcadc2c436ec60d647d36ef5b9eb3c742217ddee1)
+
+                    stop() // Stop. Exits execution context
+                }
+
+                // If any of above checks failed, revert the execution (if ETH was sent, it is returned to the sender)
+                revert(0, 0)
+            }
         }
+
+        address target = implementation();
+        delegatedFwd(target, msg.data);
     }
 }
--- a/contracts/common/DepositableDelegateProxy.sol
+++ b/contracts/common/DepositableDelegateProxy.sol
@@ -20,6 +20,9 @@ contract DepositableDelegateProxy is DepositableStorage, DelegateProxy {
                 // Only accept the deposit and emit an event if all of the following are true:
                 // the proxy accepts deposits (isDepositable), msg.data.length == 0, and msg.value > 0
                 if and(and(sload(isDepositablePosition), iszero(calldatasize)), gt(callvalue, 0)) {
+                    // Equivalent Solidity code for emitting the event:
+                    // emit ProxyDeposit(msg.sender, msg.value);
+
                     let logData := mload(0x40) // free memory pointer
                     mstore(logData, caller) // add 'msg.sender' to the log data (first event param)
                     mstore(add(logData, 0x20), callvalue) // add 'msg.value' to the log data (second event param)
--- a/test/contracts/apps/recovery_to_vault.js
+++ b/test/contracts/apps/recovery_to_vault.js
@@ -283,7 +283,7 @@ contract('Recovery to vault', ([ permissionsRoot ]) => {
             // TODO: Remove when the targetted EVM has been upgraded to Istanbul (EIP-1884)
             it('can receive ETH (Istanbul, EIP-1884)', async () => {
               const { gasUsed } = await sendEth({ target, value: 1, gas: SOLIDITY_TRANSFER_GAS - ISTANBUL_SLOAD_GAS_INCREASE })
-              console.log('Used gas:', gasUsed)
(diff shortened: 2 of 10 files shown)


contracts/common/DepositableDelegateProxy.sol
```diff
@@ -8,14 +8,34 @@ contract DepositableDelegateProxy is DepositableStorage, DelegateProxy {
...
-        } else { // all calls except for send or transfer
-            address target = implementation();
-            delegatedFwd(target, msg.data);
+        uint256 forwardGasThreshold = FWD_GAS_LIMIT;
+        bytes32 isDepositablePosition = DEPOSITABLE_POSITION;
+
+        // Optimized assembly implementation to prevent EIP-1884 from breaking deposits, reference code in Solidity:
+        // https://github.com/aragon/aragonOS/blob/v4.2.1/contracts/common/DepositableDelegateProxy.sol#L10-L20
+        assembly {
+            // If the gas left is lower than the threshold for forwarding to the implementation code,
```
sohkai:
```suggestion
            // Continue only if the gas left is lower than the threshold for forwarding to the implementation code,
```

contracts/common/DepositableDelegateProxy.sol
```diff
@@ -8,14 +8,34 @@ contract DepositableDelegateProxy is DepositableStorage, DelegateProxy {
...
+            // otherwise continue outside of the assembly block.
+            if lt(gas, forwardGasThreshold) {
+                // If the proxy accepts deposits, msg.data.length == 0 and msg.value > 0,
```
sohkai: I'd reword this a bit, to clarify it:

```
Only accept the deposit and emit an event if the following are true: the proxy accepts deposits, msg.data.length == 0, and msg.value > 0
```

test/contracts/apps/recovery_to_vault.js
```diff
@@ -1,12 +1,15 @@
 const { hash } = require('eth-ens-namehash')
```
sohkai: It'd probably make sense to move all of these new tests to a new file (perhaps `test/contracts/depositable_delegate_proxy.js`), targeting just the DepositableProxy.

test/contracts/apps/recovery_to_vault.js
```diff
@@ -17,22 +20,53 @@ const TokenReturnFalseMock = artifacts.require('TokenReturnFalseMock')
...
+const ISTANBUL_SLOAD_GAS_INCREASE = 600
 
-contract('Recovery to vault', ([permissionsRoot]) => {
-  let aclBase, appBase, appConditionalRecoveryBase
+contract('Recovery to vault', ([ permissionsRoot ]) => {
+  let aclBase, appBase, appConditionalRecoveryBase, ethSender
   let APP_ADDR_NAMESPACE, ETH
 
   // Helpers
+  const sendEth = async ({ target, value, gas = SOLIDITY_TRANSFER_GAS, shouldOOG }) => {
```
sohkai: Perhaps we can break down this test helper a bit... right now there's a bit too much logic here.

test/contracts/apps/recovery_to_vault.js
```diff
@@ -17,22 +20,53 @@ const TokenReturnFalseMock = artifacts.require('TokenReturnFalseMock')
 const TokenReturnMissingMock = artifacts.require('TokenReturnMissingMock')
 const VaultMock = artifacts.require('VaultMock')
 const KernelDepositableMock = artifacts.require('KernelDepositableMock')
+const EthSender = artifacts.require('EthSender')
 
 const APP_ID = hash('stub.aragonpm.test')
 const EMPTY_BYTES = '0x'
-const SEND_ETH_GAS = 31000 // 21k base tx cost + 10k limit on depositable proxies
+const TX_BASE_GAS = 21000
+const SEND_ETH_GAS = TX_BASE_GAS + 10000 // 10k limit on depositable proxies
```
sohkai: I was looking at some of the other tests, and wondering: which branch is this meant to trigger? This is meant to trigger the proxy deposit branch, right? If so, we should probably add a `- 1` here and in the `app/kernel_funds.js` tests to this to make it more clear, since we do a strict less than 10k check in the code.

test/contracts/apps/recovery_to_vault.js
izqui: Good idea. Reverted the changes in `recovery_to_vault.js` and created some specific tests for `DepositableDelegateProxy`.

Some of the things that we are testing here are already being tested in `recovery_to_vault.js`, `app_funds.js` and `kernel_funds.js`, so we could remove some tests from there.

sohkai: I wouldn't mind keeping those tests there; I view most of those tests as integration tests more than unit tests (pretty much only the bits in common are pure unit tests).

test/contracts/common/depositable_delegate_proxy.js
```diff
@@ -0,0 +1,173 @@
...
+        await sendEthToProxy({ shouldOOG: true, value, gas })
+      })
+
+      it('can receive ETH from contract', async () => {
+        const { tx } = await ethSender.sendEth(proxy.address, { value })
+        const receipt = await web3.eth.getTransactionReceipt(tx)
+        const logs = decodeEventsOfType(receipt, DepositableDelegateProxyMock.abi, 'ProxyDeposit')
+        assertAmountOfEvents({ logs }, 'ProxyDeposit')
+        assertEvent({ logs }, 'ProxyDeposit', { sender: toChecksumAddress(ethSender.address), value })
+      })
```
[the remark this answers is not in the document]
izqui: It is not possible to perform a call with value from a contract with less than 2300 gas, as calls that transfer value get a 2300 gas stipend (at the protocol level).

I was doing some debugging and saw that `receiver.transfer()` sets the gas for the call to 0 if the value sent is greater than 0, and if it is 0 the compiler sets the gas for the call to 2300 gas (which is hardcoded in the bytecode). This results in the contract that receives a `.transfer` always having at least 2300 gas regardless of whether it is coming from the protocol stipend or from the call being sent with 2300 gas.

Even if you do a 'low-level call' (`receiver.call.value(v).gas(g)()`) if `v` is greater than 0, the actual gas that the receiver will be called with is `2300 + g`. This makes it impossible to test the Istanbul scenario in which we'd need receiver to be called with `1700` gas.

test/contracts/common/depositable_delegate_proxy.js
```diff
@@ -0,0 +1,173 @@
...
+    ethSender = await EthSender.new()
+    proxyTargetBase = await ProxyTarget.new()
+    proxyTargetWithFallbackBase = await ProxyTargetWithFallback.new()
+  })
+
+  beforeEach(async () => {
+    proxy = await DepositableDelegateProxyMock.new()
+  })
+
+  const itForwardsToImplementationIfGasIsOverThreshold = () => {
```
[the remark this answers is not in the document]
izqui: It's implicitly guaranteeing it as by not setting the gas explicitly, it uses the gas limit value in `truffle.js`, but I think it is a good idea to make it explicit for these tests

test/contracts/common/depositable_delegate_proxy.js
```diff
@@ -0,0 +1,172 @@
...
+const { getBalance } = require('../../helpers/web3')
+
+// Mocks
+const DepositableDelegateProxyMock = artifacts.require('DepositableDelegateProxyMock')
+const EthSender = artifacts.require('EthSender')
+const ProxyTargetWithoutFallback = artifacts.require('ProxyTargetWithoutFallback')
+const ProxyTargetWithFallback = artifacts.require('ProxyTargetWithFallback')
+
+const TX_BASE_GAS = 21000
+const SEND_ETH_GAS = TX_BASE_GAS + 9999 // 10k gas is the threshold for depositing
```
sohkai: Should we also change [`recovery_to_vault`'s limit](https://github.com/aragon/aragonOS/blob/19756078c111e3ee4a7d06aab39a7ceacc919c61/test/contracts/apps/recovery_to_vault.js#L23) to 30999 (and related values in `kernel_funds` and `app_funds`)?

sohkai:
> Anyway, I was wondering how useful is being able to set the depositable flag on run time

A fundraising app that dynamically wants to start and stop receiving direct deposits.

izqui:
Thanks for the reviews! Addressed all the comments and it is ready to be merged as soon as the CI passes.

bingen:
> A fundraising app that dynamically wants to start and stop receiving direct deposits.

Is this used by [Aragon's fundraising app](https://github.com/AragonBlack/fundraising)? I couldn't find it.
Anyway I'm still not sure it's worth the trouble. I guess for this kind of cases I would just make it non-depositable and then force receiving through a function where I kind implement this. I also guess in most cases it will be needed so for any other particular business logic.

izqui:
> Wouldn't it suffice to have it at deploy time, and then have DepositableDelegateProxy and NonDepositableDelegateProxy, or something like that? Kind of with Pinned and Upgradeable, although I guess having all the combinations could be messy.

This would increase the size of the Kernel considerably if we want to support multiple combinations. We can think about it for the next major version, but I would keep changes to the minimum for this release.

bingen:
Although precisely because how hard are migrations for Proxies the sooner the better, I completely  agree that this is not a good time for big changes, so let's discuss later.
But then I would remove Pinned proxies (which I never saw the point of). Actually, there's no single usage of it in our templates, right? Do we know of any real use of them?
About the Kernel size, do you mean the bytecode size?

bingen:
For future reference, we have some numbers here:
https://github.com/bingen/aragonOS/commit/ac3116c3016a194e9bf3d7b8a5d3261474dfa306

PR #554, opened 2019-09-03
v4.3.0

facuspagnuolo:


sohkai:
I've changed this branch's target to `master`, and we should release off of there.

Afterwards we can re-create the `next` branch off of master.

PR #559, opened 2019-10-08
chore: use shared Aragon truffle config

0xGabi:
This `truffle` config file is still used quite a bit as main configuration and Infura have not been much reliable in some cases reaching the request limits.

Long term we should encourage the use of a [truffle specific package](https://github.com/aragon/contract-helpers/tree/master/packages/truffle-config-v4) once we release it.


sohkai:
Given that we will be replacing this Very Soon<sup>TM</sup> with the separated truffle config, I'd rather just make that change and re-release aragonOS then.

0xGabi:
@sohkai I have updated the PR to use the new truffle v4 package

package.json
```diff
@@ -50,10 +50,9 @@
     "web3-utils": "1.0.0-beta.33"
   },
   "dependencies": {
+    "@aragon/truffle-config-v4": "^1.0.0",
```
sohkai: Let's put this in the `devDependencies`

PR #560, opened 2019-10-17
chore(.gitignore): reintroduce lock file commits

yxliang01:
As per discussion at https://github.com/aragon/aragonOS/pull/373#issuecomment-543268208

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,6 @@
 # Dependencies
 node_modules/
 
-# Lock files
-package-lock.json
-yarn.lock
-
 # coverage
 coverage/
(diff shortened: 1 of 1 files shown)


sohkai:
Thanks @yxliang01! We haven't quite decided on whether we should use yarn or package-lock.jsons, and we will almost certainly only use one or the other.

I'll leave this PR here until we do decide and only include one of them :).

yxliang01:
Hmm only committing one of them makes sense.

sohkai:
We've decided to use yarn.locks across our contract repos and will introduce them shortly into this repo as well.

PR #565, opened 2020-01-10
ACL: remove ACLOracle canPerform() gas limit

willjgriff:
This PR will enable the 1Hive Oracles to function as expected since the Istanbul hard fork. Unfortunately our Oracles `canPerform()` functions can not be guaranteed to execute under the gas limit of `30000`.

Relevant functions in the Oracles can be seen here:
https://github.com/1Hive/token-oracle/blob/master/contracts/TokenBalanceOracle.sol#L59
https://github.com/1Hive/dandelion-voting-app/blob/master/contracts/DandelionVoting.sol#L252

I've replaced the Oracle `canPerform()` `staticcall` gas limit of `30000` with `gasleft()` minus some amount for returning an error (preventing a revert if the `staticcall` runs out of gas). As suggested by @sohkai.

I've observed that if the `staticcall` reverts with an oog after being given all of `gasleft()` it still seems to reserve about `100000` gas, as can be seen with a `gasleft()` directly after the oog `staticcall` (tested on remix). The function doesn't revert and returns false. I've still included the error gas allowance but it may not be necessary.

I had to extract `testRevertOracle` and `testStateModifyingOracle` from the `testOracle` test because when the `staticcall` reverts it consumes all the available gas, leaving very little to none for the remaining tests in the same transaction.

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -3,7 +3,6 @@ pragma solidity 0.4.24;
 import "../apps/AragonApp.sol";
 import "../common/ConversionHelpers.sol";
 import "../common/TimeHelpers.sol";
-import "../lib/math/SafeMath.sol";
 import "./ACLSyntaxSugar.sol";
 import "./IACL.sol";
 import "./IACLOracle.sol";
@@ -12,7 +11,6 @@ import "./IACLOracle.sol";
 /* solium-disable function-order */
 // Allow public initialize() to be first
 contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
-    using SafeMath for uint256;
 
     /* Hardcoded constants to save gas
     bytes32 public constant CREATE_PERMISSIONS_ROLE = keccak256("CREATE_PERMISSIONS_ROLE");
@@ -429,7 +427,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
         bool ok;
 
         if (gasleft() > ERROR_GAS_ALLOWANCE) {
-            uint256 oracleCheckGas = gasleft().sub(ERROR_GAS_ALLOWANCE);
+            uint256 oracleCheckGas = gasleft() - ERROR_GAS_ALLOWANCE;
             assembly {
                 ok := staticcall(oracleCheckGas, _oracleAddr, add(checkCalldata, 0x20), mload(checkCalldata), 0, 0)
             }
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -42,12 +42,9 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     address public constant ANY_ENTITY = address(-1);
     address public constant BURN_ENTITY = address(1); // address(0) is already used as "no permission manager"
 
-    uint256 internal constant ERROR_GAS_ALLOWANCE = 50000;
-
     string private constant ERROR_AUTH_INIT_KERNEL = "ACL_AUTH_INIT_KERNEL";
     string private constant ERROR_AUTH_NO_MANAGER = "ACL_AUTH_NO_MANAGER";
     string private constant ERROR_EXISTENT_MANAGER = "ACL_EXISTENT_MANAGER";
-    string private constant ERROR_ORACLE_OOG = "ACL_ORACLE_OOG";
 
     // Whether someone has a permission
     mapping (bytes32 => bytes32) internal permissions; // permissions hash => params hash
@@ -418,21 +415,17 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     }
 
     function checkOracle(IACLOracle _oracleAddr, address _who, address _where, bytes32 _what, uint256[] _how) internal view returns (bool) {
+        bytes4 sig = _oracleAddr.canPerform.selector;
+
         // a raw call is required so we can return false if the call reverts, rather than reverting
-        bytes memory checkCalldata = abi.encodeWithSelector(_oracleAddr.canPerform.selector, _who, _where, _what, _how);
+        bytes memory checkCalldata = abi.encodeWithSelector(sig, _who, _where, _what, _how);
 
         bool ok;
-        uint256 gasLeftBeforeCall = gasleft();
-
-        if (gasLeftBeforeCall > ERROR_GAS_ALLOWANCE) {
-            uint256 oracleCanPerformGas = gasLeftBeforeCall - ERROR_GAS_ALLOWANCE;
-            assembly {
-                ok := staticcall(oracleCanPerformGas, _oracleAddr, add(checkCalldata, 0x20), mload(checkCalldata), 0, 0)
-            }
+        assembly {
+            ok := staticcall(gas, _oracleAddr, add(checkCalldata, 0x20), mload(checkCalldata), 0, 0)
         }
 
         if (!ok) {
-            require(gasleft() > max(gasLeftBeforeCall / 64, ERROR_GAS_ALLOWANCE), ERROR_ORACLE_OOG);
             return false;
         }
 
@@ -451,10 +444,6 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
         }
 
         return result;
-    }
-
-    function max(uint256 _a, uint256 _b) internal pure returns (uint256) {
-        return _a > _b ? _a : _b;
     }
 
     /**
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -45,7 +45,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     address public constant ANY_ENTITY = address(-1);
     address public constant BURN_ENTITY = address(1); // address(0) is already used as "no permission manager"
 
-    uint256 internal constant ERROR_GAS_ALLOWANCE = 100000;
+    uint256 internal constant ERROR_GAS_ALLOWANCE = 50000;
 
     string private constant ERROR_AUTH_INIT_KERNEL = "ACL_AUTH_INIT_KERNEL";
     string private constant ERROR_AUTH_NO_MANAGER = "ACL_AUTH_NO_MANAGER";
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -418,22 +418,21 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     }
 
     function checkOracle(IACLOracle _oracleAddr, address _who, address _where, bytes32 _what, uint256[] _how) internal view returns (bool) {
-        bytes4 sig = _oracleAddr.canPerform.selector;
-
         // a raw call is required so we can return false if the call reverts, rather than reverting
-        bytes memory checkCalldata = abi.encodeWithSelector(sig, _who, _where, _what, _how);
+        bytes memory checkCalldata = abi.encodeWithSelector(_oracleAddr.canPerform.selector, _who, _where, _what, _how);
 
         bool ok;
+        uint256 gasLeftBeforeCall = gasleft();
 
-        if (gasleft() > ERROR_GAS_ALLOWANCE) {
-            uint256 oracleCheckGas = gasleft() - ERROR_GAS_ALLOWANCE;
+        if (gasLeftBeforeCall > ERROR_GAS_ALLOWANCE) {
+            uint256 oracleCanPerformGas = gasLeftBeforeCall - ERROR_GAS_ALLOWANCE;
             assembly {
-                ok := staticcall(oracleCheckGas, _oracleAddr, add(checkCalldata, 0x20), mload(checkCalldata), 0, 0)
+                ok := staticcall(oracleCanPerformGas, _oracleAddr, add(checkCalldata, 0x20), mload(checkCalldata), 0, 0)
             }
         }
 
         if (!ok) {
-            require(gasleft() > ERROR_GAS_ALLOWANCE, ERROR_ORACLE_OOG);
+            require(gasleft() > max(gasLeftBeforeCall / 64, ERROR_GAS_ALLOWANCE), ERROR_ORACLE_OOG);
             return false;
         }
 
@@ -452,6 +451,10 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
         }
 
         return result;
+    }
+
+    function max(uint256 _a, uint256 _b) internal pure returns (uint256) {
+        return _a > _b ? _a : _b;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


contracts/acl/ACL.sol
```diff
@@ -421,7 +424,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
 
         // a raw call is required so we can return false if the call reverts, rather than reverting
         bytes memory checkCalldata = abi.encodeWithSelector(sig, _who, _where, _what, _how);
-        uint256 oracleCheckGas = ORACLE_CHECK_GAS;
+        uint256 oracleCheckGas = gasleft().sub(ERROR_GAS_ALLOWANCE);
```
sohkai: Wondering about the use of `SafeMath` here, since we don't use it in the [`DelegateProxy`](https://github.com/aragon/aragonOS/blob/next/contracts/common/DelegateProxy.sol#L20) when reserving gas from the delegatecall.

What does the EVM do if this underflows and we pass the extremely large, underflowed value to `staticcall`? Does it allow the called contract to eat up all the remaining gas?

contracts/acl/ACL.sol
```diff
@@ -421,7 +424,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
 
         bool ok;
         assembly {
```
sohkai: Although we've kept this function from having any reverts, I'm wondering if it would be useful to revert if we detect that the oracle used all the available gas.

contracts/test/helpers/ACLHelper.sol
```diff
@@ -57,3 +57,14 @@ contract ConditionalOracle is IACLOracle {
         return how[0] > 0;
     }
 }
+
+contract OverGasLimitOracle is IACLOracle {
+    uint256 outOfGasIterations = 99999; // Actual limit is ~7500, set very high in case of gas cost updates
+
+    function canPerform(address, address, bytes32, uint256[]) external view returns (bool) {
+        for (uint256 i = 0; i < outOfGasIterations; i++) {
```
sohkai: We probably don't even need a limit; could we just do an infinite `while` loop?

sohkai:
> I've observed that if the staticcall reverts with an oog after being given all of gasleft() it still seems to reserve about 100000 gas, as can be seen with a gasleft() directly after the oog staticcall (tested on remix)

I think this is due to the 63/64 gas forward change from [EIP-150](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-150.md); if you test with a lower starting gas amount, you most likely will get a different value.

I'm not super sure what value we should use here, but perhaps we should also decrease the saved amount to ~50k (save at least 1-2 SSTOREs after the ACL call)?

contracts/acl/ACL.sol
willjgriff: My tests on remix suggest it saves ~1/64th of the total gas available, as you mentioned elsewhere, even when passed `uint(-1)` (max number). In which case shall I remove `SafeMath`?

contracts/acl/ACL.sol
willjgriff: I can add a revert statement here. However, since we just get a `false` return value whether it reverts or is OOG, some of the other tests will have to be modified/removed as they expect it not to revert even if the `staticcall` reverts/returns false.

willjgriff: I guess it could give a developer slightly more context as to why a transaction is failing which could be useful.

sohkai: > However, since we just get a false return value whether it reverts or is OOG

We will have to use the `gasleft()` call to detect whether the oracle call ate all the gas (i.e. remaining gas is under `ERROR_GAS_ALLOWANCE`).

contracts/acl/ACL.sol
sohkai: Let's skip SafeMath, but error with the revert if the subtraction would underflow.

I'm thinking something like

```solidity
bool success = false;

if (gasleft() > ERROR_GAS_ALLOWANCE) {
  // do call with adjusted gas limit
}

if (!success) {
  if (gasleft() < ERROR_GAS_ALLOWANCE) {
    revert('ERROR');
  }
  return false;
}

Comment (willjgriff) [reply contracts/acl/ACL.sol]: I'll remove SafeMath and can add the revert check. However, it might be worth noting that `gasleft()` will always be `> 50000` for transactions given over `~3200000` gas due to the 1/64th gas withholding. This means there could be times when the `staticcall` goes OOG but it doesn't reach the `revert()` statement and just returns `false`. The only alternative I can see would be to increase the `ERROR_GAS_ALLOWANCE` to `~125000` gas.

Comment (willjgriff) [reply contracts/acl/ACL.sol]: It turns out that `hasPermission()` calls `evalParam()` twice so even with the lower `ERROR_GAS_ALLOWANCE` it does revert as expected since the second time round it only has the left over gas from the first time, which is tiny and will be smaller than the `ERROR_GAS_ALLOWANCE` once it reaches the revert. Pretty confusing though.

File: contracts/test/helpers/ACLHelper.sol [hunk 4]
```diff
@@ -57,3 +57,10 @@ contract ConditionalOracle is IACLOracle {
         return how[0] > 0;
     }
 }
+
+contract OverGasLimitOracle is IACLOracle {
+    function canPerform(address, address, bytes32, uint256[]) external view returns (bool) {
+        while (true) {}
+        return true;
```
sohkai: As mentioned in the call, we could set a random storage slot to a non-zero value here to make our tests a bit faster :).

contracts/acl/ACL.sol
```diff
@@ -421,14 +422,18 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
 
         // a raw call is required so we can return false if the call reverts, rather than reverting
         bytes memory checkCalldata = abi.encodeWithSelector(sig, _who, _where, _what, _how);
-        uint256 oracleCheckGas = ORACLE_CHECK_GAS;
 
         bool ok;
-        assembly {
-            ok := staticcall(oracleCheckGas, _oracleAddr, add(checkCalldata, 0x20), mload(checkCalldata), 0, 0)
+
+        if (gasleft() > ERROR_GAS_ALLOWANCE) {
```
sohkai: As mentioned in the call, we should use `gasleft()` once and then do operations on the same cached value afterwards to avoid any potential underflows.

contracts/acl/ACL.sol
```diff
@@ -421,14 +422,18 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
+            uint256 oracleCheckGas = gasleft() - ERROR_GAS_ALLOWANCE;
+            assembly {
+                ok := staticcall(oracleCheckGas, _oracleAddr, add(checkCalldata, 0x20), mload(checkCalldata), 0, 0)
+            }
         }
 
         if (!ok) {
+            require(gasleft() > ERROR_GAS_ALLOWANCE, ERROR_ORACLE_OOG);
```
sohkai: As mentioned in the call, we should use `max(gasleft / 64, ERROR_GAS_ALLOWANCE)` as the comparison value here.

contracts/acl/ACL.sol
sohkai: > It turns out that hasPermission() calls evalParam() twice so even with the lower

Ahh yeah, this happens if you set it for `ANY_ENTITY`. However I think we can make a more concise test once we handle the 1/64 gas.

Good catch on the 1/64 stipend though!

contracts/test/helpers/ACLHelper.sol
willjgriff: Used a memory slot instead as it's a view function, to prevent confusion. Test still much faster.

willjgriff:
In the most recent commit, as well as doing the changes requested, I have had to remove the `sig` variable to prevent a `stack too deep` error. I couldn't inline the `oracleCanPerformGas` calculation as it uses a storage variable which can't be accessed in assembly. Any alternative suggestions are welcome. I also had to move the `StateModifyingOracle` test to a JS and remove it from the Solidity tests as this Oracle consumes all the gas it's given as if it's gone OOG.

bingen:
> I think this is due to the 63/64 gas forward change from [EIP-150](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-150.md); if you test with a lower starting gas amount, you most likely will get a different value.

Good catch I forgot about this! Trying to recall it I was reading that EIP and got confused about this:
```
if compustate.gas < gas + extra_gas:
            gas = min(gas, max_call_gas(compustate.gas - extra_gas))
```
If that if condition holds, then `compustate.gas - extra_gas < gas`, and `max_call_gas(compustate.gas - extra_gas) <= compustate.gas - extra_gas`,  therefore the `min` operator inside is not needed, right? What am I missing here? (Not a big deal though, it would just be unnecessary)

contracts/acl/ACL.sol
```diff
@@ -417,18 +418,21 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
+        if (gasLeftBeforeCall > ERROR_GAS_ALLOWANCE) {
+            uint256 oracleCanPerformGas = gasLeftBeforeCall - ERROR_GAS_ALLOWANCE;
+            assembly {
+                ok := staticcall(oracleCanPerformGas, _oracleAddr, add(checkCalldata, 0x20), mload(checkCalldata), 0, 0)
+            }
         }
 
         if (!ok) {
+            require(gasleft() > max(gasLeftBeforeCall / 64, ERROR_GAS_ALLOWANCE), ERROR_ORACLE_OOG);
             return false;
```
bingen: I would refactor all this as:
```
        uint256 gasLeftBeforeCall = gasleft();

        // I don't think we even need this, but if we do I would convert it to a require:
        // require(gasLeftBeforeCall > ERROR_GAS_ALLOWANCE, ERROR_ORACLE_OOG);
        assembly {
            ok := staticcall(gasLeftBeforeCall, _oracleAddr, add(checkCalldata, 0x20), mload(checkCalldata), 0, 0)
        }

        if (!ok) {
            uint256 gasLeftAfterCall = gasLeft();
            require(gasLeftAfterCall > gasLeftBeforeCall / 64 && gasLeftAfterCall > ERROR_GAS_ALLOWANCE), ERROR_ORACLE_OOG);
            return false;
        }
```

(and we can get rid of the `max` function below)

willjgriff: If we use all the available gas in the `staticcall` is there a chance there won't be enough left over for returning the error (even with the 1/64th rule)? After testing I can't find a `gas` limit where that's the case (eg just a `revert` without an error) but it wasn't exhaustive. Also, due to a `stack too deep` error I can't add any more variables eg `gasLeftAfterCall` unless I remove something eg `oracleCanPerformGas`. So I can do this if you think the second `require` statement can always be reached.

bingen: > If we use all the available gas in the `staticcall` is there a chance there won't be enough left over for returning the error (even with the 1/64th rule)?

In that case we could leave the commented `require(gasLeftBeforeCall > ...` check, with the proper amount to make sure that 1/64 of that amount is enough.
Anyway, even if that happen, what would be the problem? It would just revert, right? So compared to what would happen later in `require(gasLeftAfterCall >...` we would just miss the error message? Or am I missing something?

willjgriff: Yes we would just miss the error message. @sohkai do you have a preference? 

I would like to remove the `max()` function but I can't unless I remove the `oracleCanPerformGas` variable (or some unrelated variable somewhere) which would mean passing all of the gas to the function.

sohkai:
Not necessarily affecting the plans, but interesting to know that `gas` may be changing in the future: [EIP-2489 deprecates `gas` and always returns `UINT256_MAX`](https://github.com/ethereum/EIPs/pull/2489).

We use `gas` in a few other important places (`DelegateProxy` and `CallsScriptExecutor`) that would be affected, so I'm not totally concerned about this case, especially since we're already covering the 63/64 case.

sohkai:
@willjgriff I've pushed a few commits. You'll notice I've reverted most of the logic around reserving gas, and just opted to send all the remaining gas to the oracle.

First of all, sorry for making you run around in a few circles in hopes that we could make this mechanism work and provide a better debugging message.

I've thought about it for a while now, and came to the conclusion that trying to save gas is relatively futile.

While executing the revert statement is light on gas—usually lower than 1000 gas, placing the call gas requirement to safely execute the revert to be 40-60k with the 63/64 gas forward rule (it's only really in trouble inside the 2300 gas stipends)—in actual debugging we are likely to suspect out of gas issues through the gas used in a transaction as well as specific traces (parity's is much more helpful than geth's in this manner). In particular, tools like Tenderly also help us debug which parts of a transaction may have used up a lot of gas, and it should be fairly obvious if the oracle took up all the required gas.

I started thinking about just leaving the 1/64 check, but then also started suspecting if this was really required (as I mention above, we have other heuristics to suspect out of gas scenarios). Hardcoding `64` may also not be the best idea in an environment that has no future guarantees, and I started specifically weighing whether the enhanced debugging experience would be better than potentially asking all organizations to upgrade their ACLs if a future hardfork moved this value (especially lower, e.g. `32`).

test/contracts/acl/acl_params.js
```diff
@@ -0,0 +1,101 @@
...
+          await acl.grantPermissionP(ANY_ADDR, mockAppAddress, MOCK_APP_ROLE, [param])
+          assert.isFalse(await acl.hasPermission(ANY_ADDR, mockAppAddress, MOCK_APP_ROLE))
+        })
+      })
+
+      describe('when permission is set for specific address', async () => {
+        // Note `evalParams()` is only called once when calling `hasPermission` for a specific address
+        it('ACL disallows actions', async () => {
+          await acl.grantPermissionP(permissionsRoot, mockAppAddress, MOCK_APP_ROLE, [param])
+          assert.isFalse(await acl.hasPermission(permissionsRoot, mockAppAddress, MOCK_APP_ROLE))
```
sohkai: Note that in these cases that are now not throwing due to sending all gas (truffle by default is using 6.9M gas), it may be interesting to experiment with sending a transaction with lower gas values to see if they start failing and adding some tests around this.

test/contracts/acl/acl_params.js
```diff
@@ -0,0 +1,206 @@
...
+          })
+
+          describe('when permission is set for specific address', async () => {
+            const MEDIUM_GAS = 190000
+            // Note that these gas values are still quite high for causing reverts in "low gas"
+            // situations, as we incur some overhead with delegating into proxies and other checks.
+            // Assuming we incur 40-60k gas overhead for this, we only have ~140,000 gas left.
+            // After the oracle call, we only have 140,000 / 64 ~= 2000 gas left, which begins to
+            // quick run out with SLOADs.
+            const LOW_GAS = 180000
```
sohkai: This seems like quite a lot of gas, since I initially only assumed that we would be reverting under very low gas constraints.

However, in these tests, we are using a full organization, with proxies and all, and this adds a lot of overhead. We revert here because we run out of gas after the SLOAD for the `ANY_ADDR`'s permission params (if we flip the order that `hasPermission()` checks entities such that it checks the `ANY_ADDR` before the specific `who`, we can lower this `LOW_GAS` value).

PR #579, opened 2020-05-27
ACL: Fix inline doc and functions indentation

facuspagnuolo:


contracts/acl/ACL.sol
```diff
@@ -32,7 +32,6 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     uint8 internal constant ORACLE_PARAM_ID       = 203;
     uint8 internal constant LOGIC_OP_PARAM_ID     = 204;
     uint8 internal constant PARAM_VALUE_PARAM_ID  = 205;
-    // TODO: Add execution times param type?
```
facuspagnuolo, line 35: Moved here https://github.com/aragon/aragonOS/issues/580

bingen (approved):

sohkai (approved):
LGTM!

contracts/acl/ACL.sol
Sweetilona3: Well see!😉🤑

PR #581, opened 2020-05-31
Disputable: Implement disputable base app

facuspagnuolo:
image


contracts/apps/disputable/DisputableApp.sol
```diff
@@ -0,0 +1,186 @@
+/*
+ * SPDX-License-Identitifer:    GPL-3.0-or-later
+ */
+
+pragma solidity 0.4.24;
```
sohkai:
```suggestion
/*
 * SPDX-License-Identitifer:    MIT
 */

pragma solidity ^0.4.24;
```

contracts/apps/disputable/DisputableApp.sol
```diff
@@ -0,0 +1,186 @@
...
+import "../../lib/math/SafeMath64.sol";
+
+
+contract DisputableApp is IDisputable, AragonApp {
+    /* Validation errors */
+    string internal constant ERROR_AGREEMENT_NOT_SET = "DISPUTABLE_AGREEMENT_NOT_SET";
+    string internal constant ERROR_AGREEMENT_ALREADY_SET = "DISPUTABLE_AGREEMENT_ALREADY_SET";
+    string internal constant ERROR_SENDER_NOT_AGREEMENT = "DISPUTABLE_SENDER_NOT_AGREEMENT";
+
+    // This role is not required to be validated in the Disputable app, the Agreement app is already doing the check
```
sohkai:
```suggestion
    // This role is used to protect who can challenge actions in derived Disputable apps.
    // However, it is not required to be validated in the app itself as the connected Agreement is responsible for performing the check on a challenge.
```

contracts/apps/disputable/DisputableApp.sol
```diff
@@ -0,0 +1,186 @@
...
+
+    event AgreementSet(IAgreement indexed agreement);
+
+    modifier onlyAgreement() {
+        require(address(_getAgreement()) == msg.sender, ERROR_SENDER_NOT_AGREEMENT);
+        _;
+    }
+
+    /**
+    * @notice Set disputable agreements to `_agreement`
```
sohkai:
```suggestion
    * @notice Set Agreement to `_agreement`
```

contracts/apps/disputable/DisputableApp.sol
```diff
@@ -0,0 +1,186 @@
...
+    * @param _agreement Agreement instance to be linked
+    */
+    function setAgreement(IAgreement _agreement) external auth(SET_AGREEMENT_ROLE) {
+        IAgreement agreement = _getAgreement();
+        require(agreement == IAgreement(0) || _agreement == IAgreement(0), ERROR_AGREEMENT_ALREADY_SET);
```
sohkai: Wondering if we should expose the ability to unset the Agreement.

We were discussing before that it'd be a pretty tricky migration, and perhaps just won't let apps do it?

contracts/apps/disputable/DisputableApp.sol
```diff
@@ -0,0 +1,186 @@
...
+    function onDisputableActionVoided(uint256 _disputableActionId) external onlyAgreement {
+        _onDisputableActionVoided(_disputableActionId);
+    }
+
+    /**
+    * @notice Tells whether the Agreement app is a forwarder or not
+    * @dev IForwarder interface conformance
+    * @return Always true
+    */
+    function isForwarder() external pure returns (bool) {
```
sohkai: I wonder if we should enforce this at this level. I can see disputable apps that are not forwarders (e.g. the Registry).

contracts/apps/disputable/DisputableApp.sol
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    /**
+    * @dev Create a new action in the agreement without lifetime set
+    * @param _disputableActionId Identification number of the disputable action in the context of the disputable
+    * @param _submitter Address of the user that has submitted the action
+    * @param _context Link to a human-readable text giving context for the given action
+    * @return Unique identification number for the created action in the context of the agreement
+    */
+    function _newAgreementAction(uint256 _disputableActionId, address _submitter, bytes _context) internal returns (uint256) {
```
sohkai: Not sure if this overload is _that_ useful, and it might be a point of confusion between the two.

Also wonder if you had any strong opinions on the parameter order; `_disputableActionId, _context, _submitter, _lifetime` might feel more natural?

contracts/apps/disputable/DisputableApp.sol
```diff
@@ -0,0 +1,186 @@
...
+    /**
+    * @dev Close action in the agreement
+    * @param _actionId Identification number of the disputable action in the context of the agreement
+    */
+    function _closeAgreementAction(uint256 _actionId) internal {
+        IAgreement agreement = _ensureAgreement();
+        agreement.closeAction(_actionId);
+    }
+
+    /**
```
sohkai: Maybe we should lift this must-implement functions up to the top of the contract, or separate them into another interface, e.g. `DisputableAppImplementation`?

I'd like to make it obvious to app developers that they need to implement these when working with `DisputableApp`, and why they're implementing these instead of the `IDisputable` interface's functions.

contracts/apps/disputable/IArbitrable.sol
```diff
@@ -0,0 +1,56 @@
+/*
```
sohkai: Not really sure where this and the `IArbitrator` interfaces live best, but perhaps we should mention that it's used by the Court as well?

I wonder if we should eventually maintain some sort of `@aragon/contract-interfaces` package that holds things like this?

contracts/apps/disputable/IDisputable.sol
```diff
@@ -0,0 +1,37 @@
...
+
+import "./IAgreement.sol";
+import "../../common/IForwarder.sol";
+import "../../lib/token/ERC20.sol";
+import "../../lib/standards/ERC165.sol";
+
+
+contract IDisputable is IForwarder, ERC165 {
+    bytes4 internal constant ERC165_INTERFACE_ID = bytes4(0x01ffc9a7);
+    bytes4 internal constant DISPUTABLE_INTERFACE_ID = bytes4(0xef113021);
```
sohkai, line 14: Just checking, do we have a ERC165 bytes generator somewhere?

I feel like this is normally a trivial thing based on the ABI but so many EIPs have discussions around their ERC165 interfaces so I wasn't sure.

contracts/lib/arbitration/IArbitrable.sol
```diff
@@ -0,0 +1,56 @@
+/*
```
sohkai, line 1: It looks like this and the next contract are duplicated in `apps/disputable` (they probably belong better in `lib/` actually)

contracts/test/mocks/lib/arbitration/ArbitrableMock.sol
```diff
@@ -0,0 +1,15 @@
+pragma solidity ^0.4.24;
```
sohkai, line 1: Similar comment to above, this seems to be duplicated

sohkai (commented):
Pretty small notes and nothing major!

The only things I wasn't sure about were:

- Being able to unset the Agreement
- Always being a forwarder
- Some interface structuring / documentation

contracts/apps/disputable/DisputableApp.sol
facuspagnuolo: It could be a forwarder, but I agree it is not necessary to enforce it at this level.

contracts/apps/disputable/DisputableApp.sol
facuspagnuolo: I think it avoids the user from calling `_ensureAgreement`, I like the reordering for the parameters :)

contracts/apps/disputable/IArbitrable.sol
facuspagnuolo: I think it's fine to have these here for now

contracts/apps/disputable/DisputableApp.sol
facuspagnuolo: Note that this is actually allowing to unset the agreement, it only requires the current agreement is not set in that case

contracts/apps/disputable/DisputableApp.sol
facuspagnuolo: Not sure if I'd add another interface for this. Currently, it is "obvious" since simply inheriting from this contract won't compile, developers must implement these methods. 

I lifted them to the top to make it clearer

contracts/apps/disputable/IDisputable.sol
facuspagnuolo: Added one and a few tests for it :)

facuspagnuolo:
@sohkai thanks again for the review! I addressed almost all your comments, let me know your thoughts!

contracts/apps/disputable/DisputableApp.sol
sohkai: Not sure I understand this; doesn't the `_agreement == IAgreement(0)` allow us to unset it?

If anything, it seems like we might want `_agreement != IAgreement(0)`?

contracts/apps/disputable/DisputableApp.sol
sohkai: But the other `_newAgreementAction()` call does the `_ensureAgreement()` check; the user would only have to put the `lifetime` argument in.

contracts/apps/disputable/DisputableAragonApp.sol
```diff
@@ -97,28 +89,54 @@ contract DisputableAragonApp is IDisputable, AragonApp {
...
+    * @dev Void disputable action
+    * @param _disputableActionId Identification number of the disputable action to be voided
+    */
+    function _onDisputableActionVoided(uint256 _disputableActionId) internal;
+
     /**
     * @dev Create a new action in the agreement without lifetime set
     * @param _disputableActionId Identification number of the disputable action in the context of the disputable
     * @param _submitter Address of the user that has submitted the action
     * @param _context Link to a human-readable text giving context for the given action
```
sohkai: If we keep this overload, let's remember to update the natspec :)

Although I'm surprised this even compiles; solidity usually errors if the natspec is wrong.

test/contracts/lib/arbitration/arbitrable.js
```diff
@@ -10,14 +10,20 @@ contract('Arbitrable', () => {
...
       assert.isTrue(await arbitrable.supportsInterface('0x88f3ee69'), 'does not support IArbitrable')
-    })
 
-    it('supports ERC165', async () => {
+      assert.equal(await arbitrable.ARBITRABLE_INTERFACE(), '0x88f3ee69', 'IArbitrable interface ID does not match')
       assert.equal(await arbitrable.interfaceID(), await arbitrable.ARBITRABLE_INTERFACE(), 'IArbitrable interface ID does not match')
     })
+
+    it('does not support 0xffffffff', async () => {
+      assert.isFalse(await arbitrable.supportsInterface('0xffffffff'), 'does support 0xffffffff')
```
sohkai:
```suggestion
      assert.isFalse(await arbitrable.supportsInterface('0xffffffff'), 'should not support 0xffffffff')
```

contracts/apps/disputable/IArbitrable.sol
sohkai: Cool; yeah, for now maybe we can add a README to every `lib/` subdirectory to note where they come from or etc. so there's some context (especially as we don't follow 792 😉)

test/contracts/lib/arbitration/arbitrable.js
```diff
@@ -10,14 +10,20 @@ contract('Arbitrable', () => {
   describe('supportsInterface', () => {
     it('supports ERC165', async () => {
       assert.isTrue(await arbitrable.supportsInterface('0x01ffc9a7'), 'does not support ERC165')
+
+      assert.equal(await arbitrable.ERC165_INTERFACE(), '0x01ffc9a7', 'ERC165 interface ID does not match')
+      assert.equal(await arbitrable.erc165interfaceID(), await arbitrable.ERC165_INTERFACE(), 'ERC165 interface ID does not match')
```
sohkai, line 15: On the interface, I like having these tests, but was mostly curious as to how we came up with our own magic numbers for `IDisputable`.

Are we doing the sanity check via https://github.com/aragon/aragonOS/pull/581/files#diff-a4bee50cd93746db86bb3d3cef7232ebR26-R32? Again, just wondering if there's an automated tool to generate these based on an interface's ABI, so we literally can't miss it if we ever change something.

contracts/apps/disputable/DisputableAragonApp.sol
```diff
@@ -0,0 +1,178 @@
+/*
+ * SPDX-License-Identitifer:    MIT
+ */
+
+pragma solidity 0.4.24;
```
sohkai:
```suggestion
pragma solidity ^0.4.24;
```

contracts/lib/standards/ERC165.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity ^0.4.24;
```
sohkai, line 5: Nitpick: let's also add the SPDX comment to this file

contracts/apps/disputable/DisputableAragonApp.sol
```diff
@@ -0,0 +1,178 @@
...
+import "./IDisputable.sol";
+import "../AragonApp.sol";
+import "../../lib/token/ERC20.sol";
+import "../../lib/math/SafeMath64.sol";
+
+
+contract DisputableAragonApp is IDisputable, AragonApp {
+    /* Validation errors */
+    string internal constant ERROR_AGREEMENT_NOT_SET = "DISPUTABLE_AGREEMENT_NOT_SET";
+    string internal constant ERROR_AGREEMENT_ALREADY_SET = "DISPUTABLE_AGREEMENT_ALREADY_SET";
```
sohkai: In the interest of saving bytecode space, I wonder if we can combine the two errors above into just a single "AGREEMENT_STATE_INVALID" error or etc.

contracts/apps/disputable/DisputableAragonApp.sol
```diff
@@ -0,0 +1,178 @@
...
+    /**
+    * @dev Tell the agreement linked to the disputable instance
+    * @return Agreement linked to the disputable instance
+    */
+    function getAgreement() external view returns (IAgreement) {
+        return _getAgreement();
+    }
+
+    /**
+    * @dev Challenge disputable action
```
sohkai: Just wondering how to make this maximally obvious from a reader's perspective.

Was thinking of a couple of options:

- We could move these up to be with their external counterparts, and mark the whole region as "disputables logic"
- We could change the `@dev` comments to be a bit more clear about this functionality (e.g. `@dev Internal implementation of the onDisputableActionChallenged() hook`)
- We could add some sort of comment heading to the implemented disputable hooks to explain that we're just wrapping them around with the `onlyAgreement` check

sohkai (commented):
Added a couple more comments and replied to a few.

In case those get lost, they're at:

- https://github.com/aragon/aragonOS/pull/581#discussion_r435793181
- https://github.com/aragon/aragonOS/pull/581#discussion_r435793635

contracts/apps/disputable/IAgreement.sol
```diff
@@ -0,0 +1,37 @@
...
+        ERC20 _collateralToken,
+        uint256 _actionAmount,
+        uint256 _challengeAmount,
+        uint64 _challengeDuration
+    )
+        external;
+
+    function unregister(address _disputable) external;
+
+    function canProceed(uint256 _actionId) external view returns (bool);
```
sohkai: What do you think about also including a getter for getting the disputable's settings info, similar to the ones we've already implemented in [`getDisputableInfo()`](https://github.com/aragon/aragon-apps/blob/implement_agreements_app/apps/agreement/contracts/Agreement.sol#L587) and [`getCollateralRequirement()`](https://github.com/aragon/aragon-apps/blob/implement_agreements_app/apps/agreement/contracts/Agreement.sol#L603), but based on a disputable and an action id?

Not sure about the name, but I think it'll be useful as an expected part of the interface of any Agreement so we can easily fetch the collateral requirements from just the app contract and the `IAgreement` ABI.

contracts/apps/disputable/DisputableAragonApp.sol
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    /**
+    * @notice Challenge disputable action #`_disputableActionId`
+    * @param _disputableActionId Identification number of the disputable action to be challenged
+    * @param _challengeId Identification number of the challenge in the context of the Agreement
+    * @param _challenger Address challenging the disputable
+    */
+    function onDisputableActionChallenged(uint256 _disputableActionId, uint256 _challengeId, address _challenger) external onlyAgreement {
+        _onDisputableActionChallenged(_disputableActionId, _challengeId, _challenger);
```
bingen, line 44: What’s the point of using these internal functions? Is it because of the modifier?

contracts/apps/disputable/DisputableAragonApp.sol
```diff
@@ -0,0 +1,178 @@
...
+        _onDisputableActionVoided(_disputableActionId);
+    }
+
+    /**
+    * @notice Set Agreement to `_agreement`
+    * @param _agreement Agreement instance to be linked
+    */
+    function setAgreement(IAgreement _agreement) external auth(SET_AGREEMENT_ROLE) {
+        IAgreement agreement = _getAgreement();
+        require(agreement == IAgreement(0) || _agreement == IAgreement(0), ERROR_AGREEMENT_ALREADY_SET);
```
bingen: Not a big deal, but it wouldn’t make sense to try to remove an agreement which is already removed. Maybe we could add a check to require that not both of them are zero.

contracts/apps/disputable/IAgreement.sol
sohkai: And perhaps we should expose `getChallenge()` as well?

Thinking about this from the perspective of a subgraph/reducer running with only the Voting app and `IArbitrable` ABI in mind. Ideally on each event emitted from one of the `IDisputable`'s callbacks, we are able to reach in and fetch any updated information from the Agreement.

contracts/lib/arbitration/IArbitrable.sol
bingen, line 1: We already have these interfaces in the Court repo. Could we import them from there? Or have wherever a single source of truth?

bingen (approved):
LGTM!

contracts/lib/arbitration/IArbitrable.sol
facuspagnuolo: The source of truth should be aragonOS, although we cannot do that cause we are using different solidity versions :/

contracts/apps/disputable/DisputableAragonApp.sol
facuspagnuolo: Yes. I added more details to the inline doc to make it clearer

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: > What do you think about also including a getter for getting the disputable's settings info

Not sure I follow what's the information you expect to fetch from this new getter, is it the disputable info or the disputable's collateral requirement?

contracts/apps/disputable/DisputableAragonApp.sol
facuspagnuolo: I don't think the compiler fails if the order is not correct

test/contracts/lib/arbitration/arbitrable.js
facuspagnuolo: > On the interface, I like having these tests, but was mostly curious as to how we came up with our own magic numbers for `IDisputable`.

It wasn't magic haha, I computed them manually before writing the tests

> Are we doing the sanity check via https://github.com/aragon/aragonOS/pull/581/files#diff-a4bee50cd93746db86bb3d3cef7232ebR26-R32? Again, just wondering if there's an automated tool to generate these based on an interface's ABI, so we literally can't miss it if we ever change something.

The link is not working for me, although I see your point. I haven't found any tool to compute interface IDs for `ERC165`, we could probably write a tool to have this but we will need to force a weird interface inheritance in the contracts (or make sure we don't inherit at all) to make sure the IDs are computed correctly. Specially thinking about `IArbitrable` and `IDisputable` that inherit from `ERC165`, which makes me think we will definitely moving the potential error to another place instead of solving it.

contracts/apps/disputable/DisputableApp.sol
facuspagnuolo: > Not sure I understand this; doesn't the `_agreement == IAgreement(0)` allow us to unset it?

Yes, but I think the idea was to provide the ability to unset disputable apps from Agreement instances, isn't that the case?

contracts/apps/disputable/DisputableApp.sol
facuspagnuolo: I think offering different syntax sugar for internal functions is great, especially cause these don't occupy space in terms of bytecode size if you don't use it. However, I know this a personal opinion, we can remove it.

contracts/apps/disputable/DisputableAragonApp.sol
facuspagnuolo: I like the idea of making this obvious to the reader definitely.

> We could move these up to be with their external counterparts, and mark the whole region as "disputables logic"

I think this breaks the way we expect functions to be ordered, I don't like this type of inconsistency to be honest. I would try to see if there is a way to make sure this is obvious without doing this.

> We could change the `@dev` comments to be a bit more clear about this functionality

This one I like :)

> We could add some sort of comment heading to the implemented disputable hooks to explain that we're just wrapping them around with the `onlyAgreement` check

This one as well

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: > And perhaps we should expose `getChallenge()` as well?

The problem with this one is that forces you to basically have a lot of information for each challenge, I think adding this one will force us to add the rest of the getters, otherwise it won't make a lot of sense.

contracts/apps/disputable/DisputableAragonApp.sol
```diff
@@ -1,8 +1,8 @@
 /*
```
sohkai, line 1: trick

contracts/apps/disputable/DisputableApp.sol
sohkai: We've simplified this part now, but on the topic of syntax sugar for internal functions: it will depend, but for the case of a single parameter that defaults to 0, I would hesitate to add it simply because it adds more code to look at and remember.

I wish we had kwargs in solidity, but until that point, I think less can go wrong if you only give someone one way to call something :).

test/contracts/lib/arbitration/arbitrable.js
sohkai: > we will need to force a weird interface inheritance in the contracts

Hmm interesting, can the EIP165 interface differ from being the `XOR` of all external function selectors in a given interface (e.g. `IDisputable`'s)? If so, could we take the ABI to automatically generate the selector in tests, to make sure ours match?

For example, I would've expected our `IDisputable` interface identifier to change since we now have more exposed methods.

contracts/apps/disputable/IAgreement.sol
sohkai: > Not sure I follow what's the information you expect to fetch from this new getter, is it the disputable info or the disputable's collateral requirement?

This would be the collateral requirements, since these can change per action.

> The problem with this one is that forces you to basically have a lot of information for each challenge

Let's discuss!

sohkai (approved):
Looking good to me!

Left a few comments in some threads above, in case they get hidden:
- https://github.com/aragon/aragonOS/pull/581#discussion_r436670371
- https://github.com/aragon/aragonOS/pull/581#discussion_r436676334
- https://github.com/aragon/aragonOS/pull/581#discussion_r436695124

test/contracts/lib/arbitration/arbitrable.js
facuspagnuolo: > I would've expected our IDisputable interface identifier to change since we now have more exposed methods.

It does indeed, that's the con, we are not automatically generating it. I need to update it :/

> we will need to force a weird interface inheritance in the contracts

What I meant here is that in order to build an automatic tool, it is not possible to distinguish between the methods coming from inherited contracts, and that's the case for `IDisputable` and `IAgreement`. Both carry their parent's methods, and it is not possible to identify them from the compilation output, we could use the AST tho but it will take a bit more of time :)

facuspagnuolo:
Thanks @sohkai! Let me know your thoughts about the last tweaks!

test/contracts/lib/arbitration/arbitrable.js
sohkai: > it is not possible to distinguish between the methods coming from inherited contracts, and that's the case for `IDisputable` and `IAgreement`

Gotcha. This kind of begs another question: does EIP-165 say anything about this case? In theory, if your interface inherits from another interface, you should consider this new interface as the combination of the whole, not just the new parts.

contracts/apps/disputable/IAgreement.sol
```diff
@@ -0,0 +1,101 @@
...
+    event ActionSubmitted(uint256 indexed actionId);
+    event ActionClosed(uint256 indexed actionId);
+    event ActionChallenged(uint256 indexed actionId, uint256 indexed challengeId);
+    event ActionSettled(uint256 indexed actionId, uint256 indexed challengeId);
+    event ActionDisputed(uint256 indexed actionId, uint256 indexed challengeId);
+    event ActionAccepted(uint256 indexed actionId, uint256 indexed challengeId);
+    event ActionVoided(uint256 indexed actionId, uint256 indexed challengeId);
+    event ActionRejected(uint256 indexed actionId, uint256 indexed challengeId);
+    event DisputableAppRegistered(address indexed disputable);
+    event DisputableAppUnregistered(address indexed disputable);
```
sohkai: Should we add an event for a `DisputableApp`'s collateral settings being changed (since we expose the getter)?

Also wondering if we should just name these as `Disputable` rather than `DisputableApp`, since everywhere else we are standardized on just `Disputable` as the concept.

contracts/apps/disputable/IAgreement.sol
```diff
@@ -0,0 +1,101 @@
...
+    function settle(uint256 _actionId) external;
+
+    function disputeAction(uint256 _actionId, bool _finishedSubmittingEvidence) external;
+
+    function register(
+        address _disputable,
+        ERC20 _collateralToken,
+        uint256 _actionAmount,
+        uint256 _challengeAmount,
+        uint64 _challengeDuration
```
sohkai: Just wondering, can we save a tiny bit of gas by pushing the `uint64` together with an address? Not sure if the storage squeezing works for calldata too (probably not?).

contracts/apps/disputable/IAgreement.sol
```diff
@@ -0,0 +1,101 @@
...
+    )
+        external;
+
+    function unregister(address _disputable) external;
```
sohkai: More of a name thing, but I wonder if we should name this concept as `activate/deactivate`? I think the concept of "registering" ("installing") a Disputable is really different from it being active/de-active (only applicable once "registered").

(And also, small english technicality, [`deregister()`](https://english.stackexchange.com/questions/25931/unregister-vs-deregister) may also be a better choice if we keep the current mechanism).

contracts/apps/disputable/IAgreement.sol
```diff
@@ -0,0 +1,101 @@
...
+            ChallengeState state,
+            bool submitterFinishedEvidence,
+            bool challengerFinishedEvidence,
+            uint256 disputeId,
+            uint256 ruling
+        );
+
+    function getCurrentSettingId() external view returns (uint256);
+
+    function getSetting(uint256 _settingId) external view returns (string title, bytes content, IArbitrator arbitrator);
```
sohkai: Just a thought, but wondering if it would also be worth storing the effective start time of the setting in the storage?

And also a super tiny thing, but we might be able to lower gas usage if the `arbitrator` is the first slot, since it's likely we'll reference that most on-chain.

contracts/apps/disputable/IAgreement.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    function getDisputableInfo(address _disputable) external view returns (bool registered, uint256 currentCollateralRequirementId);
+
+    function getCollateralRequirement(address _disputable, uint256 _collateralId) external view
+        returns (
+            ERC20 collateralToken,
+            uint256 actionAmount,
+            uint256 challengeAmount,
+            uint64 challengeDuration
+        );
```
sohkai: What do you think about re-ordering this file so the the action-related getters are last? Perhaps this ordering of high-level concepts:

1. Signers
2. Settings
3. Disputables
4. Actions

contracts/apps/disputable/IAgreement.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    function unregister(address _disputable) external;
+
+    function getSigner(address _signer) external view returns (uint256 lastSettingIdSigned, bool mustSign);
+
+    function getAction(uint256 _actionId) external view
+        returns (
+            address disputable,
+            uint256 disputableActionId,
+            uint256 collateralId,
```
sohkai: I wonder if this naming might be confusing (you might think the collateral settings live globally in the app, rather than per disputable). Is `disputableCollateralRequirementId` too verbose?

contracts/apps/disputable/IAgreement.sol
```diff
@@ -0,0 +1,101 @@
...
+    function closeAction(uint256 _actionId) external;
+
+    function challengeAction(uint256 _actionId, uint256 _settlementOffer, bool _finishedSubmittingEvidence, bytes _context) external;
+
+    function settle(uint256 _actionId) external;
+
+    function disputeAction(uint256 _actionId, bool _finishedSubmittingEvidence) external;
+
+    function register(
+        address _disputable,
```
sohkai: I wonder if we should just enforce the `IDisputable` type in the interface?

But I guess at this interface level, the Agreement actually doesn't care what gets registered as a potential disputable?

contracts/apps/disputable/IAgreement.sol
```diff
@@ -0,0 +1,101 @@
...
+
+
+contract IAgreement is IArbitrable, IACLOracle {
+
+    event Signed(address indexed signer, uint256 settingId);
+    event ActionSubmitted(uint256 indexed actionId);
+    event ActionClosed(uint256 indexed actionId);
+    event ActionChallenged(uint256 indexed actionId, uint256 indexed challengeId);
+    event ActionSettled(uint256 indexed actionId, uint256 indexed challengeId);
+    event ActionDisputed(uint256 indexed actionId, uint256 indexed challengeId);
```
sohkai, line 23: Maybe we could also log the dispute id from the arbitrator?

contracts/apps/disputable/IAgreement.sol
```diff
@@ -0,0 +1,101 @@
...
+pragma solidity ^0.4.24;
+
+import "../../acl/IACLOracle.sol";
+import "../../lib/token/ERC20.sol";
+import "../../lib/arbitration/IArbitrable.sol";
+
+
+contract IAgreement is IArbitrable, IACLOracle {
+
+    event Signed(address indexed signer, uint256 settingId);
```
sohkai, line 14: Wonder if we should also emit an event if the arbitrator changed?

sohkai (commented):
A couple of notes from looking at `Agreements` and its `IAgreement` interface.

contracts/apps/disputable/IDisputable.sol
```diff
@@ -0,0 +1,42 @@
...
+
+    function onDisputableActionAllowed(uint256 _disputableActionId) external;
+
+    function onDisputableActionRejected(uint256 _disputableActionId) external;
+
+    function onDisputableActionVoided(uint256 _disputableActionId) external;
+
+    function getAgreement() external view returns (IAgreement);
+
+    function getDisputableAction(uint256 _disputableActionId) external view returns (uint64 endDate, bool challenged, bool finished);
```
sohkai: The `challenged` and `finished` bits don't look that useful, as they can be fetched from the `IAgreement`.

I was originally thinking of returning the action `submitter` and `context` from this function, but agree that it is still better owned by the Agreement (as they relate more to creating a dispute than an arbitrary action). `endDate` may also not be that useful anymore since we have `canClose()`.

Maybe something that could be useful to return here is a `bytes` array with the action's evm script (or perhaps have different flags to declare what kind of data is in the bytes)?

contracts/apps/disputable/DisputableAragonApp.sol
```diff
@@ -0,0 +1,167 @@
...
+    // bytes32 public constant CHALLENGE_ROLE = keccak256("CHALLENGE_ROLE");
+    bytes32 public constant CHALLENGE_ROLE = 0xef025787d7cd1a96d9014b8dc7b44899b8c1350859fb9e1e05f5a546dd65158d;
+
+    // bytes32 public constant SET_AGREEMENT_ROLE = keccak256("SET_AGREEMENT_ROLE");
+    bytes32 public constant SET_AGREEMENT_ROLE = 0x8dad640ab1b088990c972676ada708447affc660890ec9fc9a5483241c49f036;
+
+    // bytes32 internal constant AGREEMENT_POSITION = keccak256("aragonOS.appStorage.agreement");
+    bytes32 internal constant AGREEMENT_POSITION = 0x6dbe80ccdeafbf5f3fff5738b224414f85e9370da36f61bf21c65159df7409e9;
+
+    event AgreementSet(IAgreement indexed agreement);
```
sohkai: Maybe this should be part of the interface

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: Yes, it does actually :)

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: Cool, let's add the one for the settings as well

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: > Just a thought, but wondering if it would also be worth storing the effective start time of the setting in the storage?

For now this is actually being effective immediately, we don't have a way to schedule it for a feature, are you suggesting to have it? If not, can't we rely on the block timestamp? actually storing the current timestamp is actually the same

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: I prefer being explicit haha, maybe `collateralRequirementId` ?

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: It doesn't and having circular dependencies it's a PITA with solidity :/
I'd avoid it to be honest, I prefer being explicit about `IAgreement` on the `IDisputable` :)

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: I think it's necessary, and actually is being handled as a new setting, so people will need to re-sign

facuspagnuolo:
Thanks once again @sohkai, addressed/answered all your comments, let me know your thoughts!

contracts/apps/disputable/IAgreement.sol
bingen: Does it? Did you try it or have any reference? According to [official documentation](https://solidity.readthedocs.io/en/v0.5.8/miscellaneous.html#layout-of-call-data) `calldata` is packed following the [ABI specification](https://solidity.readthedocs.io/en/v0.5.8/abi-spec.html), and if I understood it correctly, there is no saving here. See the `baz` example [here](https://solidity.readthedocs.io/en/v0.5.8/abi-spec.html#examples).
Trying to research a bit about it, everything I found seems to confirm this:

https://mudit.blog/solidity-gas-optimization-tips/
“Elements in Memory and Call Data cannot be packed. There is no gas saving in solidity by using smaller variables in function calls and memory”

https://medium.com/coinmonks/gas-optimization-in-solidity-part-i-variables-9d5775e43dde
“Variable packing only occurs in storage — memory and call data does not get packed. You will not save space trying to pack function arguments or local variables.”

contracts/apps/disputable/IAgreement.sol
```diff
@@ -16,12 +16,14 @@ contract IAgreement is IArbitrable, IACLOracle {
     event ActionClosed(uint256 indexed actionId);
     event ActionChallenged(uint256 indexed actionId, uint256 indexed challengeId);
     event ActionSettled(uint256 indexed actionId, uint256 indexed challengeId);
-    event ActionDisputed(uint256 indexed actionId, uint256 indexed challengeId);
+    event ActionDisputed(uint256 indexed actionId, uint256 indexed challengeId, uint256 indexed disputeId);
```
sohkai: Now I'm being annoying, but I wonder if we should actually include the arbitator's address as the indexed parameter and push the `disputeId` to be just data?

Thinking that it'd be more useful to know which arbitrator the dispute was created on, rather than needing to fetch both the `ArbitratorChanged` and `ActionDisputed` events (or use an archival node) to figure that out.

contracts/lib/arbitration/IArbitrator.sol
```diff
@@ -0,0 +1,47 @@
+/*
+ * SPDX-License-Identifier:    MIT
+ */
+
+pragma solidity ^0.4.24;
+
+import "../../lib/token/ERC20.sol";
+
+
+interface IArbitrator {
```
sohkai, line 15: What do you think about clarifying where we got this interface from (i.e. Aragon Court)

contracts/lib/arbitration/IArbitrable.sol
```diff
@@ -0,0 +1,56 @@
...
+ * SPDX-License-Identifier:    MIT
+ */
+
+pragma solidity ^0.4.24;
+
+import "./IArbitrator.sol";
+import "../../lib/standards/ERC165.sol";
+
+
+contract IArbitrable is ERC165 {
```
sohkai, line 17: Similar to above

contracts/apps/disputable/IAgreement.sol
sohkai: Should we add it to this interface?

contracts/apps/disputable/IAgreement.sol
```diff
@@ -0,0 +1,103 @@
...
+
+import "../../acl/IACLOracle.sol";
+import "../../lib/token/ERC20.sol";
+import "../../lib/arbitration/IArbitrable.sol";
+
+
+contract IAgreement is IArbitrable, IACLOracle {
+
+    event Signed(address indexed signer, uint256 settingId);
+    event ActionSubmitted(uint256 indexed actionId);
```
sohkai: Wondering if we should also include the disputable's address as an indexed parameter here too

contracts/apps/disputable/IAgreement.sol
```diff
@@ -0,0 +1,103 @@
...
+
+    function sign() external;
+
+    function newAction(uint256 _disputableActionId, bytes _context, address _submitter) external returns (uint256);
+
+    function closeAction(uint256 _actionId) external;
+
+    function challengeAction(uint256 _actionId, uint256 _settlementOffer, bool _finishedSubmittingEvidence, bytes _context) external;
+
+    function settle(uint256 _actionId) external;
```
sohkai: Given the other functions, maybe this should be named `settleAction()` 😄

sohkai (commented):
Looking good and as with the others, just pushed a few cosmetic commits.

I have a couple last notes on the interfaces; let me know what your thoughts are!

(Also left a comment above, in case it gets lost: https://github.com/aragon/aragonOS/pull/581#discussion_r445654727)

contracts/apps/disputable/IAgreement.sol
sohkai: Sounds good!

I was thinking about this statement I made:

> you might think the collateral settings live globally in the app, rather than per disputable

And realized that, even if it's a bit fictional, the current interface doesn't include anything that forces this arrangement, and so it's fine to be more general.

contracts/apps/disputable/IAgreement.sol
sohkai: So I think this is an interesting getter—normally we wouldn't include the `bytes` and `string` arguments in the getter and log them in events instead.

But since:

1. This is likely to be used a lot from other contexts that have not reduced all the events or want to do an off-chain query (especially for the title)
1. Changing these settings is unlikely to happen very often
1. We will most likely want to display this information in the client without a long(er) load time to fetch all the events

It makes sense to keep the information we otherwise wouldn't find any use for on-chain.

And because of this, I thought it might be useful to also know through the getter when the Agreement was in effect.

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: I meant it will be denoted by the new `settingId`, what thing are you thinking to add to the interface then?

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: I had tried it on Remix and it did consume less gas changing the order:

```
contract Bla {
    event Log();
    
    // 1060 gas
    function foo(address a, uint256 b, uint64 c) external {
        emit Log();
    }
    
    // 1038 gas
    function bar(address a, uint64 b, uint256 c) external {
        emit Log();
    }
}
```

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: Discussed offline, we agreed on removing the `disputeId` since:
1. The information can be retrieved from the exposed getters
2. AragonCourt is already emitting an event indexing the address of the subject (agreement in this case)

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: Discussed offline, we agreed on not adding a separate field or event for this. We can rely on the `Signed` event + the checkpointed information that can be retrieved using `getSetting(uint256 settindID)`

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: Update: This was a false positive, checked again reversing the functions and I got swapped results. Let's revert this change!

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: Discussed offline: similarly to other conclusions, we prefer not having this information off-chain vs spending 20k gas to have it since it can be at least retrieved off-chain

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: Discussed offline: we agreed on adding it since indexing this address adds ~100 gas and it will allow filtering by disputable apps

contracts/apps/disputable/IDisputable.sol
facuspagnuolo: Yeah, I completely agreed with you on this. Let's remove it for now and decide later what we need when we start building the frontend so we can base our decision on solid evidence :)

facuspagnuolo:
@sohkai addressed and answered all your comments! Let me know your thoughts

package.json
```diff
@@ -47,6 +47,7 @@
     "truffle-bytecode-manager": "^1.1.1",
     "truffle-extract": "^1.2.1",
     "web3-eth-abi": "1.2.5",
+    "web3-provider-engine": "^15.0.12",
```
sohkai: Do we need to include this one too?

sohkai (approved):
Finally, I think this has crossed the finish line 🛣 🏁 🏎 🎊

package.json
facuspagnuolo: For some reason, it was complaining locally, let me try removing it

PR #583, opened 2020-06-10
Extend forwarder interface

### LINKED ISSUE #557: IForwarder: provide ability to mark `forward()` as payable
Needed in situations where a forwarder requires some ETH value to be sent. An example would be the applications implementing `IForwarderFee`, e.g. [Tollgate application](https://github.com/aragonone/tollgate), which for now only allows tokens as the fee.

Some potential options:

1. Provide a different `IForwarder` interface, e.g. `IForwarderPayable` (or bundled in `IForwarderFee`), that marks `forward()` as `payable`
1. Change `IForwarder` to be an abstract contract, and mark `forward()` as `payable` but by default reverting if `msg.value > 0`; users would then have to override their own `forward()` and remember to keep the revert if they don't want to accept ETH

It's a little bit ugly, but option 1 seems to be the better choice:

- Does not introduce changes to already deployed contracts
- Most apps will not need their `forward()` function to be `payable`

facuspagnuolo:
Fixes https://github.com/aragon/aragonOS/issues/557

This PR includes the following fixes:
- Provide new forwarder payable interface
- Provide new forwarder interface with an extra data field
- Fix forwarder interfaces to use `external` functions


contracts/forwarding/IForwarder.sol
```diff
@@ -0,0 +1,14 @@
+/*
+ * SPDX-License-Identifier:    MIT
+ */
+
+pragma solidity ^0.4.24;
+
+
+interface IForwarder {
+    function isForwarder() external pure returns (bool);
```
bingen: If we are still using solidity 0.4.24, aren’t we having this problem anymore? Are we enforcing apps to declare them as external? A lot of these functions are `public` right now in aragon-apps.

contracts/forwarding/IForwarderPayable.sol
```diff
@@ -0,0 +1,16 @@
+/*
+ * SPDX-License-Identifier:    MIT
+ */
+
+pragma solidity ^0.4.24;
+
+import "./IForwarder.sol";
```
bingen, line 7: What’s this for?

contracts/forwarding/IForwarderV2.sol
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+pragma solidity ^0.4.24;
+
+import "./IForwarderPayable.sol";
+
+
+// TODO: Cannot inherit interfaces in Solidity 0.4.24
+contract IForwarderV2 is IForwarderPayable {
+    function forward(bytes evmCallScript, bytes context) external payable;
```
bingen: We cannot be sure that this `context` works with more than one step in the forwarding chain, right? What would be an example of a use case for this `context` field?

bingen (approved):

contracts/forwarding/IForwarder.sol
facuspagnuolo: I think these changes will be released as a new major (breaking)

contracts/forwarding/IForwarderV2.sol
facuspagnuolo: > We cannot be sure that this context works with more than one step in the forwarding chain, right?

Right

> What would be an example of a use case for this context field?

Agreements - disputable app's actions :)

contracts/forwarding/IForwarder.sol
sohkai: I think all functions in interfaces should be marked as `external` anyway (is this enforced in solc 0.5?), but AFAIK on 0.4 there was an issue with the dynamically sized types (e.g. `bytes`) that prevented this from being possible (see previous contract's notes).

contracts/forwarding/IForwarderPayable.sol
```diff
@@ -0,0 +1,14 @@
+/*
+ * SPDX-License-Identifier:    MIT
+ */
+
+pragma solidity ^0.4.24;
+
+
+interface IForwarderPayable {
```
sohkai: We should likely inherit from the fee interface here. Otherwise, tooling wouldn't know how much (and which token) you should pay to forward.

contracts/forwarding/IForwarder.sol
```diff
@@ -0,0 +1,14 @@
...
+
+    function canForward(address sender, bytes evmCallScript) external view returns (bool);
+
+    function forward(bytes evmCallScript) external;
```
sohkai: We missed this a long time ago, but we should rename all the `bytes` arguments to `evmScript` instead (a calls script is a specific type of EVM Script in our lingo).

contracts/forwarding/IForwarderV2.sol
```diff
@@ -0,0 +1,13 @@
...
+ * SPDX-License-Identifier:    MIT
+ */
+
+pragma solidity ^0.4.24;
+
+import "./IForwarderPayable.sol";
+
+
+// TODO: Cannot inherit interfaces in Solidity 0.4.24
+contract IForwarderV2 is IForwarderPayable {
```
sohkai: I was wondering if we should instead call this `IForwarderRequiringContext` or etc. and base it off of `IForwarder`.

I don't think we should allow for the base `forward(bytes evmCallScript)` to be implemented in this case.

contracts/forwarding/IForwarderV2.sol
sohkai: Do we still need this to be `payable` with the current work on the per-transaction fees?

contracts/forwarding/IForwarder.sol
bingen: IIRC, in 0.5 it’s enforced, but you are also allowed to overwrite them to public in your implementation. In 0.4 it’s not enforced (there’s only a warning), but the overriding is not possible, which is causing those issues. So if we make it `external` here, we will have to make sure that all apps change it to external too.

contracts/forwarding/IForwarderPayable.sol
facuspagnuolo: I completely agree, the reason why I didn't do that is to avoid the following compiler error:
```
TypeError: Overriding function changes state mutability from "nonpayable" to "payable".
```

contracts/forwarding/IForwarderV2.sol
facuspagnuolo: Yea, definitely the name was to be discussed, proposing `IForwarderWithContext` now, let me know your thoughts!

contracts/forwarding/IForwarderV2.sol
facuspagnuolo: No, but I think it's great to provide all the possible flavors, do you agree?

contracts/forwarding/IForwarderPayable.sol
sohkai: Hmm weird, [`IForwarderFee`](https://github.com/aragon/aragonOS/blob/next/contracts/common/IForwarderFee.sol#L8) shouldn't have any problem with conflicting interface declarations though, right?

facuspagnuolo: Sorry I misread, I thought you were saying about inheriting from `IForwarder`. Sure, let's give it a try

sohkai (approved):
@facuspagnuolo Did you want to try updating the inheritance tree for the forwarders or should we merge?

facuspagnuolo:
I'll analyze the alternative and push another PR if it convinces me :)

PR #584, opened 2020-06-10
ACL: Multiple enhancements

### LINKED ISSUE #582: ACLOracle: update interface to receive `who` properly
Currently, the `ACLOracle` interface does not receive the `who` argument properly when the permission is actually set to `ANY_ENTITY`.

facuspagnuolo:
Fixes #582 

This PR includes:
- Optimization for internal calls: `external` + `internal` over `public` functions
- Allow knowing the original address being evaluated for permissions set to `ANY_ENTITY`

### DIFF
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -391,8 +391,8 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
 
         if (param.id == LOGIC_OP_PARAM_ID) {
             return (Op(param.op) == Op.IF_ELSE)
-                ? _evalIfElseOp(param, _paramsHash, _user, _who, _where, _what, _how)
-                : _evalNonIfElseOp(param, _paramsHash, _user, _who, _where, _what, _how);
+                ? _evalIfElseLogicOp(param, _paramsHash, _user, _who, _where, _what, _how)
+                : _evalNonIfElseLogicOp(param, _paramsHash, _user, _who, _where, _what, _how);
         }
 
         uint256 value;
@@ -423,9 +423,9 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     }
 
     /**
-    * @dev Internal function to eval an IF-ELSE operator
+    * @dev Internal function to eval an IF-ELSE logic operator
     */
-    function _evalIfElseOp(Param _param, bytes32 _paramsHash, address _user, address _who, address _where, bytes32 _what, uint256[] _how)
+    function _evalIfElseLogicOp(Param _param, bytes32 _paramsHash, address _user, address _who, address _where, bytes32 _what, uint256[] _how)
         internal
         view
         returns (bool)
@@ -441,9 +441,9 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     }
 
     /**
-    * @dev Internal function to eval an IF-ELSE operator
+    * @dev Internal function to eval a non-IF-ELSE logic operator
     */
-    function _evalNonIfElseOp(Param _param, bytes32 _paramsHash, address _user, address _who, address _where, bytes32 _what, uint256[] _how)
+    function _evalNonIfElseLogicOp(Param _param, bytes32 _paramsHash, address _user, address _who, address _where, bytes32 _what, uint256[] _how)
         internal
         view
         returns (bool)
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -107,7 +107,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     * @param _app Address of the app in which the role will be allowed (requires app to depend on kernel for ACL)
     * @param _role Identifier for the group of actions in app given access to perform
     */
-    function grantPermission(address _entity, address _app, bytes32 _role) external {
+    function grantPermission(address _entity, address _app, bytes32 _role) external onlyPermissionManager(_app, _role) {
         _grantPermissionP(_entity, _app, _role, new uint256[](0));
     }
 
@@ -119,7 +119,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     * @param _role Identifier for the group of actions in app given access to perform
     * @param _params Permission parameters
     */
-    function grantPermissionP(address _entity, address _app, bytes32 _role, uint256[] _params) external {
+    function grantPermissionP(address _entity, address _app, bytes32 _role, uint256[] _params) external onlyPermissionManager(_app, _role) {
         _grantPermissionP(_entity, _app, _role, _params);
     }
 
@@ -290,7 +290,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     /**
     * @dev Internal function to grant a permission with parameters
     */
-    function _grantPermissionP(address _entity, address _app, bytes32 _role, uint256[] _params) internal onlyPermissionManager(_app, _role) {
+    function _grantPermissionP(address _entity, address _app, bytes32 _role, uint256[] _params) internal {
         bytes32 paramsHash = _params.length > 0 ? _saveParams(_params) : EMPTY_PARAM_HASH;
         _setPermission(_entity, _app, _role, paramsHash);
     }
--- a/contracts/acl/ACL.sol
+++ b/contracts/acl/ACL.sol
@@ -57,13 +57,13 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     event ChangePermissionManager(address indexed app, bytes32 indexed role, address indexed manager);
 
     modifier onlyPermissionManager(address _app, bytes32 _role) {
-        require(msg.sender == getPermissionManager(_app, _role), ERROR_AUTH_NO_MANAGER);
+        require(msg.sender == _getPermissionManager(_app, _role), ERROR_AUTH_NO_MANAGER);
         _;
     }
 
     modifier noPermissionManager(address _app, bytes32 _role) {
         // only allow permission creation (or re-creation) when there is no manager
-        require(getPermissionManager(_app, _role) == address(0), ERROR_EXISTENT_MANAGER);
+        require(_getPermissionManager(_app, _role) == address(0), ERROR_EXISTENT_MANAGER);
         _;
     }
 
@@ -109,8 +109,9 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     */
     function grantPermission(address _entity, address _app, bytes32 _role)
         external
+        onlyPermissionManager(_app, _role)
     {
-        grantPermissionP(_entity, _app, _role, new uint256[](0));
+        _grantPermissionP(_entity, _app, _role, new uint256[](0));
     }
 
     /**
@@ -122,11 +123,10 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     * @param _params Permission parameters
     */
     function grantPermissionP(address _entity, address _app, bytes32 _role, uint256[] _params)
-        public
+        external
         onlyPermissionManager(_app, _role)
     {
-        bytes32 paramsHash = _params.length > 0 ? _saveParams(_params) : EMPTY_PARAM_HASH;
-        _setPermission(_entity, _app, _role, paramsHash);
+        _grantPermissionP(_entity, _app, _role, _params);
     }
 
     /**
@@ -201,7 +201,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
      * @return Length of the array
      */
     function getPermissionParamsLength(address _entity, address _app, bytes32 _role) external view returns (uint) {
-        return permissionParams[permissions[permissionHash(_entity, _app, _role)]].length;
+        return permissionParams[permissions[_permissionHash(_entity, _app, _role)]].length;
     }
 
     /**
@@ -217,7 +217,7 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
         view
         returns (uint8, uint8, uint240)
     {
-        Param storage param = permissionParams[permissions[permissionHash(_entity, _app, _role)]][_index];
+        Param storage param = permissionParams[permissions[_permissionHash(_entity, _app, _role)]][_index];
         return (param.id, param.op, param.value);
     }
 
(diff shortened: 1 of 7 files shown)


contracts/acl/ACL.sol
```diff
@@ -227,8 +227,8 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     * @param _role Identifier for a group of actions in app
     * @return address of the manager for the permission
     */
-    function getPermissionManager(address _app, bytes32 _role) public view returns (address) {
-        return permissionManager[roleHash(_app, _role)];
+    function getPermissionManager(address _app, bytes32 _role) external view returns (address) {
```
bingen, line 224: What’s the reason for this change?

contracts/acl/IACL.sol
```diff
@@ -8,7 +8,5 @@ pragma solidity ^0.4.24;
 interface IACL {
     function initialize(address permissionsCreator) external;
 
-    // TODO: this should be external
-    // See https://github.com/ethereum/solidity/issues/4832
-    function hasPermission(address who, address where, bytes32 what, bytes how) public view returns (bool);
+    function hasPermission(address who, address where, bytes32 what, bytes how) external view returns (bool);
```
bingen, line 11: I’m not aware of any app inheriting from here (other than `ACL` here), so I guess it’s fine, but this may break them.

contracts/acl/ACL.sol
```diff
@@ -109,8 +109,9 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
     */
     function grantPermission(address _entity, address _app, bytes32 _role)
         external
+        onlyPermissionManager(_app, _role)
```
bingen: Is this supposed to be an optimization? I don’t see the benefit in terms of gas usage per call, while I guess the bytecode will increase (for instance, modifiers are inlined).
Besides, if we do this, I would either make the `internal` call `private` or mark it as “unsafe” as we are not checking the `onlyPermissionManager` condition.

contracts/acl/ACL.sol
```diff
@@ -336,15 +395,17 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
         Param memory param = permissionParams[_paramsHash][_paramId];
 
         if (param.id == LOGIC_OP_PARAM_ID) {
-            return _evalLogic(param, _paramsHash, _who, _where, _what, _how);
+            return (Op(param.op) == Op.IF_ELSE)
```
bingen, line 393: I don’t fully get the rationale behind this change either.

contracts/test/helpers/ACLHelper.sol
```diff
@@ -15,27 +15,27 @@ contract ACLHelper {
 
 
 contract AcceptOracle is IACLOracle {
-    function canPerform(address, address, bytes32, uint256[]) external view returns (bool) {
+    function canPerform(address, address, address, bytes32, uint256[]) external view returns (bool) {
```
bingen, line 18: I wonder, does it make sense to assign an ACL Oracle with anything else than `ANY_ENTITY`? Because otherwise we could keep only one param and send the real sender instead of `ANY_ENTITY` in `_hasPermission`.

bingen (approved):

contracts/acl/ACL.sol
facuspagnuolo: Many of these changes were due to avoid "stack too deep" issues

contracts/acl/ACL.sol
facuspagnuolo: I'm proposing to avoid having `public` functions and provide both `external` and `internal` functions instead so we reduce gas costs when using these

contracts/acl/IACL.sol
facuspagnuolo: I think this will be released as a major version so I think we can introduce breaking changes

contracts/test/helpers/ACLHelper.sol
facuspagnuolo: > does it make sense to assign an ACL Oracle with anything else than `ANY_ENTITY`?

Probably it doesn't, but I wouldn't assume that.

contracts/acl/ACL.sol
facuspagnuolo: Yea, we can move the modifier to the `internal` method then

contracts/acl/ACL.sol
bingen: How do you reduce gas costs? Unless the compiler with optimizations enabled decides to inline the internal method, which not only is kind of uncertain but also would increase the bytecode size, I don’t understand the gas benefit of doing it this way.

contracts/test/helpers/ACLHelper.sol
sohkai: Contrived, but you could assign a specific entity (e.g. an Agent) from not performing something via the ACL Oracle.

This could, for example, be used to disallow the AN DAO from changing court values too steeply.

bingen: I don’t get it: so would you assign the permission to the Agent instead of `ANY_ENTITY`? Then only the agent would be able to call the function protected by the ACL Oracle, right?

sohkai: Yes, exactly.

contracts/acl/ACL.sol
sohkai: I think in this case we don't get any gas savings; IIRC you only get them with dynamically sized `calldata` parameters on `external` interfaces.

But I don't mind forcing anything we don't use internally to be `external` anyway—the idea of making something easy to inherit has come and gone, I think.

contracts/acl/ACL.sol
```diff
@@ -239,45 +233,50 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
-        address _where,
-        bytes32 _what,
-        uint256[] _how
-    )
-        public
+    /**
+    * @dev Function called by apps to evaluate ACL params
+    * @param _paramsHash Params hash identifier
+    * @param _user Sender of the original call
+    * @param _who Sender of the original call
```
sohkai:
```suggestion
    * @param _who Grantee of the role
```

contracts/acl/ACL.sol
```diff
@@ -288,11 +287,19 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
         _setPermissionManager(_manager, _app, _role);
     }
 
+    /**
+    * @dev Internal function to grants a permission with parameters
```
sohkai:
```suggestion
    * @dev Internal function to grant a permission with parameters
```

contracts/acl/ACL.sol
```diff
@@ -288,11 +287,19 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
+    */
+    function _grantPermissionP(address _entity, address _app, bytes32 _role, uint256[] _params) internal onlyPermissionManager(_app, _role) {
```
sohkai: I feel more comfortable moving the modifier to the external functions, since that's where it lives for the other functions.

I don't think we'll be running into bytecode size limits with this app :).

contracts/acl/ACL.sol
```diff
@@ -302,6 +309,17 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
         }
     }
 
+    /**
+    * @dev Internal function that sets management
```
sohkai:
```suggestion
    * @dev Internal function that sets management of a role
```

contracts/test/helpers/ACLHelper.sol
bingen: But if you assign the role to a particular entity (Agent for instance), what do you want the new address param for?

contracts/acl/ACL.sol
```diff
@@ -317,14 +335,49 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
+    {
+        if (_paramsHash == EMPTY_PARAM_HASH) {
+            return true;
+        }
+
+        return _evalParam(_paramsHash, 0, _user, _who, _where, _what, _how);
+    }
+
+    /**
+    * @dev Internal function to perform an ACL check
```
sohkai:
```suggestion
    * @dev Internal function to perform an ACL check on a single permission parameter
```

contracts/acl/ACL.sol
```diff
@@ -317,14 +335,49 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
+    function _evalParams(bytes32 _paramsHash, address _user, address _who, address _where, bytes32 _what, uint256[] _how)
+        internal
+        view
+        returns (bool)
+    {
+        if (_paramsHash == EMPTY_PARAM_HASH) {
+            return true;
+        }
+
+        return _evalParam(_paramsHash, 0, _user, _who, _where, _what, _how);
```
sohkai, line 375:
```suggestion
        // `_evalParam()` will internally traverse all the parameters, starting from the first parameter (0)
        return _evalParam(_paramsHash, 0, _user, _who, _where, _what, _how);
```

contracts/acl/IACLOracle.sol
```diff
@@ -6,5 +6,5 @@ pragma solidity ^0.4.24;
 
 
 interface IACLOracle {
-    function canPerform(address who, address where, bytes32 what, uint256[] how) external view returns (bool);
+    function canPerform(address user, address who, address where, bytes32 what, uint256[] how) external view returns (bool);
```
sohkai, line 23: Might be worth adding some comments for the parameters, since `user` and `who` are easily mixed up :).

contracts/acl/ACL.sol
```diff
@@ -363,30 +418,37 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
-        uint32 param2;
+        return _evalParam(_paramsHash, result ? successParam : failureParam, _user, _who, _where, _what, _how);
+    }
 
-        (param1, param2,) = decodeParamsList(uint256(_param.value));
-        bool r1 = _evalParam(_paramsHash, param1, _who, _where, _what, _how);
+    /**
+    * @dev Internal function to eval an IF-ELSE operator
+    */
+    function _evalNonIfElseOp(Param _param, bytes32 _paramsHash, address _user, address _who, address _where, bytes32 _what, uint256[] _how)
```
sohkai:
```suggestion
    /**
    * @dev Internal function to eval a non-IF-ELSE logic operator
    */
    function _evalNonIfElseLogicOp(Param _param, bytes32 _paramsHash, address _user, address _who, address _where, bytes32 _what, uint256[] _how)
```

contracts/acl/ACL.sol
```diff
@@ -409,28 +471,31 @@ contract ACL is IACL, TimeHelpers, AragonApp, ACLHelpers {
...
         // a raw call is required so we can return false if the call reverts, rather than reverting
-        bytes memory checkCalldata = abi.encodeWithSelector(sig, _who, _where, _what, _how);
+        bytes memory checkCalldata = abi.encodeWithSelector(sig, _user, _who, _where, _what, _how);
+        return _checkOracle(_oracleAddr, checkCalldata);
+    }
 
+    /**
+    * @dev Internal function to perform an ACL oracle check
+    */
+    function _checkOracle(IACLOracle _oracleAddr, bytes memory _callData) internal view returns (bool) {
```
sohkai: Could we inline this function with the one before? I don't think it's worth having two internal ones that just use each other :).

contracts/test/tests/TestACLInterpreter.sol
```diff
@@ -87,9 +87,13 @@ contract TestACLInterpreter is ACL, ACLHelper {
 
         // conditional oracle returns true if first param > 0
         ConditionalOracle conditionalOracle = new ConditionalOracle();
-
         assertEval(arr(uint256(1)), ORACLE_PARAM_ID, Op.EQ, uint256(conditionalOracle), true);
         assertEval(arr(uint256(0), uint256(1)), ORACLE_PARAM_ID, Op.EQ, uint256(conditionalOracle), false);
+
+        // only owner oracle returns true if sender is the defined owner
+        OnlyOwnerOracle onlyOwnerOracle = new OnlyOwnerOracle(msg.sender);
```
sohkai, line 94: Might be worth adding another "complex" example with just a bit of logic and the `OnlyOwnerOracle()` to make sure we're threading through the `user` correctly.

test/contracts/acl/acl_params.js
```diff
@@ -114,7 +114,7 @@ contract('ACL params', ([permissionsRoot, specificEntity, noPermission, mockAppA
 
     // Both the assert and oog gas cases should be similar, since assert should eat all the
     // available gas
-    for (let [description, FailingOutOfGasOracle] of [
+    for (const [description, FailingOutOfGasOracle] of [
       ['asserts', AssertOracle],
       ['uses all available gas', OverGasLimitOracle],
     ]) {
```
sohkai, line 133: Similar to above, we might as well add a block to test the `OnlyOwnerOracle` :).

contracts/acl/IACLOracle.sol
```diff
@@ -6,5 +6,5 @@ pragma solidity ^0.4.24;
 
 
 interface IACLOracle {
```
sohkai, line 13: Let's include an `IACLOracleV1.sol` with the old interface.

Some existing apps, like the `TokenBalanceOracle`, will likely want to support both signatures for the time being.

sohkai (approved):
Have a couple of small notes, but looks good to me otherwise 🏎!

test/contracts/acl/acl_params.js
facuspagnuolo: Added! Let me know!

contracts/test/tests/TestACLInterpreter.sol
facuspagnuolo: Added! Let me know!

facuspagnuolo:
@sohkai addressed your comments! Let me know your thoughts!

contracts/test/helpers/ACLHelper.sol
sohkai: Right, attending to the initial question:

> Because otherwise we could keep only one param and send the real sender 

It is contrived, but I consider the old behaviour as "not a bug" specifically because your ACL Oracle could behave differently if it was defined to be completely open for `ANY_ADDRESS` vs. specific to an address.

sohkai (approved):
LGTM, I just added a commit to separate the ACLOracle tests into separate ones with a bit more comments :).

contracts/test/helpers/ACLHelper.sol
bingen: Hm, that makes ACL Oracles even more complex and harder to understand.

sohkai: Agreed—in this case, it was a couple of different concerns:

- This information about the `ANY_ENTITY` grantee was available in the old version
- We needed a different function signature so any combination of ACLv1, ACLv2, ACLOracleV1, and ACLOracleV2 could work together (assuming the app implemented both)

contracts/acl/ACL.sol
bingen: I agree if we don’t use it internally, but this function we do use it, so I don’t see the point. I don’t think there is gas saving here.

PR #586, opened 2020-06-23
Disputable apps: add missing pieces for transaction fees module

bingen:


contracts/lib/arbitration/ITransactionFeesOracle.sol
```diff
@@ -0,0 +1,12 @@
...
+import "../token/ERC20.sol";
+
+
+interface ITransactionFeesOracle {
+    function setFee(bytes32 appId, ERC20 token, uint256 amount) external;
+    function setFees(bytes32[] _appIds, ERC20[] _tokens, uint256[] _amounts) external;
+    function unsetFee(bytes32 _appId) external;
+    function unsetFees(bytes32[] _appIds) external;
+    function getFee(bytes32 appId) external view returns (ERC20, uint256, address);
+}
```
facuspagnuolo, line 12: Do we need this interface here?

contracts/lib/arbitration/IArbitrator.sol
```diff
@@ -44,4 +44,11 @@ interface IArbitrator {
     * @return feeAmount Total amount of fees that must be allowed to the recipient
     */
     function getSubscriptionFees(address _subscriber) external view returns (address recipient, ERC20 feeToken, uint256 feeAmount);
+
+    /**
+    * @dev Tell address of a module based on a given ID
+    * @param _id ID of the module being queried
+    * @return Address of the requested module
+    */
+    function getModule(bytes32 _id) external view returns (address);
```
facuspagnuolo: As mentioned [here](https://github.com/aragon/aragon-court/pull/270#discussion_r445616442), instead of adding this to the `IArbitrator` I would denote that the Agreement app relies on the AragonCourt interface instead. WDYT?

contracts/apps/disputable/IDisputable.sol
```diff
@@ -39,4 +39,6 @@ contract IDisputable is ERC165 {
     function supportsInterface(bytes4 _interfaceId) external pure returns (bool) {
         return _interfaceId == DISPUTABLE_INTERFACE_ID || _interfaceId == ERC165_INTERFACE_ID;
     }
+
+    function appId() public view returns (bytes32);
```
facuspagnuolo, line 43: Maybe we can leverage the new breaking version we are working on to make these `external`

facuspagnuolo (approved):
LGTM! Let's discuss about the `external` version for `appId`

contracts/lib/arbitration/IArbitrator.sol
bingen: But I don’t get it. What do you mean with “denote”? In `Agreement` settings we have an `IArbitrator`, not an `AragonCourt` object.

facuspagnuolo: I mean to state in the Agreement app that we are relying on another interface and not on `IArbitrator`. So I'd simply create a new interface inheriting from `IArbitrator` that adds that method and we use it from the Agreement app. Sth like: 


```sol
contract IAragonCourtArbitrator extends IArbitrator {
    function getModule(bytes32 _id) external view returns (address);
}
```

contracts/lib/arbitration/ITransactionFeesOracle.sol
bingen: You are right!

contracts/lib/arbitration/IArbitrator.sol
bingen: And would you define `IAragonCourtArbitrator` in `aragon-court` repo too?
Not so sure about this option. While it’s true that the modules seem specific to Aragon Court implementation, not inherent to Arbitrator functionality, we need `getModule` to replace what we really need, something like `getTransactionFees` (similar to `getSubscriptionFees`), which is inherent to arbitrator functionality, but that we cannot add because we can’t upgrade the Controller.
Anyway, I’ll do it as you suggest.

contracts/apps/disputable/IDisputable.sol
bingen: What’s the benefit?

contracts/lib/arbitration/ITransactionFeesOracle.sol
bingen: We need  it for `Agreement`. Would you rather put it in `aragon-apps` repo? Taking into account that even `IAgreement` is here, it seems it makes sense to have it here too.

contracts/apps/disputable/IDisputable.sol
facuspagnuolo: To avoid having the compiler warning:
> Warning: Functions in interfaces should be declared external.

contracts/lib/arbitration/IArbitrator.sol
facuspagnuolo: For the record, we discussed offline the idea of providing here the interface for the transaction fees oracle. Then `IAgreement` will rely on both, `IArbitrator` and `ITransactionFeesOracle`, besides the `ITransactionFeesOracle` implementation is set as a module in Aragon Court.

facuspagnuolo:
@bingen I have merged the base disputable app PR against `next`. Feel free to rebase or pull to improve the diff for the reviews please 🙏

contracts/apps/disputable/IDisputable.sol
bingen: Where do you see that warning? I can’t find it. Despite the name, `IDisputable` is not actually an interface, so the warning is not showing up for me.

facuspagnuolo: not sure to be honest, however this is a minor detail, not a blocker at all

contracts/test/mocks/apps/disputable/DisputableAppMock.sol
```diff
@@ -36,11 +36,15 @@ contract DisputableAppMock is DisputableAragonApp {
...
-        iDisputable.onDisputableActionAllowed.selector ^
-        iDisputable.onDisputableActionRejected.selector ^
-        iDisputable.onDisputableActionVoided.selector ^
-        iDisputable.getAgreement.selector;
+            iDisputable.onDisputableActionChallenged.selector ^
+            iDisputable.onDisputableActionAllowed.selector ^
+            iDisputable.onDisputableActionRejected.selector ^
+            iDisputable.onDisputableActionVoided.selector ^
+            iDisputable.getAgreement.selector ^
+            iDisputable.getDisputableAction.selector ^
```
facuspagnuolo: We removed this one for now, the CI is failing because of this, let's fix it and merge after it passes!

facuspagnuolo (approved):

contracts/test/mocks/apps/disputable/DisputableAppMock.sol
bingen: Yes, I saw, but forgot to push last night, doing it now.

contracts/apps/disputable/IAgreement.sol
```diff
@@ -12,6 +12,7 @@ import "../../lib/arbitration/IArbitrable.sol";
 contract IAgreement is IArbitrable, IACLOracle {
 
     event Signed(address indexed signer, uint256 settingId);
+    event TransactionFeesOracleSet(address transactionFeesOracle);
```
facuspagnuolo: I think we can remove this event and leverage this as a new part of the agreement settings

contracts/apps/disputable/IDisputable.sol
sohkai, line 43: Somewhat related discussion, but maybe we should create a new interface, e.g. `IAragonApp` to enforce the `kernel()` and `appId()` functions?

That way `IDisputable` can declare that it's always an `IAragonApp` (and we can adjust the EIP-165 values to be separated between `IDisputable` and `IAragonApp`).

-----

We could also move out `supportsInterface()` out of the base app—it's kind of weird that it implements it as an app may choose to support any number of interfaces (e.g. the token receiver ones from 721 or 1155).

sohkai (approved):
LGTM, although I'm not actually sure if the `TransactionFeesOracle` contract should live here (are users meant to import it?).

contracts/apps/disputable/IDisputable.sol
facuspagnuolo: I like the idea of `IAragonApp`, I'll create an issue for that :)

facuspagnuolo:
@sohkai the thing is that `IAgreement` relies on it, would you declare it in the agreement app directly?

My rationale between these interfaces that are being shared between Aragon Court and Aragon apps, is that they should live in aragonOS. Currently, we have an incompatibility problem since Aragon Court is using solc 0.5.8. Thus, these interfaces are being repeated in Aragon Court's repo. What do you think?

sohkai:
Didn't see that we were using it in `IAgreement`. Sounds good to me then!

> My rationale between these interfaces that are being shared between Aragon Court and Aragon apps, is that they should live in aragonOS

Maybe we should have an `@aragon/contract-interfaces` package or etc. to hold these, but not really sure I like that.

bingen:
> Didn't see that we were using it in IAgreement

There was a commit not yet pushed at the time of your comment ;-)

PR #588, opened 2020-07-01
Disputable apps: Implement Aragon app fees cashier interface

bingen:
- Add payment endpoint
- Rename methods to be “transaction specific”
- Add events


bingen:
Mostly motivated by discussion [here](https://github.com/aragon/aragon-apps/issues/1186#issuecomment-652076269)

sohkai (approved):
LGTM, although I actually don't mind the old names :shrug:

bingen:
The new names are because if we end up having a Subscripitons contract that implements also the Transactions interface, things like `getFees` would be too ambiguous.

bingen:
Also, I’m thinking of removing the `Oracle` part from the interface name, as now it also has the payments endpoint.

facuspagnuolo (commented):
Based on our discussion offline, I would:
- Rename the interface to `AragonAppFeesCashier`
- Rename the `(.*)(TransactionFee)(s)?` methods to `$1AppFee$3`
- Rename `uint256 _actionId` to a `bytes _data` instead

contracts/lib/arbitration/IAragonAppFeesCashier.sol
```diff
@@ -0,0 +1,17 @@
...
+interface IAragonAppFeesCashier {
+    event AppFeeSet(bytes32 indexed appId, ERC20 token, uint256 amount);
+    event AppFeeUnset(bytes32 indexed appId);
+    event AppFeePaid(address indexed by, bytes32 appId, uint256 actionId);
+
+    function setAppFee(bytes32 _appId, ERC20 _token, uint256 _amount) external;
+    function setAppFees(bytes32[] _appIds, ERC20[] _tokens, uint256[] _amounts) external;
+    function unsetAppFee(bytes32 _appId) external;
+    function unsetAppFees(bytes32[] _appIds) external;
+    function payAppFees(bytes32 _appId, uint256 _actionId) external;
```
facuspagnuolo: Let use a `bytes data` here instead of `uint256 actionId`

facuspagnuolo (approved):
LGTM!

PR #589, opened 2020-07-01
ACL: update user/who parameters to who/grantee

sohkai:
Updates the parameter names for the new `user`/`who` concepts for ACLOracleV2 to be `who`/`grantee`.

I wanted to keep `who` the same as before, as we use `who` as the original sender everywhere else (e.g. `hasPermission()`).


facuspagnuolo:
Let's rebase this one against `next` to facilitate the review 🙏

sohkai:
Updated now!

facuspagnuolo (approved):
Much clearer! Let's fix the linter and merge :)

PR #590, opened 2020-07-02
Forwarding: update forwarding interface names, add tests

sohkai:
This admittedly does a couple of refactors / changes to the Forwarding interfaces from #583:

- Moves the old base `IForwarder` interface to be `IAbstractInterface`, so we can reserve `IForwarder` as the "normal" (most basic and most commonly used) forwarding type
- Adds the `IForwarderFee` interface to the payable interfaces
- Updates the forwarding types to be an enum (encoded as uint8 I think?), with just two types
  - We can detect payable simply by the `forwardFee()` method returning `EtherTokenConstant`
- Maintains the `isForwarder()` interface for backwards compatibility with aragonOS 4
- Adds some tests for behaviour

May be easiest to review by just looking at the [branch's code](https://github.com/aragon/aragonOS/blob/extend_forwarder_interface_update/contracts/forwarding).

------

> Maintains the `isForwarder()` interface for backwards compatibility with aragonOS 4

This is somewhat debatable, but I wanted to keep this because:

- It allows off-chain tooling to continue accessing `isForwarder()`, and in the majority of the case (e.g. `IForwarder`), this will continue to just work without any additional checks
- It allows updated tooling to do both the `isForwarder()` and `forwarderType()` checks, and if `isForwarder()` is true but `forwarderType()` returns 0, know they're dealing with an older aragonOS 4 contract (which only has one forwarding interface).


facuspagnuolo (approved):
LGTM!

PR #593, opened 2020-07-15
Disputable: Support payable actions

facuspagnuolo:


test/contracts/apps/disputable/disputable_app.js
```diff
@@ -135,14 +135,25 @@ contract('DisputableApp', ([_, owner, agreement, anotherAgreement, someone]) =>
...
-        const agreement = await AgreementMock.new()
+        agreement = await AgreementMock.new()
         await disputable.setAgreement(agreement.address, { from: owner })
       })
 
       it('does not revert', async () => {
         await disputable.newAction(0, '0x00', owner)
       })
+
+      it('transfer funds when sent', async () => {
```
sohkai:
```suggestion
      it('receives ETH when sent', async () => {
```

Maybe this is a better description? My thought line here is that this test is pretty hard to tell what it's actually testing at the moment, especially out of context.

contracts/apps/disputable/DisputableAragonApp.sol
```diff
@@ -129,7 +129,7 @@ contract DisputableAragonApp is IDisputable, AragonApp {
     */
     function _newAgreementAction(uint256 _disputableActionId, bytes _context, address _submitter) internal returns (uint256) {
         IAgreement agreement = _ensureAgreement();
-        return agreement.newAction(_disputableActionId, _context, _submitter);
+        return agreement.newAction.value(msg.value)(_disputableActionId, _context, _submitter);
```
sohkai, line 132: I wonder if we shouldn't expose this as a parameter instead; maybe the app takes a cut of ETH itself or etc.

sohkai (commented):
Couple of notes, but looking good!

contracts/apps/disputable/DisputableAragonApp.sol
facuspagnuolo: If the app manipulates it, the agreement app won't be able to pay fees, unless the user is sending more than the required amount. However, how would a parameter help here?

bingen (approved):

contracts/apps/disputable/DisputableAragonApp.sol
sohkai: Yes, I was imagining the user sending more fees to the app (because of app-specific logic).

As an example, in the case of a registry, maybe you need to stake some additional ETH into the app as well as pay the fee.

The parameter would help here because the app then has control over how much value is sent.

PR #594, opened 2020-07-15
Disputables: Avoid calling close action if closed

facuspagnuolo:
Related to https://github.com/aragon/aragon-apps/pull/1214

This PR adds an improvement to avoid calling `IAgreement#closeAction` from the internal helper method of `DisputableAragonApp` if the requested action is already closed.

### DIFF
--- a/contracts/apps/disputable/DisputableAragonApp.sol
+++ b/contracts/apps/disputable/DisputableAragonApp.sol
@@ -138,7 +138,11 @@ contract DisputableAragonApp is IDisputable, AragonApp {
     */
     function _closeAgreementAction(uint256 _actionId) internal {
         IAgreement agreement = _ensureAgreement();
-        agreement.closeAction(_actionId);
+        (,,,,, bool closed,,) = agreement.getAction(_actionId);
+
+        if (!closed) {
+            agreement.closeAction(_actionId);
+        }
     }
 
     /**
--- a/test/contracts/apps/disputable/disputable_app.js
+++ b/test/contracts/apps/disputable/disputable_app.js
@@ -1,6 +1,7 @@
 const { assertRevert } = require('../../../helpers/assertThrow')
 const { getEventArgument } = require('../../../helpers/events')
 const { getNewProxyAddress } = require('../../../helpers/events')
+const { decodeEventsOfType } = require('../../../helpers/decodeEvent')
 const { assertEvent, assertAmountOfEvents } = require('../../../helpers/assertEvent')(web3)
 
 const ACL = artifacts.require('ACL')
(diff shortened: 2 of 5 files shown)


bingen (approved):
LGTM

test/contracts/apps/disputable/disputable_app.js
```diff
@@ -142,8 +143,12 @@ contract('DisputableApp', ([_, owner, agreement, anotherAgreement, someone]) =>
         await disputable.setAgreement(agreement.address, { from: owner })
       })
 
-      it('does not revert', async () => {
-        await disputable.newAction(0, '0x00', owner)
+      it('submits a new action', async () => {
+        const { receipt } = await disputable.newAction(0, '0x00', owner)
```
sohkai, line 147: Looks like some opportunities to use the `EMPTY_BYTES` :)

test/contracts/apps/disputable/disputable_app.js
```diff
@@ -142,8 +143,12 @@ contract('DisputableApp', ([_, owner, agreement, anotherAgreement, someone]) =>
...
+
+        const logs = decodeEventsOfType(receipt, AgreementMock.abi, 'NewAction')
+        assertAmountOfEvents({ logs }, 'NewAction')
+        assertEvent({ logs }, 'NewAction', { disputableActionId: 0, context: '0x00', submitter: owner })
```
sohkai, line 151: Minor nitpick, but we could use embed use the `decodeAbiFor` option now :)

contracts/apps/disputable/DisputableAragonApp.sol
```diff
@@ -138,7 +138,11 @@ contract DisputableAragonApp is IDisputable, AragonApp {
     */
     function _closeAgreementAction(uint256 _actionId) internal {
         IAgreement agreement = _ensureAgreement();
-        agreement.closeAction(_actionId);
+        (,,,,, bool closed,,) = agreement.getAction(_actionId);
+
+        if (!closed) {
+            agreement.closeAction(_actionId);
```
sohkai, line 144: Not opposed to it, but I think we should do a small trial run to see how much the gas difference actually is, if this is just an optimization.

The extra call to `agreement.getAction()` may remove the large part of the savings if `closeAction()` is fairly simple on already-closed actions.

sohkai (approved):
Couple of quick comments, main one being sure that this is actually an optimization!

test/contracts/apps/disputable/disputable_app.js
facuspagnuolo: Remember we don't have the helpers here yet haha, but I'll extract this to a constant so we don't miss it :)

test/contracts/apps/disputable/disputable_app.js
facuspagnuolo: We don't have the helpers here yet haha

contracts/apps/disputable/DisputableAragonApp.sol
facuspagnuolo: Yea! I tested this and it is not an optimizing it actually, I'll roll it back

facuspagnuolo:
Closing in favor of https://github.com/aragon/aragon-apps/pull/1231

PR #597, opened 2020-07-20
Add IAragonApp interface

### LINKED ISSUE #587: Apps: Create IAragonApp interface
See https://github.com/aragon/aragonOS/pull/586#discussion_r447643866 for reference

We can create a new interface called `IAragonApp` or similar that enforces `kernel()` and `appId()` functions

bingen:
Fixes #587 
See: https://github.com/aragon/aragonOS/pull/586#discussion_r447643866


contracts/apps/IAragonApp.sol
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+pragma solidity ^0.4.24;
+
+import "../kernel/IKernel.sol";
+
+
+contract IAragonApp {
+    function kernel() public view returns (IKernel);
+    function appId() public view returns (bytes32);
```
facuspagnuolo, line 15: Should we leverage this change and make these `external`?

contracts/apps/AppStorage.sol
```diff
@@ -6,9 +6,10 @@ pragma solidity ^0.4.24;
 
 import "../common/UnstructuredStorage.sol";
 import "../kernel/IKernel.sol";
+import "./IAragonApp.sol";
 
 
-contract AppStorage {
+contract AppStorage is IAragonApp {
```
facuspagnuolo, line 12: Does it work if we enforce this in `AragonApp` directly?

bingen: Nope, as `AppProxyBase` needs it. Actually, this was a big headache for my last commit (cc88e41), until I realized that `supportsInterface` was being shortcut and was not going through the delegated call because of this inheritance. :see_no_evil:

contracts/apps/IAragonApp.sol
bingen: Why? Those functions are being called by other contracts and in Solidity 4 we can’t override from `external` to `public`.

facuspagnuolo: I don't see a problem in introducing a breaking change if we are going to use semver properly to release a breaking version, if those contracts want to use this new version then they must update their usage

bingen: Ok, but what’s the advantage of making them `external`?

contracts/apps/AppStorage.sol
facuspagnuolo: But the `AppProxyBase` doesn't need to support the ERC165, I'm thinking about:

- Make `AragonApp` inherit from `IAragonApp`
- Leave the `AppStorage` contract as it is (without inheriting from `IAragonApp`
- Remove the forwarding for `supportsInterface` in `AppProxyBase`

Note that we want the AragonApp to support the `IAragonApp` interface, not the proxy. If the proxy does to shortcut its execution that's another part of the story I think. If we want to enforce the `kernel` and `appId` methods in the proxy as well, then I'd split the ERC165 part, otherwise any app proxy will be forced to that

bingen: I tried to do that, but it didn’t compile because `kernel` and `appId` are defined in `AppStorage`. In particular it says: `Redeclaring an already implemented function as abstract`.

> otherwise any app proxy will be forced to that

I don’t fully get this. It will be the same in both cases: the proxy will redirect to the implementation when called `supportsInterface`.

bingen: EDIT: I realize that changing the inheritance order, it works, so trying that solution again. It feels kind of weird to me that the contract implementing the functions can’t explicitly inherit the interface defining them though.

contracts/apps/IAragonApp.sol
facuspagnuolo: Well, it is the expected way to define interfaces in solc... you won't notice a warning here because it is defined as a `contract`, but otherwise it would complain. However, not a blocker at all, we still need to define what do we want to include in the next breaking version, so we can discuss whether to address this or not

facuspagnuolo (approved):
LGTM!

contracts/test/mocks/apps/disputable/DisputableAppMock.sol
```diff
@@ -41,8 +42,13 @@ contract DisputableAppMock is DisputableAragonApp {
             iDisputable.onDisputableActionVoided.selector ^
             iDisputable.getAgreement.selector ^
             iDisputable.canChallenge.selector ^
-            iDisputable.canClose.selector ^
-            iDisputable.appId.selector;
+            iDisputable.canClose.selector;
+    }
+
+    function aragonAppInterfaceID() external pure returns (bytes4) {
```
sohkai: Let's move this out to a different test mock :)

test/contracts/apps/disputable/disputable_app.js
```diff
@@ -14,7 +14,8 @@ contract('DisputableApp', ([_, owner, agreement, anotherAgreement, someone]) =>
   let disputable, disputableBase, dao, acl
 
   const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000'
-  const DISPUTABLE_INTERFACE = '0x737c65f9'
+  const DISPUTABLE_INTERFACE = '0xf3d3bb51'
+  const ARAGON_APP_INTERFACE = '0x54053e6c'
```
sohkai, line 20: Curious about the rational for hardcoding these constants that we own (and change)—would it be a safer test if we got them from a deployed instance instead?

contracts/apps/disputable/IDisputable.sol
```diff
@@ -10,8 +10,7 @@ import "../../lib/standards/ERC165.sol";
 
 
 contract IDisputable is ERC165 {
-    bytes4 internal constant ERC165_INTERFACE_ID = bytes4(0x01ffc9a7);
-    bytes4 internal constant DISPUTABLE_INTERFACE_ID = bytes4(0x737c65f9);
+    bytes4 internal constant DISPUTABLE_INTERFACE_ID = bytes4(0xf3d3bb51);
```
sohkai, line 15: It would be useful to add a comment here to note down which methods are included in the interface ID (here and in the other contracts where we define these)!

contracts/apps/AragonApp.sol
```diff
@@ -18,7 +19,7 @@ import "../evmscript/EVMScriptRunner.sol";
 // Unless overriden, this behaviour enforces those contracts to be usable only behind an AppProxy.
 // ReentrancyGuard, EVMScriptRunner, and ACLSyntaxSugar are not directly used by this contract, but
 // are included so that they are automatically usable by subclassing contracts
-contract AragonApp is AppStorage, Autopetrified, VaultRecoverable, ReentrancyGuard, EVMScriptRunner, ACLSyntaxSugar {
+contract AragonApp is IAragonApp, AppStorage, Autopetrified, VaultRecoverable, ReentrancyGuard, EVMScriptRunner, ACLSyntaxSugar {
```
sohkai: I'm a bit concerned with `IAragonApp` and `AppStorage` both declaring essentially the same functions.

If I understand correctly, it's only because of the `supportsInterface()` that we wouldn't want `AppStorage` to inherit `IAragonApp`, and I'm dubious of how important it is for `supportsInterface()` to exist by default in every app.

I'm not even sure if it'll be that useful to declare an ERC165 interface ID for `IAragonApp`, but if it is considered useful, I might move the `supportsInterface()` implementation into the `AragonApp` contract instead.

test/contracts/apps/disputable/disputable_app.js
bingen: We also do that when we check them against `interfaceID()` methods.

contracts/apps/AragonApp.sol
bingen: I don’t like this option that much, because to me it feels natural that the `supportsInterface` method lives where the rest of the interface is defined, plus we have to remove the `ERC165` dependency from it too, which was nice in my opinion, but as it seems you both agree, I’m changing it.

contracts/apps/disputable/IDisputable.sol
```diff
@@ -10,6 +10,8 @@ import "../../lib/standards/ERC165.sol";
 
 
 contract IDisputable is ERC165 {
+    // Includes setAgreement, onDisputableActionChallenged, onDisputableActionAllowed,
+    // onDisputableActionRejected, onDisputableActionVoided, getAgreement, canChallenge and canClose:
```
sohkai:
```suggestion
    // onDisputableActionRejected, onDisputableActionVoided, getAgreement, canChallenge, and canClose methods:
```

test/contracts/apps/disputable/disputable_app.js
```diff
@@ -44,17 +46,19 @@ contract('DisputableApp', ([_, owner, agreement, anotherAgreement, someone]) =>
...
+      const erc165 = await ERC165.new()
       assert.isTrue(await disputable.supportsInterface(ERC165_INTERFACE), 'does not support ERC165')
 
-      assert.equal(await disputable.ERC165_INTERFACE(), ERC165_INTERFACE, 'ERC165 interface ID does not match')
-      assert.equal(await disputable.erc165interfaceID(), ERC165_INTERFACE, 'ERC165 interface ID does not match')
+      assert.equal(await erc165.interfaceID(), ERC165_INTERFACE, 'ERC165 interface ID does not match')
+      assert.equal(await erc165.ERC165_INTERFACE(), ERC165_INTERFACE, 'ERC165 interface ID does not match')
     })
 
     it('supports Aragon App interface', async () => {
```
sohkai, line 56: We should also add a suite of tests for the base `AragonApp` contract to test that it follows ERC165 and the `IAragonApp` interface.

contracts/apps/AragonApp.sol
```diff
@@ -5,21 +5,24 @@
...
 
 // Contracts inheriting from AragonApp are, by default, immediately petrified upon deployment so
 // that they can never be initialized.
 // Unless overriden, this behaviour enforces those contracts to be usable only behind an AppProxy.
 // ReentrancyGuard, EVMScriptRunner, and ACLSyntaxSugar are not directly used by this contract, but
 // are included so that they are automatically usable by subclassing contracts
-contract AragonApp is IAragonApp, AppStorage, Autopetrified, VaultRecoverable, ReentrancyGuard, EVMScriptRunner, ACLSyntaxSugar {
+contract AragonApp is ERC165, AppStorage, Autopetrified, VaultRecoverable, ReentrancyGuard, EVMScriptRunner, ACLSyntaxSugar {
+    // Includes appId and kernel methods:
+    bytes4 internal constant ARAGON_APP_INTERFACE_ID = bytes4(0x54053e6c);
```
sohkai: Does this need to move to the implementation contract? I was thinking the interfaces should declare their expected constants, especially in the case of ERC165 identifiers.

contracts/apps/disputable/IDisputable.sol
```diff
@@ -36,9 +37,7 @@ contract IDisputable is ERC165 {
     * @param _interfaceId The interface identifier being queried, as specified in ERC-165
     * @return True if the contract implements the requested interface and if its not 0xffffffff, false otherwise
     */
-    function supportsInterface(bytes4 _interfaceId) external pure returns (bool) {
-        return _interfaceId == DISPUTABLE_INTERFACE_ID || _interfaceId == ERC165_INTERFACE_ID;
+    function supportsInterface(bytes4 _interfaceId) public pure returns (bool) {
```
sohkai: Also wondering if this should be in the `Disputable` implementation

sohkai (commented):
Have a couple more comments... maybe it would be worth discussing the abstractions offline :).

> It feels natural that the `supportsInterface` method lives where the rest of the interface is defined

I wouldn't necessarily say so; ultimately the implementation contract has to decide which interfaces it'd like to publicly advertise. I also wouldn't couple an interface with ERC165, especially as a lot of ERC interfaces only suggest that a contract support ERC165.

contracts/apps/AragonApp.sol
bingen: Sounds good to me. Actually, as you know, I would have everything in the interface, haha.

contracts/apps/disputable/IDisputable.sol
bingen: But, what’s the reason? If you declare a contract to be `IDisputable`, then that contract will implement Disputable Interface, by definition (and ERC165 too). So the behavior of this function will be correct. If you are able to trick the compiler in a way that I can’t imagine, or for some reason that I can’t imagine either you want to hide that your contract implements those interfaces, you can always override the one here. But for regular cases you already have this “bureaucratic” function implemented in the base, so you don’t have to bother with it.
What’s the advantage of removing it from here?

sohkai: My main rationale was to align it with `IAragonApp`, which leaves its ERC-165 `supportsInterface()` out of the interface contract. Rather than include it implicitly, I'd prefer to leave such things to be done in the same, and obvious, manner at the implementation site, even if it's not DRY.

Also note that interface files are usually meant to be exported and copied around, and usually these do not include methods from ERC-165, as this is a separate interface.

However, I'm seeing that `IArbitrable` is also defining it inside of the interface, but `IArbitrator` does not even declare an ERC-165 interface at all.

PR #599, opened 2020-07-22
[DO NOT MERGE] Disputable: Move set agreement to Kernel

facuspagnuolo:
This PR moves the `setAgreement` security check to the Kernel, allowing us to define Agreement<>Kernel permissions instead of Agreement<>Disputable ones.

### DIFF
--- a/contracts/kernel/IKernel.sol
+++ b/contracts/kernel/IKernel.sol
@@ -17,11 +17,12 @@ interface IKernelEvents {
 
 // This should be an interface, but interfaces can't inherit yet :(
 contract IKernel is IKernelEvents, IVaultRecoverable {
-    function setAgreement(IDisputable _disputableApp, IAgreement _agreement) external;
-
     function acl() public view returns (IACL);
     function hasPermission(address who, address where, bytes32 what, bytes how) public view returns (bool);
 
     function setApp(bytes32 namespace, bytes32 appId, address app) public;
     function getApp(bytes32 namespace, bytes32 appId) public view returns (address);
+
+    // solium-disable function-order
+    function setAgreement(IDisputable _disputableApp, IAgreement _agreement) external;
 }
--- a/contracts/kernel/IKernel.sol
+++ b/contracts/kernel/IKernel.sol
@@ -6,6 +6,8 @@ pragma solidity ^0.4.24;
 
 import "../acl/IACL.sol";
 import "../common/IVaultRecoverable.sol";
+import "../apps/disputable/IAgreement.sol";
+import "../apps/disputable/IDisputable.sol";
 
 
 interface IKernelEvents {
@@ -15,6 +17,8 @@ interface IKernelEvents {
 
 // This should be an interface, but interfaces can't inherit yet :(
 contract IKernel is IKernelEvents, IVaultRecoverable {
+    function setAgreement(IDisputable _disputableApp, IAgreement _agreement) external;
+
     function acl() public view returns (IACL);
     function hasPermission(address who, address where, bytes32 what, bytes how) public view returns (bool);
 
--- a/contracts/kernel/Kernel.sol
+++ b/contracts/kernel/Kernel.sol
@@ -5,14 +5,14 @@ import "./KernelConstants.sol";
 import "./KernelStorage.sol";
 import "../acl/IACL.sol";
 import "../acl/ACLSyntaxSugar.sol";
+import "../apps/disputable/IAgreement.sol";
+import "../apps/disputable/IDisputable.sol";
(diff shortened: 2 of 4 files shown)


contracts/kernel/IKernel.sol
```diff
@@ -15,6 +17,8 @@ interface IKernelEvents {
 
 // This should be an interface, but interfaces can't inherit yet :(
 contract IKernel is IKernelEvents, IVaultRecoverable {
+    function setAgreement(IDisputable _disputableApp, IAgreement _agreement) external;
```
sohkai: Let's move this to be last in `IKernel` :).

contracts/kernel/IKernel.sol
```diff
@@ -15,6 +17,8 @@ interface IKernelEvents {
...
+
     function acl() public view returns (IACL);
     function hasPermission(address who, address where, bytes32 what, bytes how) public view returns (bool);
```
sohkai, line 22: Two comments:

1. Are we able to move all the interfaces to be external?
2. Should we use `AragonApp` as the type for `setApp()`'s `app` parameter? I'm debating if this interface dependency tree is worth it, but if not, then we may also want to type `setAgreement()` with just addresses.

test/contracts/apps/disputable/disputable_app.js
```diff
@@ -122,7 +126,11 @@ contract('DisputableApp', ([_, owner, agreement, anotherAgreement, someone]) =>
       const from = someone
 
       it('reverts', async () => {
-        await assertRevert(disputable.setAgreement(agreement, { from }), 'APP_AUTH_FAILED')
+        await assertRevert(dao.setAgreement(disputable.address, agreement, { from }), 'KERNEL_AUTH_FAILED')
+      })
+
+      it('reverts', async () => {
```
sohkai: We can probably come up with better test names, since these two are not the same test case :)

test/contracts/apps/disputable/disputable_app.js
```diff
@@ -67,17 +69,19 @@ contract('DisputableApp', ([_, owner, agreement, anotherAgreement, someone]) =>
...
+          await dao.setAgreement(disputable.address, agreement, { from })
 
           const currentAgreement = await disputable.getAgreement()
           assert.equal(currentAgreement, agreement, 'disputable agreement does not match')
         })
 
         it('emits an event', async () => {
-          const receipt = await disputable.setAgreement(agreement, { from })
+          const { tx } = await dao.setAgreement(disputable.address, agreement, { from })
+          const receipt = await web3.eth.getTransactionReceipt(tx)
```
sohkai, line 80: Oh interesting, was the receipt we got as the return of `dao.setAgreement()` not sufficient?

contracts/kernel/Kernel.sol
```diff
@@ -11,19 +11,30 @@ import "../common/Petrifiable.sol";
 import "../common/VaultRecoverable.sol";
 import "../factory/AppProxyFactory.sol";
 import "../lib/misc/ERCProxy.sol";
+import "../apps/disputable/IAgreement.sol";
+import "../apps/disputable/IDisputable.sol";
```
sohkai: Let's move these imports up to be right after the ACL imports :)

sohkai (approved):
A few minor comments, but LGTM!

For other's sake, we decided to batch this breaking change with a few more breaking changes down the road as we otherwise cannot use the same `DAOFactory`s and DAO templates as before to create test organizations with Agreements.

contracts/kernel/IKernel.sol
facuspagnuolo: I was planning to discuss this as one of the things to change in the new breaking version, but yea, I'd love to polish all the interfaces we provide

test/contracts/apps/disputable/disputable_app.js
facuspagnuolo: For some reason, it wasn't working. I think this will be fixed when we upgrade to Truffle 5 + helpers

izqui:
As an alternative implementation, we could introduce the concept of `ANY_ADDRESS` for the `where` value in the ACL, instead of hardcoding any knowledge about agreements in the Kernel.

It would work similarly to how `ANY_ADDRESS` works for `who`, basically giving a particular address permission over a certain role in any app in the organization. With that we could grant the Agreement instance permission to execute `SET_AGREEMENT` in any app in the DAO.

This change in the ACL could be useful for other cases as well, such as setting permissions just once for the admin functions of an app that the DAO has multiple instances of (e.g. multiple Voting instances).

PR #600, opened 2020-07-23
APM: Remove from repo

### LINKED ISSUE #368: Move APM into its own repo

facuspagnuolo:
Depends on https://github.com/aragon/apm/pull/1.

Fixes https://github.com/aragon/aragonOS/issues/368.


contracts/test/mocks/apps/disputable/AgreementMock.sol
```diff
@@ -7,7 +7,7 @@ contract AgreementMock {
         return 0;
     }
 
-    function closeAction(uint256 /* _actionId */) external {
+    function closeAction(uint256 /* _actionId */) external pure {
```
sohkai: We should probably lift this out to a different PR :).

I think we can also make `newAction()` pure now, if it's not `payable`?

sohkai (approved):
LGTM! So glad we're doing this!

PR #602, opened 2020-07-23
Agreement: Remove ETH as an option for fees

bingen:


contracts/test/mocks/apps/disputable/DisputableAppMock.sol
```diff
@@ -16,7 +16,7 @@ contract DisputableAppMock is DisputableAragonApp {
         initialized();
```
sohkai, line 16: I think we're missing one instance in the [`AgreementMock`](https://github.com/aragon/aragonOS/blob/next/contracts/test/mocks/apps/disputable/AgreementMock.sol#L5).

sohkai (approved):
LGTM

contracts/test/mocks/apps/disputable/DisputableAppMock.sol
facuspagnuolo: Yes, if you can also clean up a few compiler warnings about this one it would be really helpful!

facuspagnuolo (approved):
LGTM!

contracts/test/mocks/apps/disputable/AgreementMock.sol
```diff
@@ -1,8 +1,10 @@
 pragma solidity 0.4.24;
 
+import "../../../../apps/disputable/IAgreement.sol";
 
-contract AgreementMock {
-    function newAction(uint256 /* _disputableActionId */, bytes /* _context */, address /* _submitter */) external payable returns (uint256) {
+
+contract AgreementMock is IAgreement {
```
facuspagnuolo, line 6: I think you will need to mock all the methods then

PR #605, opened 2020-07-31
Disputable: Add tests to check role constants

bingen:


test/contracts/apps/disputable/disputable_app.js
```diff
@@ -1,3 +1,4 @@
+const { keccak_256 } = require('js-sha3')
```
sohkai: Haha, let's use `web3-utils`' version instead!

sohkai (approved):

facuspagnuolo (approved):
LGTM! I went ahead and push come polishing

PR #606, opened 2020-08-03
ACL: Rollback v2 changes

facuspagnuolo:
I created [this branch](https://github.com/aragon/aragonOS/tree/v6_alpha) to continue working on the related aragonOS v6 changes, which includes the changes I'm rolling back in this PR of course.


sohkai (approved):
LGTM, I noticed a few changes on the ACL against `master`, but I think these are fine in terms of an audit. We won't be re-deploying the ACL anyway.

Screen Shot 2020-08-04 at 5 07 54 PM

facuspagnuolo:
Yea, I fixed some indentation and inline doc, agreed that it shouldn't affect

PR #607, opened 2020-08-04
Disputable: Add Registry sample app

facuspagnuolo:
Moving this sample app from the Agreement app subrepo


contracts/apps/disputable/examples/RegistryApp.sol
```diff
@@ -0,0 +1,232 @@
+/*
```
sohkai: Perhaps let's move this to a top level `examples/` folder; I think `contracts` should only contain contracts we want to expose to others (especially in the npm package).

facuspagnuolo: I like the idea! Addressed now, let me know your thoughts :)

sohkai: Ah, I was actually meaning even outside of the `contracts/` folder so as to not confuse any one! Do you think it would still be worth leaving it in the `contracts/examples` folder?

facuspagnuolo: Sounds better!

sohkai (approved):
Looks good!

PR #608, opened 2020-08-07
Contracts: Use test helpers

bingen:
Depends on https://github.com/aragon/contract-helpers/pull/53


facuspagnuolo (commented):
Glad that we are dropping all this code! Well, actually reusing it from somewhere else haha
I think it would be nice to also update the test suite to Truffle 5 + web3 1.x, please take a look at my comments on the helpers PR :)

test/contracts/apps/app_proxy.js
```diff
@@ -16,9 +14,11 @@ const ERCProxyMock = artifacts.require('ERCProxyMock')
 const KernelSetAppMock = artifacts.require('KernelSetAppMock')
 
 const APP_ID = hash('stub.aragonpm.test')
-const ZERO_ADDR = '0x0000000000000000000000000000000000000000'
 const EMPTY_BYTES = '0x'
```
0xGabi, line 17: Wonder if we can replace `EMPTY_BYTES` everywhere for the one exported here: https://github.com/aragon/contract-helpers/blob/master/packages/test-helpers/src/bytes.js#L1

sohkai: Wasn't sure, but perhaps we forgot to do this? Although it is different as the one in the package is `0x00`.

test/contracts/common/safe_erc20.js
```diff
@@ -79,52 +80,52 @@ contract('SafeERC20', ([owner, receiver]) => {
...
+        const approvedAmount = bn(5000)
         await tokenMock.approve(safeERC20Mock.address, approvedAmount)
 
         // Transfer to receiver through the mock
         const receipt = await safeERC20Mock.transferFrom(tokenMock.address, owner, receiver, approvedAmount)
         assertMockResult(receipt, true)
-        assert.equal((await tokenMock.balanceOf(owner)).valueOf(), initialBalance - approvedAmount, 'Balance of owner should be correct')
-        assert.equal((await tokenMock.balanceOf(receiver)).valueOf(), approvedAmount, 'Balance of receiver should be correct')
-        assert.equal((await tokenMock.balanceOf(safeERC20Mock.address)).valueOf(), 0, 'Balance of mock should stay the same')
+        assertBn(await tokenMock.balanceOf(owner), initialBalance - approvedAmount, 'Balance of owner should be correct')
```
sohkai, line 94: Not sure, but are we able to do math on BNs like this? We don't need to use `sub()`?

test/contracts/common/time_helpers.js
```diff
@@ -7,17 +9,18 @@ contract('TimeHelpers', () => {
 
   it('checks block number', async () => {
     assert.equal((await timeHelpersMock.getBlockNumberExt()).toString(), (await timeHelpersMock.getBlockNumber64Ext()).toString(), "block numbers should match")
-    assert.equal((await timeHelpersMock.getBlockNumberExt()).toString(), (await timeHelpersMock.getBlockNumberDirect()).toString(), web3.eth.blockNumber, "block number should match with real one", "block number should match with real one")
+    assert.equal((await timeHelpersMock.getBlockNumberExt()).toString(), (await timeHelpersMock.getBlockNumberDirect()).toString(), web3.eth.getBlockNumber, "block number should match with real one", "block number should match with real one")
```
sohkai, line 12: A bit confused with this (although it's more with the original test)—do we not need to call `getBlockNumber()`?

test/contracts/common/unstructured_storage.js
```diff
@@ -23,12 +23,12 @@ contract('Unstructured storage', () => {
...
       assert.equal(
         await web3.eth.getStorageAt(appStorage.address, (await appStorage.getKernelPosition())),
-        (await appStorage.kernel()).toString(),
+        (await appStorage.kernel()).toString().toLowerCase(),
         'Kernel should match'
       )
       assert.equal(
         await web3.eth.getStorageAt(appStorage.address, (await appStorage.getKernelPosition())),
-        kernel.address,
+        kernel.address.toLowerCase(),
```
sohkai, line 31: I'm starting to wonder if we should have some sort of `assertAddressEquals()` or etc. so we don't have to do this :smile:

sohkai (commented):
🙌 👏 Absolutely brilliant.

test/contracts/apps/app_proxy.js
facuspagnuolo: All the tests using `0x00` as empty bytes were failing, we should change it in the helpers I think

test/contracts/common/unstructured_storage.js
facuspagnuolo: Haha this one was a bit particular

test/contracts/common/safe_erc20.js
facuspagnuolo: Yea, it shouldn't work this way, let me fix it anyway

test/contracts/common/time_helpers.js
facuspagnuolo: I think the original was malformed, let me change it

facuspagnuolo:
Thanks for the review @sohkai, addressed your comments [here](https://github.com/aragon/aragonOS/pull/611)

PR #609, opened 2020-08-10
Chore: Move CI to GH actions

facuspagnuolo:


.github/workflows/ci_contracts.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  CI:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Install node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 10
```
sohkai: Do we still need to be on node 10 because we're on truffle 4?

sohkai (approved):
LGTM!

.github/workflows/ci_contracts.yml
facuspagnuolo: Actually not anymore :D

PR #612, opened 2020-08-11
[Internal audit] aragonOS 5

sohkai:
Includes a number of small changes made during the audit, easiest to review commit by commit.

The only contract change:

- `IArbitrable`: aligns its ERC165 implementation with the other contracts


contracts/apps/disputable/IAgreement.sol
```diff
@@ -5,11 +5,12 @@
 pragma solidity ^0.4.24;
 
 import "../../acl/IACLOracle.sol";
-import "../../lib/token/ERC20.sol";
 import "../../lib/arbitration/IArbitrable.sol";
 import "../../lib/arbitration/IAragonAppFeesCashier.sol";
+import "../../lib/token/ERC20.sol";
 
 
+// Note
```
sohkai: This is my primary note.

I've started thinking about the `IAgreement` as a "disputable action registry" (`IDisputableActionRegistry`), and backtracking on whether we need everything in this interface.

From a Disputable's standpoint, I think the main things its "action registry" should care about are:
- Registering new actions, which may be disputable
- Challenging an action
- Closing old actions
- Getting a little bit of information about a particular action or challenge

With that in mind, I started thinking that specific implementation-specific information about the "action registry" would be reduced and queried separately from the action disputable's information (e.g. when using subgraphs). For example, I noticed you only used the Agreement to find the vote's ID when disputes were created in the DisputableVoting subgraph.

contracts/apps/disputable/DisputableAragonApp.sol
```diff
@@ -136,6 +136,7 @@ contract DisputableAragonApp is IDisputable, AragonApp {
     * @param _submitter Address that submitted the action
     * @return Unique identifier for the created action in the context of the Agreement
     */
+   // Note
     function _newAgreementAction(uint256 _disputableActionId, bytes _context, address _submitter) internal returns (uint256) {
```
sohkai: From above, I was thinking about decoupling aragonOS from "agreement"s, and having the concept of disputable action registries.

In that vein, I was thinking of renaming this `_registerDisputableAction()`.

contracts/apps/disputable/DisputableAragonApp.sol
```diff
@@ -145,6 +146,7 @@ contract DisputableAragonApp is IDisputable, AragonApp {
     * @dev Close action in the Agreement
     * @param _actionId Identifier of the action in the context of the Agreement
     */
+   // Note
     function _closeAgreementAction(uint256 _actionId) internal {
```
sohkai: Similarly, this would be `_closeDisputableAction()` or `_finishDisputableAction()`.

contracts/forwarding/IAbstractForwarder.sol
```diff
@@ -21,7 +21,8 @@ contract IAbstractForwarder {
 
     /**
     * @dev Tell whether the proposed forwarding path (an EVM script) from the given sender is allowed.
-    *      The implemented `forward()` method **MUST NOT** revert if `canForward()` returns true for the same parameters.
+    *      However, this is not a strict guarantee of safety: the implemented `forward()` method is
+    *      still allowed to revert even if `canForward()` returns true for the same parameters.
```
sohkai, line 25: Disputables can revert even if `canForward()` returns true, so I've left this as "not a strict guarantee" instead.

contracts/lib/arbitration/IAragonAppFeesCashier.sol
```diff
@@ -3,15 +3,16 @@ pragma solidity ^0.4.24;
 import "../token/ERC20.sol";
 
 
+// Note
```
sohkai: Similar to the other `arbitration/` contracts, should we add some:

- Natspec
- Information about where this came from (e.g. aragon-court)?

contracts/lib/arbitration/IArbitrable.sol
```diff
@@ -15,21 +15,20 @@ import "../../lib/standards/ERC165.sol";
 *      since we are using different solidity versions.
```
sohkai: I've changed this here to align it with the other ERC165 implementations we have, but have some doubts.

I am not sure if this and `IArbitrator` are actually useful to include here, as Disputables don't need to have any concept of these interfaces. Perhaps they should be moved to the Agreement implementation instead?

contracts/test/mocks/apps/disputable/AgreementMock.sol
```diff
@@ -18,6 +18,7 @@ contract AgreementMock is IAgreement {
 
     function newAction(uint256, bytes, address) external returns (uint256) {
         // do nothing
+        // Note
```
sohkai: I know it's a mock, but perhaps we should have explicit returns for all the functions with returns :)

test/contracts/factory/evm_script_factory.js
```diff
@@ -113,6 +113,7 @@ contract('EVM Script Factory', ([permissionsRoot]) => {
 
       // The executor always uses 0x for the input and callscripts always have 0x returns
       const expectedExecutor = await evmScriptReg.getScriptExecutor(createExecutorId(1))
+      // Note
```
sohkai: This is irrelevant for Disptuables, but I saw a number of places where we had `input: null, returnData: null` when checking `ScriptResults`.

Should we be able to assert for these?

contracts/lib/arbitration/IArbitrable.sol
facuspagnuolo: It's being used in the arbitrable events

contracts/apps/disputable/IAgreement.sol
facuspagnuolo: We decided not to force the `IArbitrable` and `IACL` inheritance at this level. The rest of the changes will be included in the aragonOS v6 roadmap

test/contracts/factory/evm_script_factory.js
facuspagnuolo: This is because we are using `AppStubScriptRunner` and there is no input and return data, it is actually logging null values

facuspagnuolo:
Thx for the audit @sohkai! I addressed your comments commit per commit, if it looks good we can release a new version and update the agreement and disputable voting apps

contracts/lib/arbitration/IArbitrable.sol
facuspagnuolo: Update: after an offline discussion about this, we decided to trim the `IArbitrable` interface removing all the arbitration-related info, please review commit https://github.com/aragon/aragonOS/pull/612/commits/bbd03046a11bf605b68a33acc6067500e550ca81

contracts/apps/disputable/IAgreement.sol
```diff
@@ -5,14 +5,10 @@
...
-import "../../lib/arbitration/IAragonAppFeesCashier.sol";
 
 
 contract IAgreement {
 
-    event Signed(address indexed signer, uint256 settingId);
-    event SettingChanged(uint256 settingId);
     event DisputableAppActivated(address indexed disputable);
     event DisputableAppDeactivated(address indexed disputable);
     event CollateralRequirementChanged(address indexed disputable, uint256 collateralRequirementId);
```
sohkai: Perhaps we can also remove the `DisputableAppActivated`, `DisputableAppDeactivated`, and `CollateralRequirementChanged` events (and associated functions)?

For me these now seem like "implementation details" that Disputables don't need to worry about.

contracts/apps/disputable/IAgreement.sol
```diff
@@ -57,18 +51,6 @@ contract IAgreement {
...
-        returns (
-            IArbitrator arbitrator,
-            IAragonAppFeesCashier aragonAppFeesCashier,
-            string title,
-            bytes content
-        );
-
     function getDisputableInfo(address _disputable) external view returns (bool activated, uint256 currentCollateralRequirementId);
 
     function getCollateralRequirement(address _disputable, uint256 _collateralId) external view
```
sohkai: Similarly, I'm wondering if `getDisputableInfo()` and `getCollateralRequirement()` are worth keeping.

contracts/lib/arbitration/IAragonAppFeesCashier.sol
```diff
@@ -3,16 +3,76 @@ pragma solidity ^0.4.24;
 import "../token/ERC20.sol";
 
 
-// Note
+/**
+* @title AragonAppFeesCashier interface
```
sohkai: Let's make sure we move this inline doc to aragon-court now that it's removed here :)

sohkai (commented):
Last few things, I'm still wondering if we should eliminate more from `IAgreement` 🔪

contracts/lib/arbitration/IAragonAppFeesCashier.sol
facuspagnuolo: Done!

facuspagnuolo:
@sohkai I like the idea! I addressed your comments, let me know your thoughts. About the app fees cashier inline doc, I had actually brought the content from the court repo, so I think we are fine :)

sohkai (commented):
💃🕺 Let's merge!

PR #614, opened 2020-08-24
Chore: Fix coverage task

facuspagnuolo:


.solcover.js
```diff
@@ -1,14 +1,13 @@
...
-    norpc: true,
-    compileCommand: '../node_modules/.bin/truffle compile',
-    testCommand: 'node --max-old-space-size=4096 ../node_modules/.bin/truffle test --network coverage',
-    skipFiles,
+  skipFiles: [
+    'lib',
+    'test',
+    'acl/ACLSyntaxSugar.sol',
+    'common/DepositableStorage.sol', // Used in tests that send ETH
+    'common/UnstructuredStorage.sol' // Used in tests that send ETH
```
sohkai: I think we might be able to remove these from the skipped files now since gas is less of an issue; worth a try?

sohkai (approved):
Looks good, I think we may be able to completely remove the ganache-cli script as a vanilla `truffle test` and `truffle coverage` seem to have us covered pretty well? Alternatively we can wait for when we migrate this to buidler (which also shouldn't be very hard I think) to do that.

.solcover.js
facuspagnuolo: Let's do it!

PR #616, opened 2020-08-25
Tests: Fix solidity tests

facuspagnuolo:


test/helpers/runSolidityTest.js
```diff
@@ -55,34 +39,33 @@ const linkLib = async (contract, libName) => {
...
       artifact.abi.forEach(interface => {
         if (interface.type === 'function') {
+          // Set up hooks
           if (['beforeAll', 'beforeEach', 'afterEach', 'afterAll'].includes(interface.name)) {
-            // Set up hooks
-            global[HOOKS_MAP[interface.name]](async () => {
-              const receipt = await deployed[interface.name]()
-              processResult(receipt, false)
-            })
+            global[HOOKS_MAP[interface.name]](async () => await deployed[interface.name]())
```
sohkai, line 66: Hmm, double checking here, does this fail properly if one of the hooks throws?

contracts/test/tests/TestDelegateProxySuicide.sol
```diff
@@ -0,0 +1,38 @@
+pragma solidity 0.4.24;
```
sohkai: Haha to be politically correct, let's rename this to be "self destructed"

sohkai (approved):

test/helpers/runSolidityTest.js
facuspagnuolo: In case some of these reverts, it throws. Note that previously `processResult` was called with `mustAssert` set to false always, therefore no assertions were actually being made, we were simply executing calls

